PR #3, opened 2014-10-08
Added MultiValuedLinkedHashMap to preserve insertion order.

geoffschoeman:
Added a MultiValuedLinkedHashMap to preserve insertion order in both the map and value list. 

My use case is to convert an XML string to a map using Jackson. The requirements stated that sub-elements could have duplicate tag names and element order must be preserved.

The MultiValuedLinkedHashMap  uses a LinkedHashMap with values of LinkedList to preserve order.


chtompki:
@geoffschoeman - Do you mind rebasing to master and re-opening this pull request?

PR #10, opened 2015-04-08
Add CollectionUtils#containsQuietly method and unit tests.

AndersDJohnson:
Add `CollectionUtils#containsQuietly` method and unit tests.

Currently suppressing both `NullPointerException` and `ClassCastException` if the element is bad, but I'm open to foregoing those features for just the collection null-safety.

Let me know if you prefer a different name - I considered `contains` (same name as proxied `Collection` method, as is typical in `StringUtils`), `containsIgnoreNull` (like existing `addIgnoreNull`), etc.

I recognize that it may be better practice to use empty collections rather than `null`, but this isn't always practical nor possible with 3rd party code.

If you are amenable to this, there are a number of other `Collection` methods for which we could offer safer proxy methods.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/CollectionUtils.java
+++ b/src/main/java/org/apache/commons/collections4/CollectionUtils.java
@@ -347,6 +347,30 @@ public class CollectionUtils {
     }
 
     /**
+     * Safe check if the specified collection {@code coll} contains the specified element {@code o}.
+     * If the element is null and the collection does not permit null elements, this method will
+     * catch the resulting {@link java.lang.NullPointerException} and return <code>false</code>.
+     * If the element is of a type that is incompatible with the collection, this method will
+     * catch the resulting {@link java.lang.ClassCastException} and return <code>false</code>.
+     *
+     * @param coll  the collection to check, may be null
+     * @param o  the element, may be null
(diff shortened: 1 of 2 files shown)


netomi:
Created ticket https://issues.apache.org/jira/browse/COLLECTIONS-561 for this pull request

src/main/java/org/apache/commons/collections4/CollectionUtils.java
```diff
@@ -347,6 +347,30 @@ private CollectionUtils() {}
...
+     * @return <code>true</code> iff the collection is not null and contains the element
+     * @since 4.1
+     */
+    public static boolean containsQuietly(final Collection<?> coll, final Object o) {
+        try {
+            return coll != null && coll.contains(o);
+        }
+        catch (NullPointerException e) {
+            return false;
+        }
```
peteryhwong: We should maintain coding standard here and go with.

```java
            } catch (final NullPointerException e) {
                return false;
            } catch (final ClassCastException e) {
                return false;
            }
```

chtompki:
@AndersDJohnson - Do you mind rebasing this work on the `master` branch and opening a new pull request?

AndersDJohnson:
@chtompki Is this still a desired feature? Don't want to the invest time unless there is interest.

PR #12, opened 2015-07-11
[COLLECTIONS-573] Add CollectionUtils#deepMerge method.

kaching88:
This method provides fast and convenient way to merge all deepest content in iterableOfiterablesOfiterables... etc. in one chosen collection. 

Method name is get from conventional Arrays.deep\* methods but i am open to community suggestions.

I also fixed some trivial mistakes in javadoc on CollectionUtils page.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/CollectionUtils.java
+++ b/src/main/java/org/apache/commons/collections4/CollectionUtils.java
@@ -1922,9 +1922,6 @@ public class CollectionUtils {
 	public static <T extends Collection<E>,E> T deepMerge(final Iterable<? extends Iterable<?>> iterableOfIterables, 
 			final T collectionToFill) {
 		Iterator<? extends Iterable<?>> iterator = iterableOfIterables.iterator();
-		if (!iterator.hasNext()) {
-			return collectionToFill;
-		}
 		while (iterator.hasNext()) {
 			deepMergeRecursion(iterator.next(), collectionToFill);
 		}


src/main/java/org/apache/commons/collections4/CollectionUtils.java
```diff
@@ -1889,4 +1889,66 @@ public static int maxSize(final Collection<? extends Object> coll) {
...
+     * @param collectionToFill an {@code collection} instance to fill up by deepest {@code iterable} content  
+     * @return {@code collectionToFIll} parameter filled up by deep content {@code iterableOfIterables} parameter
+     * @throws NullPointerException when one of a parameters is null
+     * @throws ClassCastException at runtime if {@code collectionToFill} is not bounded with valid parameter
+     * @since 4.1 
+     */
+	public static <T extends Collection<E>,E> T deepMerge(final Iterable<? extends Iterable<?>> iterableOfIterables, 
+			final T collectionToFill) {
+		Iterator<? extends Iterable<?>> iterator = iterableOfIterables.iterator();
+		if (!iterator.hasNext()) {
```
peteryhwong: This if statement is not necessary as the while loop below would handle the negative case.

kaching88: Oh... good point. I prepare correction commit.

kaching88:
Jira page: https://issues.apache.org/jira/browse/COLLECTIONS-573

src/test/java/org/apache/commons/collections4/CollectionUtilsTest.java
```diff
@@ -1906,5 +1906,69 @@ public int hash(String o) {
...
+        } // this is what we want
+        try {
+            CollectionUtils.deepMerge(null, null);
+            fail("expecting NullPointerException");
+        } catch (final NullPointerException npe) {
+        } // this is what we want
+        try {
+            List<Integer> failureList = CollectionUtils.deepMerge(setOfLists, new ArrayList<Integer>());
+            @SuppressWarnings("unused")
+			Integer integ = failureList.get(0);
```
RAnders00: The specified integer might be optimized away by the JIT compiler. Instead, you could assign `failureList.get(0)` to a class-wide volatile integer, which would prevent this optimization.

src/test/java/org/apache/commons/collections4/CollectionUtilsTest.java
```diff
@@ -1906,5 +1906,69 @@ public int hash(String o) {
         } catch (final NullPointerException npe) {
         } // this is what we want
     }
-
+    
+    @Test
+    public void testDeepMege() {
```
RAnders00: You might want to correct this spelling mistake.

chtompki:
Do you mind rebasing to master and re-opening this pull request?

src/main/java/org/apache/commons/collections4/CollectionUtils.java
```diff
@@ -1889,4 +1889,66 @@ public static int maxSize(final Collection<? extends Object> coll) {
...
+		if (!(firstElement instanceof Iterable<?>)) {
+			collectionToFill.add((E)firstElement);
+			while (iterator.hasNext()) {
+				collectionToFill.add((E)iterator.next());
+			}
+			return;
+		}
+		deepMergeRecursion((Iterable<?>) firstElement, collectionToFill);
+		while (iterator.hasNext()) {
+			deepMergeRecursion((Iterable<?>) iterator.next(), collectionToFill);
```
Klapsa2503: I suggest to change the order here:
1. change the condition `if (!(firstElement instanceof Iterable<?>)) {` -> `if (firstElement instanceof Iterable<?>) {`
2. add the `else `statement
3. remove `return;`

this will simplify the code as there will be:
* more clear condition
* no return statement in the middle of the method

PR #13, opened 2015-07-12
[COLLECTIONS-574] Improve ListUtils#longestCommonSubsequence methods.

kaching88:
Improved version of ListUtils#longestCommonSubsequence to take third or more parameters. 
After seriously consideration i add method with List<List<E>> as input parameter and remove depreciation from 'two list argument' overloads. It is much easier and convenient to wrap lists than taking attempts with varargs. 
New listOflists overloads do algorithm implementation object parts instantiation only once so they are more efficient for 3 and more lists than looping old 'two list argument' methods. 
I didn't do listOfLists overload for CharSequence method variant -  i think that need of passing more than 2 arguments in that method happens much less often than with other objects.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/ListUtils.java
+++ b/src/main/java/org/apache/commons/collections4/ListUtils.java
@@ -26,7 +26,6 @@ import java.util.List;
 
 import org.apache.commons.collections4.bag.HashBag;
 import org.apache.commons.collections4.functors.DefaultEquator;
-import org.apache.commons.collections4.iterators.ArrayIterator;
 import org.apache.commons.collections4.list.FixedSizeList;
 import org.apache.commons.collections4.list.LazyList;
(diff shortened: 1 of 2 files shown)


kaching88:
Jira page: https://issues.apache.org/jira/browse/COLLECTIONS-574

kinow:
Sounds like a good improvement. I'm simply triaging pull requests and issues, so not much time to calmly review your change and try to get it merged. However, I think your change contains tabs. Might be useful to replace them by spaces, as this would be required anyway later when merging it :-) So if you have time to update the pull request in the meantime, until myself or another developer has time to review, it would save some time.

Thanks for your contribution!

Cheers
Bruno

chtompki:
@kaching88 - Do you mind rebasing to the `master` branch and re-opening this pull request?

PR #15, opened 2015-11-17
Allow minus index

lamrongol:
Allow minus index to CircularFifoQueue, get(-1) is oldest value, get(-2) is second oldest value

### DIFF
--- a/src/main/java/org/apache/commons/collections4/queue/CircularFifoQueue.java
+++ b/src/main/java/org/apache/commons/collections4/queue/CircularFifoQueue.java
@@ -260,13 +260,13 @@ public class CircularFifoQueue<E> extends AbstractCollection<E>
      */
     public E get(final int index) {
         final int sz = size();
-        if (index < 0 || index >= sz) {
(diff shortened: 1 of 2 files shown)


izebit:
I think, the changes will break down compilation for legacy code

kinow:
Good noting the backward compatibility issue @izebit . I'd still be fine with that, but before accepting the change - which I'd say could be helpful, at least for those used with arrays/collections in R or Python - we must check if this would not be the only method supporting it.

If we add this behaviour to the `CircularFifoQueue` in Commons Collections, users could infer that other types were supposed to allow similar behaviour too.

In other words, I think this is useful, but might require a JIRA ticket to keep track of this change as an epic.

Cheers
Bruno

chtompki:
@lamrongol - Do you mind rebasing to the `master` branch, and re-opening this pull request?

lamrongol:
@chtompki Honestly, I didn't think the effect to other class and can't predict the effect. I hope experts of this repository decide.

PR #16, opened 2015-12-09
Fix <Generic type> not being displayed in javadoc

RAnders00:


kinow:
Nice catch.

Before the patch

screenshot from 2017-05-21 18-11-54

And fixed with the patch

screenshot from 2017-05-21 18-21-08

Created a patch from git, and applied to the Subversion repository. Should be automatically merged in the next minutes.

Thanks for your contribution @RAnders00 !

Cheers
Bruno

PR #17, opened 2016-07-28
small improvements for collectionUtils

izebit:


src/main/java/org/apache/commons/collections4/CollectionUtils.java
```diff
@@ -16,30 +16,15 @@
...
-import java.util.Map;
-import java.util.Set;
-
 import org.apache.commons.collections4.bag.HashBag;
-import org.apache.commons.collections4.collection.PredicatedCollection;
-import org.apache.commons.collections4.collection.SynchronizedCollection;
-import org.apache.commons.collections4.collection.TransformedCollection;
-import org.apache.commons.collections4.collection.UnmodifiableBoundedCollection;
-import org.apache.commons.collections4.collection.UnmodifiableCollection;
+import org.apache.commons.collections4.collection.*;
```
kinow: It is preferable to leave import each class individually.

src/main/java/org/apache/commons/collections4/CollectionUtils.java
```diff
@@ -180,8 +165,7 @@ public void setCardinality(final O obj, final int count) {
      * undesirable. This implementation only implements Collection.
      */
     @SuppressWarnings("rawtypes") // we deliberately use the raw type here
-    public static final Collection EMPTY_COLLECTION =
-        UnmodifiableCollection.unmodifiableCollection(new ArrayList<Object>());
+    public static final Collection EMPTY_COLLECTION = Collections.emptyList();
```
kinow: Nice catch!

src/main/java/org/apache/commons/collections4/CollectionUtils.java
```diff
@@ -1112,6 +1093,7 @@ public int hashCode() {
      * @return a boolean indicating whether the collection has changed or not.
      * @throws NullPointerException if the collection or iterator is null
      */
+    @SuppressWarnings("unchecked")
```
kinow: Is that necessary? I'm comparing with a local working copy of the trunk branch, looking ad the addAll method I have, it doesn't seem necessary. Perhaps it needs to be rebased and revised against last changes?

src/main/java/org/apache/commons/collections4/CollectionUtils.java
```diff
@@ -1266,9 +1248,6 @@ public static Object get(final Object object, final int index) {
         } else if (object instanceof Iterable<?>) {
             final Iterable<?> iterable = (Iterable<?>) object;
             return IterableUtils.get(iterable, i);
-        } else if (object instanceof Collection<?>) {
```
kinow: Is this one really safe to remove? It seems to assume that the else condition will handle with the same behaviour, but I have not tested it to confirm.

Might be worth also moving that `if (object == null)` check to the top.

src/main/java/org/apache/commons/collections4/CollectionUtils.java
izebit: Of cource, it is minnor. this is not necessary, but if it is applied, warnings about not correct cast will disappear. I talked about this code: `collection.addAll((Collection<? extends C>) iterable)`

src/main/java/org/apache/commons/collections4/CollectionUtils.java
izebit: removing of this condition is safe .moreover, this check is useless, because, any collection implements interface Iterable, and if object is Collection, then previous condition `(object instanceof Iterable<?>)` is true.

src/main/java/org/apache/commons/collections4/CollectionUtils.java
izebit: :))

src/main/java/org/apache/commons/collections4/CollectionUtils.java
izebit: ok, i got it)

src/main/java/org/apache/commons/collections4/CollectionUtils.java
kinow: Nice. Will check out the code to review with the IDE, plus some coffee :zzz: I think it looks good to be merged. Thanks for the quick replies @izebit

src/main/java/org/apache/commons/collections4/CollectionUtils.java
kinow: I might still be missing something here... Reviewing online, it didn't seem to be necessary. Just imported the code to Eclipse, and the project has 41 Warnings. This annotation being one of the warnings, being marked as "Unnecessary @SuppressWarnings("unchecked")".

src/main/java/org/apache/commons/collections4/CollectionUtils.java
kinow: Oooohh, just re-read, and finally understood. Sorry, nice catch again.

src/main/java/org/apache/commons/collections4/CollectionUtils.java
izebit: Hmm, It is very interesting. My IDE, (i use intellij idea), warns me about _unchecked cast_. Maybe,  it is a false warning. if you think that is not necessary, i will not persist :)

kinow: Removing in NetBeans (how nostalgic it was to use it) also did not raise any warnings. However, NetBeans has a lot of other warnings and errors that do not appear in Eclipse. I thought the IDE's would all agree upon what to check or not, but searching the Internet I found some other reports from differences in Eclipse/IntelliJ regarding these suppresswarnings cases. Let's leave as is for now, and we can revisit it later. So all good to be merged I believe?

screenshot_2017-05-24_11-28-05

izebit: I agree with you. You are right. I had try to compile the code by `javac` with flag `-Xlint:all` and did not see any warnings. Then Intellij Idea shows false warning or the warning are not for compiler, it is more for a programmer. 
Yes, all good to be merged.

kinow:
Placeholder to keep track of the change https://issues.apache.org/jira/browse/COLLECTIONS-603

kinow:
Merged. Issue created/updated. changes.xml updated. All tests pass, site reports look good, nothing for CollectionUtils in PMD, CPD, FindBugs. No changes in methods signatures.

Thanks for your contribution @izebit !

PR #26, opened 2017-09-12
COLLECTIONS-656 Update commons-parent version, and use inherited FindBugs plugin version

kinow:
Copied settings from [text] component after Javadocs issues were fixed but then the `mvn site` command failed in the FindBugs reporting.

With this change, `mvn clean site` works with Java 8, and produces the correct site reports.

Created the ticket and this pull request to wait for feedback & reviews in case there's any issue associated with upgrading the commons-parent version (there's none in this case as far as I can tell).

Feel free to comment here or merge the pull request (and add changes.xml entry...) in case you know there is nothing wrong in updating the commons-parent version.

Thanks
Bruno

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,7 +19,7 @@
   <parent>
     <groupId>org.apache.commons</groupId>
     <artifactId>commons-parent</artifactId>
-    <version>40</version>
+    <version>42</version>
   </parent>
   <modelVersion>4.0.0</modelVersion>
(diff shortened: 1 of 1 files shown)


kinow:
Fixed in 1081b725bea339fcda33d70c71b9eb1859d60d8f and 6a10051a7b5b3c615180413fe116c61780754969. Thanks to @garydgregory !

PR #34, opened 2018-02-01
COLLECTION-599: Fix for out-of-memory errors during session replication

saleem-akbar:
While using "non-sticky" session replication in a clustered environment,
the frequent de-serialisation meant the ReferenceMap's we were using
eventually consumed most of the server memory due to this bug.


saleem-akbar:
CI not working for this branch, so will look into upgrading to 4.x which contains same fix.

PR #35, opened 2018-02-01
COLLECTION-599: Fix for out-of-memory errors during session replication

saleem-akbar:
While using "non-sticky" session replication in a clustered environment,
the frequent de-serialisation meant the ReferenceMap's we were using
eventually consumed most of the server memory due to this bug.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/map/AbstractReferenceMap.java
+++ b/src/main/java/org/apache/commons/collections4/map/AbstractReferenceMap.java
@@ -1045,6 +1045,15 @@ public abstract class AbstractReferenceMap<K, V> extends AbstractHashedMap<K, V>
         final int capacity = in.readInt();
         init();
         data = new HashEntry[capacity];
+
+        // COLLECTIONS-599: Calculate threshold before populating, otherwise it will be 0 
+        // when it hits AbstractHashedMap.checkCapacity() and so will unnecessarily 
+        // double up the size of the "data" array during population.
+        //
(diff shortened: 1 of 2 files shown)


kinow:
@saleem-akbar, thanks for taking your time to prepare a pull request.

Looks like, actually, there are two pull requests from you, for the same issue. #34 and #35. This one has conflicting files, and cannot be easily merged.

#34 is not passing the CI server build.

Could you clarify which of these two pull requests should be used for COLLECTION-599, please? Also, if you could rebase/fix whichever is the good one, please. That'd make review&merging your changes must easier (and faster).

Thank you!
Bruno

saleem-akbar:
Hi,

#34 was for an older branch (apache:COLLECTIONS_3_2_X) .. as we’re still using that branch, ideally I’d like that version built - however I’m not entirely sure if the CI for that branch is working properly?

#35 is for the latest branch (apache:master) – however I believe one of your colleagues took my PR and modified the JUnit test case to use Java 7 idioms and merged it in to master, so I would imagine #35 is now redundant and can be closed:

https://issues.apache.org/jira/browse/COLLECTIONS-599?focusedCommentId=16349105&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16349105


Cheers
Saleem

garydgregory:
I'm not sure anyone wants to volunteer to release from such an old branch. I recommend you update to 4.x.

saleem-akbar:
OK, Thanks.  

I'll have to look into that, but I'll close both PR's anyway.

PR #40, opened 2018-06-12
COLLECTIONS-685: IterableUtils has public constructor

sfuhrm:
Constructor for Utils class was not private. 
This was obviously not intended as all other Utils classes have private constructors.


sfuhrm:
BTW, TravisCI is right, technically this is an API change, but I suggest that all API uses are neither useful nor make sense.

sfuhrm:
Closing because of breaking the API.
May be interesting for commons-cli 5.0.

PR #48, opened 2018-07-24
global: make use of FunctionalInterface

grimreaper:
Now that we're using Java 8, it seems reasonable to annotate.


soc:
@grimreaper Would it make sense in general to also annotate interfaces that just duplicate the ones in `java.util.function` with `@Deprecated`?

grimreaper:
IMHO yes, though that should be done separately.

PR #49, opened 2018-07-27
[SetUniqueList] avoid use of deprecated newInstance

grimreaper:


owen-mc:
I'm pretty sure this call to `addAll` shouldn't have been deleted:
https://github.com/apache/commons-collections/commit/b1c45ac691d46a8c609f2534d2adfa59c0599527#diff-8e53271d5d8299a76d43b0e3c81740fbe660083ae71c5bf2be63846d52156f23L344
I don't see how the elements of `list` are  added to the returned set without it.

PR #53, opened 2018-09-19
updating containsKey method in PassiveExpiringMap

thePatmanMI:
correcting the containsKey method to remove all expired entries to match the class and method javadoc statements


thePatmanMI:
I can do that.  Do you have a preference on whether I should use reflection to access the underlying map(s) (to determine size) or if I should change the underlying map(s) to be package scope instead?

thePatmanMI:
are there additional problems with this change, or can it be merged in?

garydgregory:
I am waiting for the comments from @kinow to be resolved, by @thePatmanMI or another.

PR #54, opened 2018-09-20
Replace Clirr with japicmp, since clirr does not handle Java 8 defaul…

britter:
…t methods

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -519,9 +519,6 @@
     <commons.scmPubUrl>https://svn.apache.org/repos/infra/websites/production/commons/content/proper/commons-collections</commons.scmPubUrl>
     <commons.scmPubCheckoutDirectory>site-content</commons.scmPubCheckoutDirectory>
 
-    <!-- Override clirr version to enable clirr on Java 8 -->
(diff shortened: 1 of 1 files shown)


britter:
The build produces:

```
[INFO] --- japicmp-maven-plugin:0.12.0:cmp (default-cli) @ commons-collections4 ---
[INFO] Skipping execution because parameter 'skip' was set to true.
```

And I don't understand why. When calling `mvn site`, the japicmp report is created. But the check goal is always skipped...

chtompki:
Yeah I did that. If the artifact is a pom only module, japicmp kinda blows up and falls over. So we have to un-skip it down here in the projects.

chtompki:
@britter - try japicmp version 0.13....

CasualSuperman:
> Yeah I did that. If the artifact is a pom only module, japicmp kinda blows up and falls over. So we have to un-skip it down here in the projects.

That sounds like a use case for the [packagingSupported](https://siom79.github.io/japicmp/MavenPlugin.html) configuration option. It even says:
>Helpful if you define the plugin in a root pom.

PR #55, opened 2018-10-02
COLLECTIONS-697 JavaDoc FixedSizeList should warn that modifying underlying list is not prevented

george-ranjan:
I just noticed that it is not explicitly mentioned in the JavaDoc that modifying the underlying list of a FixedSizeList would actually land up modifying the list of the constructed FixedSizeList. Not sure if this was by design, but at the very list I think the JavaDoc should caution against this.

This Pull Request has the necessary updates to the JavaDoc and a test that proves the findings.

NOTE: this is my first PR and I have tried to follow the guidelines.  Pleas let me know if I need to do anything more.  Thanks!

### DIFF
--- a/src/main/java/org/apache/commons/collections4/list/FixedSizeList.java
+++ b/src/main/java/org/apache/commons/collections4/list/FixedSizeList.java
@@ -31,7 +31,6 @@ import org.apache.commons.collections4.iterators.UnmodifiableIterator;
  * The add, remove, clear and retain operations are unsupported.
  * The set method is allowed (as it doesn't change the list size).
  * <p>
- * <p>
  * NOTE:
  * Modifying the decorated list directly would results in influencing the outcome
  * of method calls on this object. For example, the bounds of this list would reflect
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/commons/collections4/list/FixedSizeListTest.java
```diff
@@ -66,4 +68,20 @@ public String getCompatibilityVersion() {
 //        writeExternalFormToDisk((java.io.Serializable) getCollection(), "src/test/resources/data/test/FixedSizeList.fullCollection.version4.obj");
 //    }
 
+
```
grimreaper: extra newline

src/main/java/org/apache/commons/collections4/list/FixedSizeList.java
```diff
@@ -31,6 +31,12 @@
  * The add, remove, clear and retain operations are unsupported.
  * The set method is allowed (as it doesn't change the list size).
  * <p>
+ * <p>
```
grimreaper: extra `<p>`

george-ranjan:
@grimreaper fixed & pushed.

grimreaper:
cool. Now we're waiting for INFRA-17094.

PR #56, opened 2018-11-07
The put method should return the previous record only if not expired

ttulka:
Useful for the following scenario:
```
if (myCache.put(key, value) == null) {
  // the key is new 
} else {
  // the key is being altered
}
```


garydgregory:
Hi @ttulka ,

Do you have a unit test that demonstrates the pronlem?

Gary

ttulka:
Hi @garydgregory 
Yes, in the commit there is a test `testPut()` in the class `src/test/java/org/apache/commons/collections4/map/PassiveExpiringMapTest.java`

Cheers, Tomas

garydgregory:
In git master. Please verify and close.

PR #57, opened 2018-11-08
COLLECTIONS-701 SetUniqueList.add() crashes due to infinite recursion…

drajakumar:
… when it receives itself

### DIFF
--- a/src/main/java/org/apache/commons/collections4/list/SetUniqueList.java
+++ b/src/main/java/org/apache/commons/collections4/list/SetUniqueList.java
@@ -16,7 +16,6 @@
  */
 package org.apache.commons.collections4.list;
 
-import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


garydgregory:
Hi @drajakumar ,

I'm not sure this patch makes sense. Take a look at ```org.apache.commons.collections4.list.Collections701Test```: For ```ArrayList``` and ```HashSet```, adding a collection to itself is fine. 

In this patch, the argument is not only silently ignored, but the behavior is not even documented. Whatever we do, we really need to document _anything_ that deviates from the standard JRE ```List``` contract.

IMO, the fix should be so that a ```SetUniqueList``` behaves like a ```ArrayList``` and ```HashSet```, it just works.

garydgregory:
You did not have to close the PR, I was hoping you would provide a more complete solution ;-)

drajakumar:
@garydgregory can you kindly check the new fix, thank you!

drajakumar:
@garydgregory can you kindly check the new fix, thank you!

PR #63, opened 2018-11-30
[COLLECTIONS-707] Added PrefixMap - WIP!

nielsbasjes:
First version of the PrefixMap implementation.
I consider this a work in progress and expect various review comments.

I have included two implementations of this interface:
- StringPrefixMap where any String can be used as prefix.
- ASCIIPrefixMap where only readable ASCII is allowed as the key. This restriction cuts the retrieval time in half and as such may be useful where speed is very important.

About the tests:
- I have to add more tests regarding special characters in the Prefix.
- Right now there are also a few very simple performance tests to verify the lookup speed. It may be needed to remove these before committing.


nielsbasjes:
I released this as a separate project.
https://github.com/nielsbasjes/prefixmap

PR #64, opened 2018-12-03
[COLLECTIONS 706] Add set utils.as set method

reckart:
PR adds the method described in https://issues.apache.org/jira/projects/COLLECTIONS/issues/COLLECTIONS-706 .

Could be discussed how to handle this case:

```
SetUtils.asSet(null)
```

The compile will warn (at least in Eclipse) that this actually means passing a null array as the var-args argument. 

* Right now, I implemented it such that in this case, the method will return `null`. 
* It might be arguable to better return an empty set in this case.
* It could even be argued to return a set with the value `null` in this case. 

Opinions?

### DIFF
--- a/src/main/java/org/apache/commons/collections4/SetUtils.java
+++ b/src/main/java/org/apache/commons/collections4/SetUtils.java
@@ -48,7 +48,8 @@ import org.apache.commons.collections4.set.UnmodifiableSortedSet;
 public class SetUtils {
 
     /**
-     * Create an unmodifiable set from the given items.
+     * Create an unmodifiable set from the given items. If the passed var-args argument is {@code
+     * null}, then the method returns {@code null}.
      * @param <E> the element type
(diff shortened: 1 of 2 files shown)


kinow:
Ola @reckart !

Out of curiosity, do you have already a practical use for that in another ASF project, like UIMA perhaps?

And regarding the change, is it different than `SetUtils.unmodifiableSet` ? If so, do you think it would be worth to add a comment in the documentation about that?

Cheers
Bruno

garydgregory:
Closing; we have in 4.3:

- org.apache.commons.collections4.SetUtils.hashSet(E...)
- org.apache.commons.collections4.SetUtils.unmodifiableSet(E...)

reckart:
olá @kinow :) The use case was not in an ASF project but in another OS project I am working on. I don't remember the exact case though.

PR #66, opened 2019-01-19
COLLECTIONS-709 Set Entry count to 0 after remove

CasualSuperman:
Currently, after removing the last element of a MultiSet, the Entry doesn't have a count of 0. This commit fixes the count after removing the last of the current Entry of a MultiSet.


grimreaper:
This seems reasonable to me.

PR #68, opened 2019-02-09
[BulkTest] remove workarounds for buggy JDK 1.6 implementations

grimreaper:
Considering that we now require Java 8, having a workaround for IBM's
version of JDK 1.6 isn't required.


garydgregory:
How do we know this/these bug/bugs have been fixed in IBM Java 8? Do we have a URL or bug reference?

grimreaper:
> How do we know this/these bug/bugs have been fixed in IBM Java 8? Do we have a URL or bug reference?

I didn't look. However, this isn't a logic change and the tests aren't getting skipped. The tests explicitly check for JDK 1.6.

PR #71, opened 2019-05-14
Fix Travis build

kinow:
Use source=8 for maven-javadoc-plugin. Build passing on JDK 11 for me, on Ubuntu LTS, Maven 3.5.4.

Hoping Travis will pass too.


kinow:
Solution applied here [found on this issue](https://bugs.openjdk.java.net/browse/JDK-8212233).

PR #74, opened 2019-05-22
Collections-699 PairingIterator

Fival85:
Hello again,

like in https://github.com/apache/commons-collections/pull/73 here is the second try to implement COLLECTIONS-699

Now i use a standalone lightweight dto inside the PairingIterator to return the result.

Is the implementation ok?

### DIFF
--- a/src/main/java/org/apache/commons/collections4/iterators/PairingIterator.java
+++ b/src/main/java/org/apache/commons/collections4/iterators/PairingIterator.java
@@ -47,120 +47,120 @@ import org.apache.commons.collections4.iterators.PairingIterator.Entry;
  */
 public class PairingIterator<F, S> implements Iterator<Entry<F, S>> {
 
-	private final Iterator<F> firstIterator;
-	private final Iterator<S> secondIterator;
+    private final Iterator<F> firstIterator;
+    private final Iterator<S> secondIterator;
 
-	/**
-	 * Constructs a new <code>PairingIterator</code> that will provide a
-	 * {@link Entry} containing the child iterator elements.
-	 * 
-	 * @param firstIterator  the first iterator
-	 * @param secondIterator the second iterator
-	 */
-	public PairingIterator(Iterator<F> firstIterator, Iterator<S> secondIterator) {
-		this.firstIterator = firstIterator;
-		this.secondIterator = secondIterator;
-	}
+    /**
+     * Constructs a new <code>PairingIterator</code> that will provide a
+     * {@link Entry} containing the child iterator elements.
+     * 
+     * @param firstIterator  the first iterator
+     * @param secondIterator the second iterator
+     */
+    public PairingIterator(Iterator<F> firstIterator, Iterator<S> secondIterator) {
+        this.firstIterator = firstIterator;
+        this.secondIterator = secondIterator;
+    }
 
-	/**
-	 * Returns {@code true} if one of the child iterators has remaining elements.
-	 */
-	@Override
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/commons/collections4/iterators/PairingIterator.java
```diff
@@ -0,0 +1,166 @@
...
+     * 
+     * @return a {@link Entry} with the next values of child iterators
+     * @throws NoSuchElementException if no child iterator has any more elements
+     */
+    @Override
+    public Entry<F, S> next() throws NoSuchElementException {
+        if (!hasNext()) {
+            throw new NoSuchElementException();
+        }
+        final F firstValue = null != firstIterator && firstIterator.hasNext() ? firstIterator.next() : null;
```
garydgregory: It does not seem necessary to call each delegate iterator's `hasNext()` method since this method just invoked its own `hasNext()`. 

In general, this begs the question of what thread-safety guarantees this class makes. If none, the Javadocs should say so.

src/main/java/org/apache/commons/collections4/iterators/PairingIterator.java
```diff
@@ -0,0 +1,166 @@
...
+ * </p>
+ * <p>
+ * If one of the iterators is null, the result {@link Entry} contains also a
+ * null values until both child iterators exhausted.
+ * </p>
+ * <p>
+ * If one iterator has more elements then the other, the result {@link Entry}
+ * will contain a null value and the value of the not empty child iterator until
+ * both of the iterator exhausted.
+ *
```
garydgregory: You are missing a close tag for 'p' here.

src/main/java/org/apache/commons/collections4/iterators/PairingIterator.java
```diff
@@ -0,0 +1,166 @@
...
+ * @param <F> type of first value. The first value of {@link Entry}
+ *
```
garydgregory: This blank line is not needed.

src/main/java/org/apache/commons/collections4/iterators/PairingIterator.java
```diff
@@ -0,0 +1,166 @@
...
+        }
+    }
+
+    /**
+     * Contains two values of different generic types.
+     *
+     * @param <F> the type of the first element
+     * @param <S> the type of the second element
+     */
+    public static class Entry<F, S> {
```
garydgregory: This is not needed if you reuse one of:
- `org.apache.commons.collections4.keyvalue.DefaultKeyValue<K, V>`
- `org.apache.commons.collections4.keyvalue.DefaultMapEntry<K, V>`
- `org.apache.commons.collections4.keyvalue.MultiKey<K>`

src/test/java/org/apache/commons/collections4/iterators/PairingIteratorTest.java
```diff
@@ -0,0 +1,271 @@
...
+import java.util.NoSuchElementException;
+
+import org.apache.commons.collections4.IteratorUtils;
+import org.apache.commons.collections4.iterators.PairingIterator.Entry;
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * Unit test suite for {@link PairingIterator}.
+ *
```
garydgregory: Extra line not needed ;-)

Fival85:
Thank you. I will check it, when i have time for it. at the moment im on vacation :)

PR #78, opened 2019-07-18
[COLLECTIONS 714] Add tests

rohanpadhye:
Adding two tests that demonstrate the bug reported in [COLLECTIONS-714](https://issues.apache.org/jira/browse/COLLECTIONS-714). The tests currently fail, since I don't have a fix for the bug.

### DIFF
--- a/src/test/java/org/apache/commons/collections4/trie/PatriciaTrieTest.java
+++ b/src/test/java/org/apache/commons/collections4/trie/PatriciaTrieTest.java
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.ConcurrentModificationException;
 import java.util.HashSet;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.NoSuchElementException;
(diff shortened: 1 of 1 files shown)


garydgregory:
Since the tests fail, please decorate the methods with @Ignore("COLLECTIONS-714")

rohanpadhye:
The `@Ignore` annotation only appears to work with JUnit4-style tests (`org.junit.Test`). The `PatriciaTrieTest` suite appears to using a JUnit3 style (`junit.framework.Test`), which doesn't support `@Ignore`. Am I mistaken here?

I have given project maintainers edit access to this pull request, so feel free to push any changes you desire directly to the `bug-collections-714` branch.

garydgregory:
@rohanpadhye You are not mistaken indeed. Commons Collections needs a port to JUnit 4 for `@Ignore` to to work properly. We cannot merge this patch as is since it will make the build fail.

saipraneethn:
You can add a prefix of X for those test cases as the JUnit will be looking for the methods starting with test.

garydgregory:
> You can add a prefix of X for those test cases as the JUnit will be looking for the methods starting with test.

Don't do that, that's why one uses `@Ignore`.

PR #86, opened 2019-09-30
Add test cases to IteratorUtilsTest

dota17:
Add testcases to IteratorUtilsTest


garydgregory:
Thank you @dota17 ! 

That really improved code coverage for this class :-)

The remaining area of improvement in this class appears to be in `getIterator(Object)` and some other little nooks and crannies.

PR #89, opened 2019-10-22
[COLLECTIONS-708]Add hashCode method to CollectionUtils

dota17:
For [COLLECTIONS 708](https://issues.apache.org/jira/projects/COLLECTIONS/issues/COLLECTIONS-708?filter=allopenissues)

### DIFF
--- a/src/main/java/org/apache/commons/collections4/CollectionUtils.java
+++ b/src/main/java/org/apache/commons/collections4/CollectionUtils.java
@@ -624,9 +624,7 @@ public class CollectionUtils {
      */
     public static <E> int hashCode(final Collection<? extends E> collection,
                                 final Equator<? super E> equator) {
-        if (null == equator) {
-            throw new NullPointerException( "Equator must not be null." );
-        }
+        Objects.requireNonNull(equator, "Equator");
         if (null == collection) {
             return 0;
         }
@@ -636,7 +634,7 @@ public class CollectionUtils {
         }
         return hashCode;
     }
-    
+
     /**
      * Wraps another object and uses the provided Equator to implement
      * {@link #equals(Object)} and {@link #hashCode()}.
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/collections4/CollectionUtils.java
```diff
@@ -609,6 +609,33 @@ public static boolean isEqualCollection(final Collection<?> a, final Collection<
         return isEqualCollection(collect(a, transformer), collect(b, transformer));
     }
 
+    /**
+     * Returns the hash code of Collection through the hash method of Equator.
```
kinow: Maybe match the text in the `@param` docs: "Returns the hash code the the input collection using the hash method of the equator."?

src/main/java/org/apache/commons/collections4/CollectionUtils.java
```diff
@@ -609,6 +609,33 @@ public static boolean isEqualCollection(final Collection<?> a, final Collection<
...
+     * <p>
+     * If the input collection is null return 1.
```
kinow: I think it returns `0`? Maybe something like: "When the collection is {@code null} it returns 0 (zero). The equator must not be {@code null}, otherwise a {@code NullPointerException} is raised." (sorry my rusty Javadoc, I think there's a way to format code/numbers/classes/etc).

src/main/java/org/apache/commons/collections4/CollectionUtils.java
```diff
@@ -624,9 +624,7 @@ public static boolean isEqualCollection(final Collection<?> a, final Collection<
      */
     public static <E> int hashCode(final Collection<? extends E> collection,
                                 final Equator<? super E> equator) {
-        if (null == equator) {
-            throw new NullPointerException( "Equator must not be null." );
-        }
+        Objects.requireNonNull(equator, "Equator");
```
garydgregory, line 627: "Equator" -> "equator"

dota17: Thanks @garydgregory . I had updated Equator to equator .

kinow (approved):

kinow:
Merged. Thanks a lot for your contributions @dota17 !

PR #91, opened 2019-10-23
[COLLECTIONS-674] Add drain method to CollectionUtils

dota17:
For [COLLECTIONS-674](https://issues.apache.org/jira/projects/COLLECTIONS/issues/COLLECTIONS-674?filter=allopenissues)

### DIFF
--- a/src/main/java/org/apache/commons/collections4/CollectionUtils.java
+++ b/src/main/java/org/apache/commons/collections4/CollectionUtils.java
@@ -1792,21 +1792,28 @@ public class CollectionUtils {
     }
 
     /**
-     * Removes elements whose index are between startIndex and endIndex in the collection and returns them.
+     * Removes elements whose index are between startIndex, inclusive and endIndex,
+     * exclusive in the collection and returns them.
      * This method modifies the input collections.
      *
      * @param <E>  the type of object the {@link Collection} contains
      * @param input  the collection will be operated, can't be null
-     * @param startIndex  the start index to remove element, can't be less than 0
-     * @param endIndex  the end index to remove, can't be less than startIndex
+     * @param startIndex  the start index (inclusive) to remove element, can't be less than 0
+     * @param endIndex  the end index (exclusive) to remove, can't be less than startIndex
      * @return collection of elements that removed from the input collection
      * @since 4.5
      */
-    public static <E> Collection<E> removeRange(Collection<E> input, int startIndex, int endIndex) {
-        if (endIndex < startIndex) {
-            throw new IllegalArgumentException("The end index can't less than the start index.");
+    public static <E> Collection<E> removeRange(final Collection<E> input, final int startIndex, final int endIndex) {
+        if (null == input) {
+            throw new IllegalArgumentException("The collection can't be null.");
         }
-        return CollectionUtils.removeCount(input, startIndex, endIndex-startIndex);
+        if (endIndex < startIndex) {
+            throw new IllegalArgumentException("The end index can't be less than the start index.");
+        }
+        if (input.size() < endIndex) {
+            throw new IndexOutOfBoundsException("The end index can't be greater than the size of collection.");
+        }
+        return CollectionUtils.removeCount(input, startIndex, endIndex - startIndex);
     }
 
     /**
@@ -1815,23 +1822,23 @@ public class CollectionUtils {
      *
      * @param <E>  the type of object the {@link Collection} contains
      * @param input  the collection will be operated, can't be null
-     * @param startIndex  the start index to remove element, can't be less than 0
+     * @param startIndex  the start index (inclusive) to remove element, can't be less than 0
      * @param count  the specified number to remove, can't be less than 1
      * @return collection of elements that removed from the input collection
      * @since 4.5
      */
-    public static <E> Collection<E> removeCount(Collection<E> input, int startIndex, int count) {
+    public static <E> Collection<E> removeCount(final Collection<E> input, int startIndex, int count) {
         if (null == input) {
             throw new IllegalArgumentException("The collection can't be null.");
         }
         if (startIndex < 0) {
-            throw new IllegalArgumentException("The start index can't less than 0.");
+            throw new IndexOutOfBoundsException("The start index can't be less than 0.");
         }
         if (count < 0) {
-            throw new IllegalArgumentException("The count can't less than 0.");
+            throw new IndexOutOfBoundsException("The count can't be less than 0.");
         }
         if (input.size() < startIndex + count) {
-            throw new IllegalArgumentException(
+            throw new IndexOutOfBoundsException(
                     "The sum of start index and count can't be greater than the size of collection.");
         }
 
--- a/src/main/java/org/apache/commons/collections4/CollectionUtils.java
+++ b/src/main/java/org/apache/commons/collections4/CollectionUtils.java
@@ -1792,6 +1792,65 @@ public class CollectionUtils {
     }
 
     /**
+     * Removes elements whose index are between startIndex and endIndex in the collection and returns them.
+     * This method modifies the input collections.
+     *
+     * @param <E>  the type of object the {@link Collection} contains
+     * @param input  the collection will be operated, can't be null
+     * @param startIndex  the start index to remove element, can't be less than 0
+     * @param endIndex  the end index to remove, can't be less than startIndex
+     * @return collection of elements that removed from the input collection
+     * @since 4.5
+     */
+    public static <E> Collection<E> removeRange(Collection<E> input, int startIndex, int endIndex) {
+        if (endIndex < startIndex) {
+            throw new IllegalArgumentException("The end index can't less than the start index.");
+        }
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/collections4/CollectionUtils.java
```diff
@@ -909,6 +909,46 @@ public int hashCode() {
...
+     * @return collection of elements that removed from the input collection
+     */
+    public static <E> Collection<E> drain(Collection<E> input, int start, int count) {
+		if (null == input) {
+			throw new IllegalArgumentException("The collection can't be null.");
+		}
+		if (start < 1) {
+			throw new IllegalArgumentException("The Start can't less than 1.");
+		}
+		if (count < 1) {
```
garydgregory: A count of 0 seems like it should be legal input, ending up removing nothing.

src/main/java/org/apache/commons/collections4/CollectionUtils.java
```diff
@@ -909,6 +909,46 @@ public int hashCode() {
...
+     * Removes the specified number of elements from the specified location in the collection and returns them.
+     * This method would have the side-effect of modifying the input collections.
+     *
+     * @param <E>  the type of object the {@link Collection} contains
+     * @param input  the collection will be operated,can't be null
+     * @param start  the specified position to remove element,can't be less than 1
+     * @param count  the specified number to remove,can't be less than 1
+     * @return collection of elements that removed from the input collection
+     */
+    public static <E> Collection<E> drain(Collection<E> input, int start, int count) {
```
garydgregory: We already have methods named "remove*" in this class and the fact that your comments starts with "Removes" shows to me that this method should also be prefixed with "remove".

src/main/java/org/apache/commons/collections4/CollectionUtils.java
```diff
@@ -909,6 +909,46 @@ public int hashCode() {
...
+		if (null == input) {
+			throw new IllegalArgumentException("The collection can't be null.");
+		}
+		if (start < 1) {
```
garydgregory: This seems wrong, collections are addressed starting at index 0. I should be able to call this method and have it remove all elements in one go by passing in a start of 0 and a count equal to the collection size.

garydgregory:
You should check your tests before committing ;-) as the build fails.

WRT API name, `removeRange` is reminiscent of `java.util.Arrays.copyOfRange(...)` so it should be OK: we are removing a range of items specified by input args.

But we can do better than `removeByDrain` IMO. "ByDrain" is confusing since 'drain' and 'remove' are synonymous.

It feels to me like a `removeCount` is more precise.

You could reuse `.CollectionUtils.checkIndexBounds(int)`

You should also clean up your formatting as it seems all over the place, looking at it in GitHub at least.

src/main/java/org/apache/commons/collections4/CollectionUtils.java
```diff
@@ -1791,6 +1791,63 @@ public static int maxSize(final Collection<? extends Object> coll) {
...
+     *
+     * @param <E>  the type of object the {@link Collection} contains
+     * @param input  the collection will be operated,can't be null
+     * @param startIndex  the start index to remove element,can't be less than 0
+     * @param endIndex  the end index to remove,can't be less than startIndex
+     * @return collection of elements that removed from the input collection
+     */
+    public static <E> Collection<E> removeRange(Collection<E> input, int startIndex, int endIndex) {
+        if (endIndex < startIndex) {
+            throw new IllegalArgumentException("The start index can't less than the end index.");
```
garydgregory: This message is backward.

src/main/java/org/apache/commons/collections4/CollectionUtils.java
```diff
@@ -1791,6 +1791,63 @@ public static int maxSize(final Collection<? extends Object> coll) {
         return list;
     }
 
+    /**
+     * Removes elements whose index are between startIndex and endIndex in the collection and returns them.
+     * This method would have the side-effect of modifying the input collections.
+     *
+     * @param <E>  the type of object the {@link Collection} contains
+     * @param input  the collection will be operated,can't be null
+     * @param startIndex  the start index to remove element,can't be less than 0
```
garydgregory: Missing space after the comma.

src/main/java/org/apache/commons/collections4/CollectionUtils.java
```diff
@@ -1791,6 +1791,63 @@ public static int maxSize(final Collection<? extends Object> coll) {
...
+     * @param endIndex  the end index to remove,can't be less than startIndex
+     * @return collection of elements that removed from the input collection
+     */
```
garydgregory: Add `@since` tag.

src/main/java/org/apache/commons/collections4/CollectionUtils.java
```diff
@@ -1791,6 +1791,63 @@ public static int maxSize(final Collection<? extends Object> coll) {
...
+
+    /**
+     * Removes the specified number of elements from the start index in the collection and returns them.
+     * This method would have the side-effect of modifying the input collections.
+     *
+     * @param <E>  the type of object the {@link Collection} contains
+     * @param input  the collection will be operated,can't be null
+     * @param startIndex  the start index to remove element,can't be less than 0
+     * @param count  the specified number to remove,can't be less than 1
+     * @return collection of elements that removed from the input collection
```
garydgregory: Add `@since` tag.

src/main/java/org/apache/commons/collections4/CollectionUtils.java
```diff
@@ -1791,6 +1791,63 @@ public static int maxSize(final Collection<? extends Object> coll) {
...
+    public static <E> Collection<E> removeRange(Collection<E> input, int startIndex, int endIndex) {
+        if (endIndex < startIndex) {
+            throw new IllegalArgumentException("The start index can't less than the end index.");
+        }
+        return CollectionUtils.removeCount(input, startIndex, endIndex-startIndex);
+    }
+
+    /**
+     * Removes the specified number of elements from the start index in the collection and returns them.
+     * This method would have the side-effect of modifying the input collections.
```
garydgregory: Replace:
This method would have the side-effect of modifying the input collections.

With:
This method modifies the input collection.

src/main/java/org/apache/commons/collections4/CollectionUtils.java
```diff
@@ -1791,6 +1791,63 @@ public static int maxSize(final Collection<? extends Object> coll) {
...
+        }
+        if (input.size() < startIndex + count) {
+            throw new IllegalArgumentException(
+                    "The sum of start index and count cann't be greater than the size of collection.");
+        }
+
+        Collection<E> result = new ArrayList<E>(count);
+        Iterator<E> iterator = input.iterator();
+        while (count > 0) {
+            if(startIndex > 0) {
```
garydgregory: "if(" -> "if ("

src/test/java/org/apache/commons/collections4/CollectionUtilsTest.java
```diff
@@ -1472,6 +1472,101 @@ public void testRetainAll() {
         } // this is what we want
     }
 
+    @Test
+    public void testRemoveRange() {
+        List<Integer> list = new ArrayList<Integer>();
```
garydgregory: FYI: You don't need to over specify generics in Java 8 anymore, so:
```
List<Integer> list = new ArrayList<Integer>();
```
can simply be:
```
List<Integer> list = new ArrayList<>();
```

garydgregory: Also please rebase on master, then I'll take another look. TY!

src/main/java/org/apache/commons/collections4/CollectionUtils.java
```diff
@@ -1791,6 +1791,65 @@ public static int maxSize(final Collection<? extends Object> coll) {
...
     }
 
+    /**
+     * Removes elements whose index are between startIndex and endIndex in the collection and returns them.
+     * This method modifies the input collections.
+     *
+     * @param <E>  the type of object the {@link Collection} contains
+     * @param input  the collection will be operated, can't be null
+     * @param startIndex  the start index to remove element, can't be less than 0
+     * @param endIndex  the end index to remove, can't be less than startIndex
```
garydgregory: The Javadoc for indexes should specify whether they are inclusive or exclusive, for example, see https://docs.oracle.com/javase/8/docs/api/java/util/List.html#subList-int-int-

src/main/java/org/apache/commons/collections4/CollectionUtils.java
```diff
@@ -1791,6 +1791,65 @@ public static int maxSize(final Collection<? extends Object> coll) {
...
+     * @return collection of elements that removed from the input collection
+     * @since 4.5
+     */
+    public static <E> Collection<E> removeRange(Collection<E> input, int startIndex, int endIndex) {
```
garydgregory: Please use `final` when possible; your IDE might do this for you (Eclipse can for example.)

PR #99, opened 2019-10-30
[COLLECTIONS-714] Throw exception when put \u0000 to trie

dota17:
For issue [COLLECTIONS-714](https://issues.apache.org/jira/projects/COLLECTIONS/issues/COLLECTIONS-714).
Don't support '\u0000' in the trie's key.


src/main/java/org/apache/commons/collections4/trie/analyzer/StringKeyAnalyzer.java
```diff
@@ -82,6 +82,9 @@ public int bitIndex(final String key, final int offsetInBits, final int lengthIn
                 k = 0;
             } else {
                 k = key.charAt(index1);
+                if (k == 0) {
+                    throw new IllegalArgumentException("Don't support '\\u0000' in the key.");
+                }
```
melloware: A better message might be "Character '\\u0000' is not supported in the key."

melloware (changes requested):

src/main/java/org/apache/commons/collections4/trie/analyzer/StringKeyAnalyzer.java
dota17: '\\u0000' is shown as \u0000, '\u0000' is shown as blank space.

dota17:
@melloware Thank you for your suggestion. 
If anyone has time to review, it would be really nice. ;-)

Thanks

src/main/java/org/apache/commons/collections4/trie/analyzer/StringKeyAnalyzer.java
garydgregory: Could/Should this method be rewritten to allow char 0 instead?

garydgregory:
Hello @dota17 
What about other control characters? Control characters are 0-31 and 127?
Either 0 is truly a special case, or all control characters are. If 0 is special, then it can have its own test. Non-0 control-characters can have their own `@ParameterizedTest`.
TY!

PR #101, opened 2019-10-31
Add three test cases in CircularFifoQueueTest

dota17:


src/test/java/org/apache/commons/collections4/queue/CircularFifoQueueTest.java
```diff
@@ -421,6 +421,54 @@ public void testGetIndex() {
...
         }
     }
+	
+	public void testAddNull() {
+        final CircularFifoQueue<E> b = new CircularFifoQueue<>(2);
+        try {
+            b.add(null);
+            fail();
+        } catch (final NullPointerException ex) {
+            assertEquals("Attempted to add null object to queue",ex.getMessage());
```
garydgregory: -1: I do not think we should test the actual text of the message, it is too brittle IMO. We should not break test when we are improving or updating exception messages later on. Testing that the exception message is not empty is OK though. Also there is a space missing after the comma.

src/test/java/org/apache/commons/collections4/queue/CircularFifoQueueTest.java
```diff
@@ -421,6 +421,54 @@ public void testGetIndex() {
...
+        fifo.add((E) "1");
+        fifo.add((E) "2");
+        fifo.add((E) "3");
+        fifo.add((E) "4");
+        fifo.add((E) "5");
+        assertEquals(5,fifo.size());
+        try {
+            fifo.get(5);
+        } catch (final NoSuchElementException ex) {
+            assertEquals("The specified index (5) is outside the available range [0, 5)",ex.getMessage());
```
garydgregory: Same as my previous comment. This is a place where I would want to change the message, there is no reason why the number 5 should be in parens. Also the range syntax is left to comprehend as an exercise to the reader, which is not great IMO.

src/test/java/org/apache/commons/collections4/queue/CircularFifoQueueTest.java
```diff
@@ -421,6 +421,54 @@ public void testGetIndex() {
...
+        fifo.add((E) "1");
+        fifo.add((E) "2");
+        fifo.add((E) "3");
+        fifo.add((E) "4");
+        fifo.add((E) "5");
+        assertEquals(5,fifo.size());
+        try {
+            fifo.get(-2);
+        } catch (final NoSuchElementException ex) {
+            assertEquals("The specified index (-2) is outside the available range [0, 5)",ex.getMessage());
```
garydgregory: Same comment as previous.

src/test/java/org/apache/commons/collections4/queue/CircularFifoQueueTest.java
dota17: Ok， I got it, thanks !

PR #103, opened 2019-10-31
Add junit for getWithNull and subList in the LazyListTest.

dota17:


src/test/java/org/apache/commons/collections4/list/FixedSizeListTest.java
```diff
@@ -83,4 +83,66 @@ public void testListAllowsMutationOfUnderlyingCollection() {
         Assert.assertEquals("Modifying an the underlying list is allowed",
                 sizeBefore + 1, fixedSizeList.size());
     }
+
+    //-----------------------------------------------------------------------
```
garydgregory: Let's not add these kinds of `//---...` lines as they appear, to me, as just noise.

dota17: Ok, I have removed all the new ones, thank you for your suggestion.

PR #109, opened 2019-11-05
[COLLECTIONS-591] Copy on write list

dota17:
[COLLECTIONS-591] Add Copy on write list


src/test/java/org/apache/commons/collections4/list/CopyOnWriteListTest.java
```diff
@@ -0,0 +1,417 @@
...
+        @Override
+        public void execute(Integer param, List<Integer> i) {
+            i.clear();
+        }
+    };
+
+    private interface Task<P, T> {
+        void execute(P param, T i);
+    }
+}
```
kinow: Not a proper review @dota17 as this would require more time than I have right now. But this file is missing the extra newline for git. :+1: and thanks for all your contributions :tada:

PR #110, opened 2019-11-06
 [COLLECTIONS-663] Update AbstractMultiValuedMap asMap()

dota17:
[COLLECTIONS-663] https://issues.apache.org/jira/browse/COLLECTIONS-663
Hi:
I think the representation of Multimap is {1: A, 1: B, 2: C}. So when A and B are deleted, only {2: C} is left. And asMap() method  returns a map in the form of {1：[A, B], 2：[C]}. In this map, {1：[], 2:[C]} is left when A and B are deleted. So wrappedCollection(key) should not be used。 It willl show as {1: A, 1: B, 2: C} when wrappedCollection(key) is used, and the asMap() method does not make sense.


src/main/java/org/apache/commons/collections4/multimap/AbstractMultiValuedMap.java
```diff
@@ -879,7 +879,8 @@ public boolean remove(final Object o) {
             public Map.Entry<K, Collection<V>> next() {
                 final Map.Entry<K, Collection<V>> entry = super.next();
                 final K key = entry.getKey();
-                return new UnmodifiableMapEntry<>(key, wrappedCollection(key));
+                final Collection<V> value = entry.getValue();
+                return new UnmodifiableMapEntry<>(key, value);
```
garydgregory, line 883: You can just inline `key` and `value` IMO.

PR #111, opened 2019-11-07
Add TransformedDeque for deque interface

dota17:
I found someone raised that adding support for deque interface in Collections.(https://issues.apache.org/jira/browse/COLLECTIONS-563) 
(https://issues.apache.org/jira/browse/COLLECTIONS-564)
This makes sense so I wrote some code to make a try. I want to know if my implementation is appropriate, if my PR is merged or given comments and suggestions, I would be happy to finish the rest of the work.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/deque/TransformedDeque.java
+++ b/src/main/java/org/apache/commons/collections4/deque/TransformedDeque.java
@@ -1,0 +1,223 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/commons/collections4/deque/CircularDeque.java
```diff
@@ -0,0 +1,636 @@
...
+ * All other operations perform in linear time or worse.
+ * </p>
+ * <p>
+ * This deque prevents null objects from being added.
+ * </p>
+ *
+ * @param <E> the type of elements in this collection
+ * @since 4.5
+ */
+public class CircularDeque<E> extends AbstractCollection<E>
```
belugabehr: This seems like a lot of copy+pase code.  Please just wrap a Java JDK `ArrayDeque`.

https://docs.oracle.com/javase/7/docs/api/java/util/ArrayDeque.html

src/main/java/org/apache/commons/collections4/deque/BlockingDeque.java
```diff
@@ -0,0 +1,525 @@
...
+ * in the deque when storing an element if the decorated deque has a capacity limit.
+ * <p>
+ * This deque exists to support blocking operations for the decorated deque. It is thread
+ * safe.
+ * </p>
+ *
+ * @param <E> the type of elements held in this deque
+ * @since 4.5
+ */
+public class BlockingDeque<E> extends AbstractCollection<E> implements Deque<E>, Serializable {
```
belugabehr: There is already a `BlockingDeque` interface in the Java JDK. 

https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/BlockingDeque.html

PR #114, opened 2019-11-11
Fix Javadoc code example in BulkTest

dota17:
The incorrect variable name will mislead readers.

### DIFF
--- a/src/test/java/org/apache/commons/collections4/BulkTest.java
+++ b/src/test/java/org/apache/commons/collections4/BulkTest.java
@@ -80,11 +80,11 @@ import junit.framework.TestSuite;
  *      }
  *
  *      public BulkTest bulkTestKeySet() {
- *          return new TestSet(makeFullMap().keySet());
+ *          return new SetTest(makeFullMap().keySet());
  *      }
  *
  *      public BulkTest bulkTestEntrySet() {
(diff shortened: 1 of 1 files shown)


garydgregory:
@dota17 
The title of this PR is "Modified the error in javadoc of BulkTest" but this is a code change, not a Javadoc change.

Do you mean to change the Javadoc or the code? Why is the code wrong?

Thank you!
Gary

dota17:
@garydgregory 

The code in the comments of BulkTest.java shows the reader how to define both simple and bulk test method. 

The method bulkTestKeySet() in class HashMapTest wants to define a bulk test method, so it should invoke the SetTest() method which is a constructer of SetTest class rather than the TestSet() method which is not defined. This spelling mistake confused me a lot at first.

Chen

kinow:
I think this is a Javadoc change @garydgregory . At first also thought it was a code change. Then looked at the IDE, and realized it was a change in the example code, in the Javadoc of `BulkTest.java`.

image

The `TestSet` does not exist in the project, but is mentioned also in the Javadocs, just a few lines above.

image

kinow:
@dota17 I've updated the title, hope that's OK.

dota17:
@kinow   Great, thank you!

PR #120, opened 2019-11-26
[COLLECTIONS-737] The test FluentIterableTest.size should be splitted

Prodigysov:
Extracted the first part of FluentIterableTest.size to a sperate test FluentIterableTest.ofNull

Fixes [COLLECTIONS-737](https://issues.apache.org/jira/browse/COLLECTIONS-737)

### DIFF
--- a/src/main/java/org/apache/commons/collections4/IterableUtils.java
+++ b/src/main/java/org/apache/commons/collections4/IterableUtils.java
@@ -804,6 +804,9 @@ public class IterableUtils {
      * @return the number of elements contained in the iterable
      */
     public static int size(final Iterable<?> iterable) {
+        if (iterable == null) {
+            return 0;
+        }
         if (iterable instanceof Collection<?>) {
             return ((Collection<?>) iterable).size();
         }
--- a/src/test/java/org/apache/commons/collections4/FluentIterableTest.java
+++ b/src/test/java/org/apache/commons/collections4/FluentIterableTest.java
@@ -415,13 +415,17 @@ public class FluentIterableTest {
     }
 
     @Test
-    public void size() {
+    public void ofNull() {
         try {
-            FluentIterable.of((Iterable<?>) null).size();
+            FluentIterable.of((Iterable<?>) null);
             fail("expecting NullPointerException");
         } catch (final NullPointerException npe) {
             // expected
         }
+    }
+
+    @Test
+    public void size() {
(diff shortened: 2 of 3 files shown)


src/test/java/org/apache/commons/collections4/FluentIterableTest.java
```diff
@@ -415,13 +415,17 @@ public void isEmpty() {
     }
 
     @Test
-    public void size() {
+    public void ofNull() {
         try {
-            FluentIterable.of((Iterable<?>) null).size();
```
dota17: This test case maybe the same as factoryMethodOf() 
line 125.

Prodigysov: cf [my reply on the JIRA issue](https://issues.apache.org/jira/browse/COLLECTIONS-737?focusedCommentId=16983996&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16983996); I have removed FluentIterableTest.ofNull.

src/main/java/org/apache/commons/collections4/IterableUtils.java
```diff
@@ -804,6 +804,9 @@ public static boolean isEmpty(final Iterable<?> iterable) {
      * @return the number of elements contained in the iterable
      */
     public static int size(final Iterable<?> iterable) {
+        if (iterable == null) {
+            return 0;
+        }
```
kinow: The return statement in this method `return IteratorUtils.size(emptyIteratorIfNull(iterable));` will return `0` anyway. But I think it's harmless to fail quicker, and avoids going through unnecessary steps. :+1:

src/test/java/org/apache/commons/collections4/FluentIterableTest.java
```diff
@@ -416,12 +416,6 @@ public void isEmpty() {
 
     @Test
     public void size() {
-        try {
-            FluentIterable.of((Iterable<?>) null).size();
-            fail("expecting NullPointerException");
-        } catch (final NullPointerException npe) {
-            // expected
-        }
```
kinow: This test was covering the case of `FluentIterable.of(null)`, but there are other tests covering it too. And as pointed by the OP, this method was - according to its name - supposed to test the `.size()` method. :+1:

Prodigysov:
Sounds good, thanks for the review!

PR #121, opened 2019-11-29
[COLLECTIONS-738] Remove the redundant assertNull in IterableUtilsTest.find

Prodigysov:
As described in the JIRA issue [COLLECTIONS-738](https://issues.apache.org/jira/browse/COLLECTIONS-738)

### DIFF
--- a/src/test/java/org/apache/commons/collections4/IterableUtilsTest.java
+++ b/src/test/java/org/apache/commons/collections4/IterableUtilsTest.java
@@ -274,7 +274,7 @@ public class IterableUtilsTest {
         assertTrue(test == null);
         assertNull(IterableUtils.find(null,testPredicate));
         try {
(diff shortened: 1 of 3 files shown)


src/test/java/org/apache/commons/collections4/IterableUtilsTest.java
```diff
@@ -274,7 +274,7 @@ public void find() {
         assertTrue(test == null);
         assertNull(IterableUtils.find(null,testPredicate));
         try {
-            assertNull(IterableUtils.find(iterableA, null));
```
kinow: The `assertNull` here was really confusing. I would expect that it would assert confirming that something **is null**. But instead the `.find` would throw a NPE, that it caught below. i.e. no need to `assertNull` if I understand it right.

PR #122, opened 2019-12-04
Improve MapUtils with the null checks, add Javadoc for the parameter indent and add JUnit for it.

dota17:
Improve MapUtils with the null checks, add Javadoc for the parameter indent and add JUnit for it.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/MapUtils.java
+++ b/src/main/java/org/apache/commons/collections4/MapUtils.java
@@ -34,6 +34,8 @@ import java.util.SortedMap;
 import java.util.TreeMap;
 import java.util.function.BiFunction;
 import java.util.function.Function;
+import java.util.Objects;
+
 
 import org.apache.commons.collections4.map.AbstractMapDecorator;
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/collections4/MapUtils.java
```diff
@@ -1186,11 +1188,12 @@ public static void debugPrint(final PrintStream out, final Object label, final M
      *
      * @param <K> the key type
      * @param <V> the value type
-     * @param map the map to invert, may not be null
+     * @param map the map to invert, must not be null
      * @return a new HashMap containing the inverted data
      * @throws NullPointerException if the map is null
      */
     public static <K, V> Map<V, K> invertMap(final Map<K, V> map) {
+        Objects.requireNonNull(map, "The map must not be null");
```
garydgregory: Please simply use the argument name like "map" or whatever the faulty argument name is.

dota17: Thanks for your advice. I have updated it.

PR #123, opened 2019-12-04
[COLLECTIONS-739] Fix inconsistent @throws comments in DefaultedMap

Prodigysov:
See [COLLECTIONS-739](https://issues.apache.org/jira/browse/COLLECTIONS-739).

### DIFF
--- a/src/main/java/org/apache/commons/collections4/map/DefaultedMap.java
+++ b/src/main/java/org/apache/commons/collections4/map/DefaultedMap.java
@@ -22,6 +22,7 @@ import java.io.ObjectOutputStream;
 import java.io.Serializable;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 
 import org.apache.commons.collections4.Factory;
 import org.apache.commons.collections4.Transformer;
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/collections4/map/DefaultedMap.java
```diff
@@ -105,7 +105,8 @@
      * @param map  the map to decorate, must not be null
      * @param factory  the factory to use to create entries, must not be null
      * @return a new defaulting map
-     * @throws NullPointerException if map or factory is null
+     * @throws NullPointerException if map is null
+     * @throws IllegalArgumentException if factory is null
```
dota17: This is indeed a judgment of NullPointerException. It is recommended to modify IllegalArgumentException to NullPointerException, and to supplement the test cases for it.

Prodigysov: Agreed. I changed to NullPointerException and added test cases for it.

src/main/java/org/apache/commons/collections4/map/DefaultedMap.java
```diff
@@ -110,7 +110,7 @@
      */
     public static <K, V> DefaultedMap<K, V> defaultedMap(final Map<K, V> map, final Factory<? extends V> factory) {
         if (factory == null) {
```
garydgregory: Not quite, to go the NPE route, you should use a liner with `Objects.requireNonNull()`.

Prodigysov: Changed to use `Objects.requireNonNull` (I noticed and followed commit 513ae545ee2b333f62f8ff989e1da161d807d76d)

PR #124, opened 2019-12-04
[COLLECTIONS-740] Add missing @throws comment for SwitchTransformer.switchTransformer.

Prodigysov:
See [COLLECTIONS-740](https://issues.apache.org/jira/browse/COLLECTIONS-740).


src/main/java/org/apache/commons/collections4/functors/SwitchTransformer.java
```diff
@@ -51,6 +51,7 @@
      * @return the <code>chained</code> transformer
      * @throws NullPointerException if array is null
      * @throws NullPointerException if any element in the array is null
+     * @throws IllegalArgumentException if predicate and transformer arrays do not have the same size
```
dota17: maybe it's better to be the same as TransformerUtils line 317
` @throws IllegalArgumentException if the arrays have different sizes`

Prodigysov: Agreed. I fixed this and also fixed some other inconsistencies of the @throw Javadoc on both sides.

PR #136, opened 2020-03-01
COLLECTIONS-754 upgrade to JUnit v5.6.0

nhojpatrick:
Handles https://issues.apache.org/jira/browse/COLLECTIONS-754

Also based upon https://issues.apache.org/jira/browse/COLLECTIONS-753 branch so pr https://github.com/apache/commons-collections/pull/135

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -449,6 +449,18 @@
       <artifactId>junit</artifactId>
       <version>4.13</version>
       <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>org.hamcrest</groupId>
+          <artifactId>hamcrest-core</artifactId>
+        </exclusion>
+      </exclusions>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -445,9 +445,27 @@
...
+        <exclusion>
+          <groupId>org.hamcrest</groupId>
+          <artifactId>hamcrest-core</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.hamcrest</groupId>
+      <artifactId>hamcrest</artifactId>
+      <version>2.2</version>
```
kinow: If this PR gets merged, then #135 doesn't need to be merged I guess? In that case, we could maybe close that other PR and leave just this one?

PR #137, opened 2020-03-06
WIP: CountingBloomFilter

aherbert:
This renames the existing CountingBloomFilter to MapCountingBloomFilter. This makes way for the CountingBloomFilter to be created as an interface to extend BloomFilter.

An implementation using a backing array has been added: ArrayCountingBloomFilter.

The now renamed MapCountingBloomFilter has not been updated. Either it is changed to implement the CountingBloomFilter interface or removed from the API.


src/main/java/org/apache/commons/collections4/bloomfilter/ArrayCountingBloomFilter.java
```diff
@@ -0,0 +1,396 @@
...
+            BitSet.valueOf(other.getBits()).stream().forEach(action);
+        }
+    }
+
+    /**
+     * Apply the action for each index in the hasher.
+     */
+    private void applyAsHasher(final Hasher hasher, final IntConsumer action) {
+        verifyHasher(hasher);
+        toSet(hasher).forEach(i -> action.accept(i));
```
Claudenw: Rather than toSet() how about a filtering consumer.  Something like:

```java
private static class UniqueFilter implements IntConsumer {
        IntConsumer child;
        long[] filter;
        UniqueFilter( IntConsumer child, Shape shape ) {
            this.child=child;
            filter = new long[ BloomFilterIndexer.getLongIndex(shape.getNumberOfBits()) ];
        }
        @Override
        public void accept(int arg0) {
            long target = filter[BloomFilterIndexer.getLongIndex(arg0) ];
            long mask = BloomFilterIndexer.getLongBit(arg0);
            if ( (target & mask) == 0)
            {
                target |= mask;
                child.accept( arg0 );
            }
        }
    }
```
the use 

```java
        UniqueFilter filter = new UniqueFilter( action, shape );
        hasher.getBits(shape).forEachRemaining( filter );
```

src/main/java/org/apache/commons/collections4/bloomfilter/ArrayCountingBloomFilter.java
```diff
@@ -0,0 +1,396 @@
...
+        if (other instanceof ArrayCountingBloomFilter) {
+            verifyShape(other);
+            return contains(((ArrayCountingBloomFilter) other).iterator());
+        }
+
+        // Note:
+        // This currently creates a StaticHasher which stores all the indexes.
+        // It would greatly benefit from direct generation of the index iterator
+        // avoiding the intermediate storage.
+        return contains(other.getHasher());
```
Claudenw: Would this work for the BloomFilter Hasher implementation ?

```java
    public class BloomFilterHasher implements Hasher {
        BloomFilter bf;
        Function<BloomFilter,PrimitiveIterator.OfInt> func;
        
        BloomFilterHasher( BloomFilter bf, Function<BloomFilter,PrimitiveIterator.OfInt> func) {
            this.bf = bf;
            this.func = func;
        }

        @Override
        public OfInt getBits(Shape shape) {
            if (!bf.getShape().equals(shape)) {
                throw new IllegalArgumentException(String.format("Hasher shape (%s) is not the same as shape (%s)",
                    bf.getShape().toString(), shape.toString()));
            }
            return func.apply( bf );
        }

        @Override
        public HashFunctionIdentity getHashFunctionIdentity() {
            return bf.getShape().getHashFunctionIdentity();
        }

        @Override
        public boolean isEmpty() {
            return bf.cardinality() == 0;
        }

    }
```

`BloomFilter.getHasher()` would have to be changed to return a `Hasher` (rather than `StaticHasher`) or perhaps a `UniqueHasher` where `UniqueHasher extends Hasher` and is only a marker to indicate that the values do not contain duplicates.

Claudenw:
All in all this looks like a good change to me.

src/main/java/org/apache/commons/collections4/bloomfilter/ArrayCountingBloomFilter.java
aherbert: This comment is a prelude to the discussion from the mailing list where the getHasher() method is dropped from the BloomFilter in favour of an iterator of the indices. 

Your example requires a function to convert a BloomFilter to an iterator. Who provides that method? Given the data is within the BloomFilter encapsulation requires it to be the BloomFilter to provide the method. So let's just go with the plan to have a BloomFilter provide an iterator. This would replace the getHasher method and its requirement to use a concrete StaticHasher class as the return type and the inefficiency of doing so for a close to capacity filter (the StaticHasher constructor currently passes an iterator through a Set to make indexes unique).

src/main/java/org/apache/commons/collections4/bloomfilter/ArrayCountingBloomFilter.java
aherbert: It would work. Note you would need a +1 on the size of the long[] and you have to write the long bit back to the array:

```java
// Ensure array is big enough
filter = new long[BloomFilterIndexer.getLongIndex(shape.getNumberOfBits()) + 1];

// ...

// Write back the bit index
if ((target & mask) == 0) {
    filter[BloomFilterIndexer.getLongIndex(arg0)] |= mask;
    child.accept(arg0);
}
```

But you are allocating an array of length 1/64 of the number of bits in the filter. We expect a hasher to provide only k-indexes with k being small (e.g. <30).

Certainly there is a potential optimisation to be had by avoiding the use of a Set with boxed integers.

It would be more space efficient to just use an int[] to store up to k-indices. You check each new index using a for loop through those already observed for complexity O(n^2) as search complexity for each index is O(n). In comparison the search complexity of a TreeSet is O(log n) for O(n log n) overall. But with the boxing of the integers and the construction of the tree it would require some performance analysis on typical values of small k to determine which is best.

For now I can mark it as a TODO in the code to investigate using a different data structure to purify the indexes when the expected count of indexes is small/medium/large. I think your class would work for large. Here we require one for small.

Since this may be useful it could be put into a factory to create an appropriate filter for different sizes:
```java
IntConsumer action = ...
int k = ...
// UniqueFilter would ensure the action is called with unique values for
// an expected total of k invocations.
IntConsumer filteredAction = UniqueFilter.forSize(k, action);
```

Claudenw: How about this as an approach:

```java
import java.util.HashSet;
import java.util.Set;
import java.util.function.IntConsumer;
import java.util.function.IntPredicate;

import org.apache.commons.collections4.bloomfilter.hasher.Shape;

public class UniqueFilter implements IntConsumer {
    final IntConsumer child;
    final IntPredicate predicate;

    /**
     * Constructs an IntConsumer that only passes unique values. 
     * @param child the IntConsumer to receive the values.
     * @param shape the shape of the bloom filter assocaited with the values.
     */
    public UniqueFilter( IntConsumer child, Shape shape) {
        this.child=child;
        /* 
         * The type of filter is determined by the median number of bits
         * expected as a proportion of the number of bits.  The assumption
         * is that lower saturations will function more efficiently using 
         * standard sets rather an a collection of bits for each possible value. 
         */
        double median = calculateMedianHamming( shape );
        if ( (median / shape.getNumberOfBits()) > 0.25)
        {
            predicate = new LongFilter( shape );
        } else {
            predicate = new SetFilter( shape );
        }
    }
    
    private double calculateMedianHamming(Shape shape) {
        int k = shape.getNumberOfHashFunctions();
        int n = shape.getNumberOfItems();
        // need a double here for the calculations to work.
        double m = shape.getNumberOfBits();
        
        int kn = k*n;
        
        return kn - m + m* Math.pow((m-1/m),kn);
        
    }
    
    private double estimateMedianHamming(Shape shape) {
        int kn = shape.getNumberOfHashFunctions()*shape.getNumberOfItems();
        int limit = Integer.min(kn, shape.getNumberOfBits());
        int range = limit - shape.getNumberOfHashFunctions();
        int estimate = shape.getNumberOfHashFunctions() + (range/2);
        return estimate;
    }
    
    @Override
    public void accept(int arg0) {
        if (predicate.test( arg0 ))
        {
            child.accept( arg0 );
        }
    }
    
    
    public static class LongFilter implements IntPredicate {
        final long[] filter;

        LongFilter(Shape shape ) {
            filter = new long[ BloomFilterIndexer.getLongIndex(shape.getNumberOfBits())+1 ];
        }
        
        @Override
        public boolean test(int arg0) {
            int idx = BloomFilterIndexer.getLongIndex(arg0);
            long mask = BloomFilterIndexer.getLongBit(arg0);
            if ( (filter[ idx ] & mask) == 0)
            {
                filter[ idx  ] |= mask;
                return true;
            }
            return false;
            
        }
    }
    
    public static class SetFilter implements IntPredicate {
        final Set<Integer> filter;

        SetFilter( Shape shape ) {
            filter = new HashSet<Integer>();
        }
        
        @Override
        public boolean test(int arg0) {
            return filter.add(arg0);
        }
    }
}
```

aherbert: It would work but this is for a case when you are merging possibly duplicate indices for a Bloom filter with a known shape and filled to the expected number of items in the shape. So you are estimating for a different use case where the Hasher is representing a Bloom filter at capacity.

For the use case here it is for the merge of a single item that has been hashed into the filter. So in this case IIUC the Hasher should return only `shape.getNumberOfHashFunctions()` in the iterator.

However there is no mandate to actually do this and the Hasher could return a lot more indexes.

If we require the filtering of indexes which may be duplicated in an optimal way from a Hasher it seems that the Hasher interface should have an estimateSize() function.

There are 2 cases:
- size is known exactly
- size is an estimate

This is dealt with in the Spliterator interface using the methods:

```java
long estimateSize();
long getExactSizeIfKnown();
```

So we could work with this idea:

1. We change the Hasher to return a spliterator and use the JDK framework. For example a StaticHasher can report that it is a known size, a DynamicHasher may not know its size.

2. We add estimateSize and getExactSizeIfKnown to the Hasher interface

I think option 1 is less clutter in the API. The interface can then have a default method to create an iterator from the spliterator if you want to process items one at a time and allow fast fail.

Thus a BloomFilter.merge(Hasher) operation can obtain the spliterator, get the size and then create a filter to make the indexes unique with an optimal data structure. Or it can skip having to filter the indexes if the Hasher reports that its Spliterator is `Spliterator.DISTINCT`.

Changing the Hasher API is for another PR. For now I will abstract out the UniqueFilter concept that accepts a Hasher and IntConsumer as arguments and ensures only unique values are passed to the consumer. This can be used when you want to merge with all the unique bit indexes in a Hasher.

Claudenw: The Hasher may return more than `shape.getNumberOfHashFunctions()` as a Hasher may contain more than one item.

aherbert: OK. I've refactored the extraction of unique indexes from a hasher to a new helper class. This can be optimised later when the Hasher interface is updated to support providing the following information:

- size (number of indexes)
- whether or not the indexes are distinct

This could be satisfied by changing `Hasher.getBits(Shape)` to create a spliterator and not an iterator.

Claudenw:
looks good to me.

PR #139, opened 2020-03-07
Make BloomFilterIndexer and HashFunctionValidator public

Claudenw:
fix for Collections-755


Claudenw:
@aherbert Would you be willing to do this merge.  I am attempting to use this code in several projects and rather than having to recode these methods using the ones from the library makes sense.  They are tested and we know they generate the proper values.

thx

PR #140, opened 2020-03-10
Bloomfilter updates2

aherbert:
Fixing coverage.

Updating the Shape object with more documentation and fixing the hash code implementation.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/bloomfilter/hasher/HashFunctionValidator.java
+++ b/src/main/java/org/apache/commons/collections4/bloomfilter/hasher/HashFunctionValidator.java
@@ -16,12 +16,42 @@
  */
 package org.apache.commons.collections4.bloomfilter.hasher;
 
+import java.util.Locale;
+import java.util.Objects;
+
 /**
  * Contains validation for hash functions.
  */
 final class HashFunctionValidator {
     /** Do not instantiate. */
     private HashFunctionValidator() {}
+
+    /**
+     * Generates a hash code for the identity of the hash function. The hash code is
+     * generated using the same properties as those tested in
+     * {@link #areEqual(HashFunctionIdentity, HashFunctionIdentity)}, that is the
+     * signedness, process type and name. The name is not case specific and is converted
+     * to lower-case using the {@link Locale#ROOT root locale}.
+     *
+     * <p>The generated value is suitable for use in generation of a hash code that satisfies
+     * the contract of {@link Object#hashCode()} if the {@link Object#equals(Object)} method
+     * is implemented using {@link #areEqual(HashFunctionIdentity, HashFunctionIdentity)}. That
+     * is two objects considered equal will have the same hash code.
+     *
+     * <p>If the hash function identity is a field within a larger object the generated hash code
+     * should be incorporated into the entire hash, for example using
+     * {@link Objects#hash(Object...)}.
+     *
+     * @param a hash function.
+     * @return hash code
+     * @see String#toLowerCase(Locale)
+     * @see Locale#ROOT
+     */
+    static int hash(HashFunctionIdentity a) {
+        return Objects.hash(a.getSignedness(),
+                            a.getProcessType(),
+                            a.getName().toLowerCase(Locale.ROOT));
+    }
 
     /**
      * Compares the identity of the two hash functions. The functions are considered
--- a/src/main/java/org/apache/commons/collections4/bloomfilter/hasher/Shape.java
+++ b/src/main/java/org/apache/commons/collections4/bloomfilter/hasher/Shape.java
@@ -28,11 +28,11 @@ import java.util.Objects;
  * <h2>Interrelatedness of values</h2>
  *
  * <dl> <dt>Number of Items ({@code n})</dt>
- * <dd>{@code n = ceil(m / (-k / log(1 - exp(log(p) / k))))}</dd> <dt>Probability of
+ * <dd>{@code n = ceil(m / (-k / ln(1 - exp(ln(p) / k))))}</dd> <dt>Probability of
  * False Positives ({@code p})</dt> <dd>{@code p = pow(1 - exp(-k / (m / n)), k)}</dd> <dt>Number
  * of Bits ({@code m})</dt>
- * <dd>{@code m = ceil((n * log(p)) / log(1 / pow(2, log(2))))}</dd> <dt>Number of
- * Functions ({@code k})</dt> <dd>{@code k = round((m / n) * log(2))}</dd> </dl>
+ * <dd>{@code m = ceil((n * ln(p)) / ln(1 / pow(2, ln(2))))}</dd> <dt>Number of
+ * Functions ({@code k})</dt> <dd>{@code k = round((m / n) * ln(2))}</dd> </dl>
  *
  * <h2>Comparisons</h2> <p> For purposes of equality checking and hashCode
  * calculations a {@code Shape} is defined by the hashing function identity, the number of
@@ -48,14 +48,14 @@ public final class Shape {
     /**
      * The natural logarithm of 2. Used in several calculations. Approximately 0.693147180559945.
      */
-    private static final double LOG_OF_2 = Math.log(2.0);
+    private static final double LN_2 = Math.log(2.0);
 
     /**
-     * log(1 / 2^log(2)). Used in calculating the number of bits. Approximately -0.480453013918201.
+     * ln(1 / 2^ln(2)). Used in calculating the number of bits. Approximately -0.480453013918201.
      *
-     * <p>log(1 / 2^log(2)) = log(1) - log(2^log(2)) = -log(2) * log(2)
+     * <p>ln(1 / 2^ln(2)) = ln(1) - ln(2^ln(2)) = -ln(2) * ln(2)
      */
-    private static final double DENOMINATOR = -LOG_OF_2 * LOG_OF_2;
+    private static final double DENOMINATOR = -LN_2 * LN_2;
 
     /**
      * Number of items in the filter ({@code n}).
@@ -87,7 +87,7 @@ public final class Shape {
      * specified number of bits ({@code m}) and hash functions ({@code k}).
      *
      * <p>The number of items ({@code n}) to be stored in the filter is computed.
-     * <pre>n = ceil(m / (-k / log(1 - exp(log(p) / k))))</pre>
+     * <pre>n = ceil(m / (-k / ln(1 - exp(ln(p) / k))))</pre>
      *
      * <p>The actual probability will be approximately equal to the
      * desired probability but will be dependent upon the calculated Bloom filter capacity
@@ -112,7 +112,7 @@ public final class Shape {
         this.numberOfHashFunctions = checkNumberOfHashFunctions(numberOfHashFunctions);
 
         // Number of items (n):
-        // n = ceil(m / (-k / log(1 - exp(log(p) / k))))
+        // n = ceil(m / (-k / ln(1 - exp(ln(p) / k))))
         final double n = Math.ceil(numberOfBits /
             (-numberOfHashFunctions / Math.log(1 - Math.exp(Math.log(probability) / numberOfHashFunctions))));
 
@@ -137,10 +137,10 @@ public final class Shape {
      * desired false-positive probability ({@code p}).
      *
      * <p>The number of bits ({@code m}) for the filter is computed.
-     * <pre>m = ceil(n * log(p) / log(1 / 2^log(2)))</pre>
+     * <pre>m = ceil(n * ln(p) / ln(1 / 2^ln(2)))</pre>
      *
      * <p>The optimal number of hash functions ({@code k}) is computed.
-     * <pre>k = round((m / n) * log(2))</pre>
+     * <pre>k = round((m / n) * ln(2))</pre>
      *
      * <p>The actual probability will be approximately equal to the
      * desired probability but will be dependent upon the calculated number of bits and hash
@@ -178,7 +178,7 @@ public final class Shape {
      * bits ({@code m}).
      *
      * <p>The optimal number of hash functions ({@code k}) is computed.
-     * <pre>k = round((m / n) * log(2))</pre>
+     * <pre>k = round((m / n) * ln(2))</pre>
      *
      * <p>The false-positive probability is computed using the number of items, bits and hash
      * functions. An exception is raised if this is greater than or equal to 1 (i.e. the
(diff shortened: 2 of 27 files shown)


src/main/java/org/apache/commons/collections4/bloomfilter/hasher/Shape.java
```diff
@@ -43,30 +43,32 @@
...
 
     /**
-     * The natural logarithm of 2. Used in several calculations. approx 0.693147180
+     * The natural logarithm of 2. Used in several calculations. Approximately 0.693147180559945.
      */
     private static final double LOG_OF_2 = Math.log(2.0);
 
     /**
-     * 1 / 2^log(2) approx -0.090619058. Used in calculating the number of bits.
+     * log(1 / 2^log(2)). Used in calculating the number of bits. Approximately -0.480453013918201.
```
Claudenw: I ahd an error in my earlier comments the log(1/ 2^log(2)) is approx. −0.090619058

src/main/java/org/apache/commons/collections4/bloomfilter/hasher/Shape.java
```diff
@@ -81,34 +83,36 @@
...
+     * (number of items). An exception is raised if this is greater than or equal to 1 (i.e. the
+     * shape is invalid for use as a Bloom filter).
+     *
+     * @param hashFunctionIdentity The identity of the hash function this shape uses
+     * @param probability The desired false-positive probability in the range {@code (0, 1)}
+     * @param numberOfBits The number of bits in the filter
+     * @param numberOfHashFunctions The number of hash functions in the filter
+     * @throws NullPointerException if the hash function identity is null
+     * @throws IllegalArgumentException if the desired probability is not in the range {@code (0, 1)}
+     * @throws IllegalArgumentException if the number of bits is not above 8
```
Claudenw: should be not >= 8

src/main/java/org/apache/commons/collections4/bloomfilter/hasher/Shape.java
```diff
@@ -124,103 +128,188 @@ public Shape(final HashFunctionIdentity hashFunctionIdentity, final double proba
...
+     * <p>The false-positive probability is computed using the number of items, bits and hash
+     * functions. An exception is raised if this is greater than or equal to 1 (i.e. the
+     * shape is invalid for use as a Bloom filter).
+     *
+     * @param hashFunctionIdentity The identity of the hash function this shape uses
+     * @param numberOfItems Number of items to be placed in the filter
+     * @param numberOfBits The number of bits in the filter
+     * @throws NullPointerException if the hash function identity is null
+     * @throws IllegalArgumentException if the number of items is not strictly positive
+     * @throws IllegalArgumentException if the number of bits is not above 8
```
Claudenw: again >= 8

src/main/java/org/apache/commons/collections4/bloomfilter/hasher/Shape.java
```diff
@@ -124,103 +128,188 @@ public Shape(final HashFunctionIdentity hashFunctionIdentity, final double proba
...
-     * @param hashFunctionIdentity The HashFunctionIdentity of the hash function this shape uses.
-     * @param numberOfItems Number of items to be placed in the filter.
+     * @param hashFunctionIdentity The identity of the hash function this shape uses
+     * @param numberOfItems Number of items to be placed in the filter
      * @param numberOfBits The number of bits in the filter.
-     * @param numberOfHashFunctions The number of hash functions in the filter.
+     * @param numberOfHashFunctions The number of hash functions in the filter
+     * @throws NullPointerException if the hash function identity is null
+     * @throws IllegalArgumentException if the number of items is not strictly positive
+     * @throws IllegalArgumentException if the number of bits is not above 8
```
Claudenw: >= 8

src/main/java/org/apache/commons/collections4/bloomfilter/hasher/Shape.java
```diff
@@ -124,103 +128,188 @@ public Shape(final HashFunctionIdentity hashFunctionIdentity, final double proba
...
+     *
+     * <p>This function is used to verify that the dynamically calculated probability for the
+     * Shape is in the valid range 0 to 1 exclusive. This need only be performed once upon
+     * construction.
+     *
+     * @param probability the probability
+     * @throws IllegalArgumentException if the calculated probability is not below 1
+     */
+    private static void checkCalculatedProbability(final double probability) {
+        // We do not need to check for p < = since we only allow positive values for
```
Claudenw: should be " p <= 0"

src/main/java/org/apache/commons/collections4/bloomfilter/hasher/Shape.java
```diff
@@ -312,32 +402,18 @@ public int getNumberOfItems() {
...
-     * numberOfItems, numberofBits and numberOfHashFunctions. This is a method so that
-     * the calculation is consistent across all constructors.
+     * Calculates the probability of false positives ({@code p}) given
+     * numberOfItems ({@code n}), numberOfBits ({@code m}) and numberOfHashFunctions ({@code k}).
+     * <pre>p = pow(1 - exp(-k / (m / n)), k)</pre>
+     *
+     * <p>This is the probability that a Bloom filter will return true for the presence of an item
+     * when it does not contain the item.
      *
      * @return the probability of collision.
```
Claudenw: I guess "collision" should be "false positive"

src/main/java/org/apache/commons/collections4/bloomfilter/hasher/function/Signatures.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.apache.commons.collections4.bloomfilter.hasher.HashFunction;
+import org.apache.commons.collections4.bloomfilter.hasher.HashFunctionIdentity;
+
+/**
+ * Allow computation of HashFunction signatures.
+ * @since 4.5
+ */
+final class Signatures {
+
```
Claudenw: This needs to be public so that 3rd party developers can use if for their HashFunctions

src/test/java/org/apache/commons/collections4/bloomfilter/hasher/function/AbstractHashFunctionTest.java
```diff
@@ -0,0 +1,50 @@
...
+     */
+    @Test
+    public void signatureTest() {
+        final HashFunction hf = createHashFunction();
+        final long expected = hf.apply(HashFunctionIdentity.prepareSignatureBuffer(hf), 0);
+        assertEquals(expected, hf.getSignature());
+        // Should be repeatable
+        final long expected2 = hf.apply(HashFunctionIdentity.prepareSignatureBuffer(hf), 0);
+        assertEquals(expected, expected2);
+        assertEquals("Apache Commons Collections", hf.getProvider());
```
Claudenw: How does a 3rd party test signatures?  I suppose they can recreate the method or we could make "Apache Commons Collections" the default value for a variable that can be set for all the tests.

src/main/java/org/apache/commons/collections4/bloomfilter/hasher/Shape.java
aherbert: Are you saying that the value is now wrong or right? I make it -0.48. Or am I doing something wrong with my calculator?

src/main/java/org/apache/commons/collections4/bloomfilter/hasher/Shape.java
aherbert: Is it the probability of false positives when the filter is at capacity (contains n items)?

src/main/java/org/apache/commons/collections4/bloomfilter/hasher/function/Signatures.java
aherbert: I don't think this necessary. It just clutters the public API with something trivial. The signature is defined in the javadoc for HashFunction. A developer can write this themselves. This is an internal function to ensure consistency.

If public it should be promoted to the package above (hasher) where the HashFunction is defined.

src/test/java/org/apache/commons/collections4/bloomfilter/hasher/function/AbstractHashFunctionTest.java
aherbert: However they want. I don't think we expect them to extend our test classes.

src/main/java/org/apache/commons/collections4/bloomfilter/hasher/HashFunctionValidator.java
```diff
@@ -16,13 +16,43 @@
  */
 package org.apache.commons.collections4.bloomfilter.hasher;
 
+import java.util.Locale;
```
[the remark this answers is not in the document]
aherbert: I think we should focus on the core public API first. Helper classes can be exposed later when the core functionality is more fixed. This PR is addressing general tidy up before changes to the BloomFilter interface.

src/main/java/org/apache/commons/collections4/bloomfilter/hasher/Shape.java
aherbert: If you think that it is clearer I'll change it.

Why 8-bits anyway? The limit seems arbitrary to me. If based on a byte then why not an int or long?

aherbert: Maybe we should swap these exception conditions around to those that actually create the exception:
```
if the number of items is not strictly positive
if the number of hash function is not strictly positive
if the number of bits is not >= 8
if the calculated probability is not below 1
if the calculated number of hash function is below 1

if the probability is not in the range (0, 1)
```
Becomes
```
if the number of items is < 1
if the number of hash function is < 1
if the number of bits is < 8
if the calculated probability is >= 1
if the calculated number of hash function is < 1

if the probability is <= 0 or >= 1
```
I think the last one is clearer as a negation (i.e the original).

src/main/java/org/apache/commons/collections4/bloomfilter/hasher/Shape.java
aherbert: I think -0.48 is correct:

[Wikipedia - Bloom Filter - Optimal number of hash functions](https://en.wikipedia.org/wiki/Bloom_filter#Optimal_number_of_hash_functions)

```
m = - n ln(p) / ln(2)^2
```
So the denominator is -ln(2)^2 = -0.48.

src/main/java/org/apache/commons/collections4/bloomfilter/hasher/Shape.java
Claudenw: I like the change in conditions statement.

8 bits is really just a recommendation.  I guess we can remove the checks.  The issue arises when we calculate the other fields.  The actual minimum is 4 (if you have 1 item and 7 if you have 2).  It falls out in the hash calculations, or the probability calculations

src/main/java/org/apache/commons/collections4/bloomfilter/hasher/Shape.java
Claudenw: my calculator entry: log( 1 ÷ (2^log(2)))
my calculator result: −0.090619058
my calculator entry: -log(2)*log(2)
my calculator result: −0.090619058

src/main/java/org/apache/commons/collections4/bloomfilter/hasher/Shape.java
Claudenw: Yes probability of false positive at a saturation of 1.

src/main/java/org/apache/commons/collections4/bloomfilter/hasher/Shape.java
aherbert: You are using log10. It should be the natural logarithm (ln).

src/main/java/org/apache/commons/collections4/bloomfilter/hasher/Shape.java
aherbert: So from that I take it if we change the check to a minimum of 1 the rest of the checks will throw an exception but the error message is not linked to what the root cause of the problem was.

aherbert: OK. I have changed the check to number of bits must be above 0 to investigate the computations. This
```java
// numberOfBits = 1
// numberOfHashFunctions = 1;
new Shape(testFunction, Math.nextDown(1.0), 1, 1);
```
Creates a Shape with a number of items as 37 and a probability of 0.9999999999999999.

Using a number of bits of 4 creates a Shape with a number of items as 147 and a probability of 0.9999999999999999.

Using a number of bits of 8 creates a Shape with a number of items as 294 and a probability of 0.9999999999999999.

This is a bad case that will create a non-sense filter shape but the question is do we allow this? 

It may be better to throw an exception when the calculated number of items is above the number of bits. If each bit represents a single item then it is impossible to filter more items than number of bits.

src/main/java/org/apache/commons/collections4/bloomfilter/hasher/Shape.java
Claudenw: DOH!

src/main/java/org/apache/commons/collections4/bloomfilter/hasher/Shape.java
Claudenw: I used to think that we should filter out unreasonable shapes (thus the bits >= 8).  But now I think that we should just check the results to ensure that they do not fall outside of actual requirements.  Specifically:

if the number of items is < 1
if the number of hash function is < 1
if the calculated number of hash function is < 1
if the probability is <= 0 or >= 1

The reasoning here is that the values outside of those values cause the functions to fail.  Anything else is valid and while we may look at it now and say that it is non-sense it may be reasonable in some situation we do not foresee.

Claudenw:
I think that if the javadoc changes are made: `log -> ln`
and the lower limit on bits is dropped to 1

We can release merge this.

aherbert:
@Claudenw I have updated the check on the number of bits to be `< 1`.

But it still seems to me you can create a shape that you would never want:
```java
// necessary
numberOfItems < numberOfBits
// sensible
numberOfHashFunctions < numberOfBits
```

I think the first check may make the test on the calculated probability redundant. Looking at the unit tests you have to construct a Shape with number of items > number of bits to create a bad probability.

I think the limitation to only allow shape creation when the probability computation is in the range (0, 1) when the filter is filled to exactly the expected number of items is strange. IMO it would be better to throw an an exception if the filter cannot be created with an actual probability at saturation close to the desired probability (e.g. within 1e-3). But then you have to define what is close. So this check perhaps should be dropped, and may be redundant if number of items is less than number of bits.

I am also not sold on the the `getProbability()` function. It is limited to `n` set to the expected number of items. But you may want to know the probability of your Bloom filter that you have just filled with x items. So perhaps we should either have a static method in Shape to compute each of the standard functions defined in the header (i.e. n, m, p, k) and/or an instance level `getProbability(n)` that uses the fixed `m` and `k`.

Claudenw:
Shape.getProbability() is intended to return the probability as defined for the shape.  As you state when it has a saturation of 1.  

If you want to just store `m` and `k` then I would like to see methods defined to estimate `n` and `p`

`n = ceil( m * ln2 / k )`

The importance of Shape is that it tells the application what the Bloom filters are intended to look like.  The saturation of individual filters may vary but it provides a baseline from which to estimate.  If you think about building a collections of filters with a gating filter.  You can start the shape of the gating filter with the inverse of the probability of the shape of the filters that are to be stored.  But to get there you have to have the `n` for the shape.

PR #144, opened 2020-03-27
[COLLECTIONS-757] Update the JavaDoc of the BidiMap

dota17:


src/main/java/org/apache/commons/collections4/bidimap/AbstractDualBidiMap.java
```diff
@@ -497,7 +494,7 @@ public boolean remove(final Object key) {
         protected boolean canRemove = false;
 
         /**
-         * Constructor.
+         * Constructor a KeySetIterator of the BidiMap
```
kinow: "Constructor of a" or "Constructs a" would sound better I think.

However, I think just "Constructor" is much simpler/easier to maintain (ditto for the other similar changes in this PR)

src/main/java/org/apache/commons/collections4/bidimap/AbstractDualBidiMap.java
```diff
@@ -620,7 +617,7 @@ public void remove() {
         private static final long serialVersionUID = 4040410962603292348L;
 
         /**
-         * Constructs a new view of the BidiMap.
+         * Constructs a new view of the BidiMap
```
kinow: I think we have some other statements in the code base that end in a period. Any specific reason to remove them? If not, I think it would be simpler to just leave them as-is.

src/main/java/org/apache/commons/collections4/bidimap/AbstractDualBidiMap.java
```diff
@@ -452,7 +449,7 @@ public void clear() {
         private static final long serialVersionUID = -7107935777385040694L;
 
         /**
-         * Constructs a new view of the BidiMap.
+         * Constructs a new KeySet of the BidiMap
```
kinow: I think the view it referred to was more in an abstract way, but I think it should be fine to be specific and mention the `KeySet` type here.

src/main/java/org/apache/commons/collections4/bidimap/DualLinkedHashBidiMap.java
```diff
@@ -26,9 +26,9 @@
 import org.apache.commons.collections4.BidiMap;
 
 /**
- * Implementation of {@code BidiMap} that uses two {@code LinkedHashMap} instances.
+ * Implementation of {@link BidiMap} that uses two {@link LinkedHashMap} instances.
```
kinow: This might have been intentional? Some people prefer to avoid adding too many links in the Javadoc (forgot where I saw a presentation about that, but it was in some guide for writing Javadocs where the author mentioned to keep the number of links low, so users are not distracted).

src/main/java/org/apache/commons/collections4/bidimap/AbstractDualBidiMap.java
garydgregory: -1: A sentence ends with a period.

src/main/java/org/apache/commons/collections4/bidimap/AbstractDualBidiMap.java
garydgregory: "Constructor." or "Constructs a new instance." is a good place holder comment, otherwise, you can describe what kind of initialization takes place.

dota17:
Thanks for the response @kinow @garydgregory .
Keep the code instead of the link ?

src/main/java/org/apache/commons/collections4/bidimap/DualLinkedHashBidiMap.java
garydgregory: That's silly IMO, these links are fine.

src/main/java/org/apache/commons/collections4/bidimap/TreeBidiMap.java
```diff
@@ -782,7 +782,7 @@ private static boolean isBlack(final Node<?, ?> node, final DataElement dataElem
     }
 
     /**
-     * force a node (if it exists) red
+     * Force a node (if it exists) red.
```
garydgregory, line 785: If a method starts with "make...", it usually reads better to start the Javadoc with "Makes..." IMO, It's confusing to have the Javadoc be different, otherwise you could say that all method that start with "set..." also "forces" a value.

src/main/java/org/apache/commons/collections4/bidimap/TreeBidiMap.java
```diff
@@ -1130,7 +1130,7 @@ private void doRedBlackDeleteFixup(final Node<K, V> replacementNode, final DataE
     }
 
     /**
-     * swap two nodes (except for their content), taking care of
+     * Swap two nodes (except for their content), taking care of
```
garydgregory, line 1133: "Swap" -> "Swaps"

src/main/java/org/apache/commons/collections4/bidimap/TreeBidiMap.java
```diff
@@ -1230,8 +1230,8 @@ private void swapPosition(final Node<K, V> x, final Node<K, V> y, final DataElem
     }
 
     /**
-     * check if an object is fit to be proper input ... has to be
-     * Comparable and non-null
+     * Check if an object is fit to be proper input ... has to be
```
garydgregory, line 1233: "Check" -> "Checks"

src/main/java/org/apache/commons/collections4/bidimap/DualLinkedHashBidiMap.java
kinow: I found the document I mentioned earlier. It was not exactly a presentation, but the Oracle docs for javadocs: https://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#styleguide (where it says "Use in-line links economically"). There's also a similar style guide in UI/UX, to avoid cluttering user interfaces. However, I'm fine whichever way :+1:

PR #148, opened 2020-04-18
Change line separators from CRLF to LF.

Isira-Seneviratne:
This fixes the build issue caused by the line separators for these files being CRLF instead of LF as expected in the style check.


Isira-Seneviratne:
Closing as this is a duplicate.

PR #149, opened 2020-04-18
[COLLECTIONS-760]: Add tests for MapUtils

Isira-Seneviratne:


src/test/java/org/apache/commons/collections4/MapUtilsTest.java
```diff
@@ -56,6 +58,7 @@
  */
 @SuppressWarnings("boxing")
 public class MapUtilsTest extends AbstractAvailableLocalesTest {
+    private static final String THREE = "Three";
```
kinow, line 61: Hi @Isira-Seneviratne should we have constants for "Two", and "key"? As this is a test class, I think we can have it either way, with or without constants, as the class size is not huge as well. Up to you :+1:

Isira-Seneviratne: Assigning "Three" to a constant variable was recommended by SonarLint. It did not provide the same suggestion for "Two" and "key" as those are shorter strings.

Isira-Seneviratne: On that note, maybe a configuration for [SonarQube](https://www.sonarqube.org/) code analysis could be added?

kinow: I believe we were using it at one point. ASF Infra has or had a SonarQube instance running. Now I think there are projects in the ASF using SonarQube, though I am not sure whether that's SonarQube hosted within ASF, or the SonarCloud cloud service. Feel free to reach out via the dev-list and ask it there (as not all Apache Commons developers are subscribed to GitHub, the mailing lists are the best way to discuss things like this :)

PR #160, opened 2020-05-10
add Simple bloom filter

Claudenw:
This pull requests adds a generic `SimpleBloomFilter<T>`.  The user provides a `Function<T,SimpleBuilder>` implementation.  The class adds `bloomFilter.merge( T )` and `bloomFilter.contains( T )`.

Additional changes were made to Hasher.Builder to use generics:  `Hasher.Builder<H extends Hasher, B extends Builder>`.  This allows the default methods in the interface to work correctly when a minimal specific implementation is provided. 

Also added additional `with()` methods to `Hasher.Builder<H extends Hasher, B extends Builder>` to accept primitives (int, long, double, float, short, char) as well as BigInteger and BigDecimal.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/bloomfilter/BitSetBloomFilter.java
+++ b/src/main/java/org/apache/commons/collections4/bloomfilter/BitSetBloomFilter.java
@@ -25,7 +25,7 @@ import org.apache.commons.collections4.bloomfilter.hasher.Shape;
 import org.apache.commons.collections4.bloomfilter.hasher.StaticHasher;
 
 /**
- * A bloom filter using a Java BitSet to track enabled bits. This is a standard
+ * A bloom filter uses a Java BitSet to track enabled bits. This is a standard
  * implementation and should work well for most Bloom filters.
  * @since 4.5
  */
--- a/src/main/java/org/apache/commons/collections4/bloomfilter/SimpleBloomFilter.java
+++ b/src/main/java/org/apache/commons/collections4/bloomfilter/SimpleBloomFilter.java
@@ -78,7 +78,7 @@ public class SimpleBloomFilter<T> extends BitSetBloomFilter implements BloomFilt
     /**
      * Constructs a SimpleBloomFilter from the shape and function. This constructor
      * creates an empty Bloom filter.
-     * </p><p>
+     * <p>
      * If the object {@code T} is to be considered as a
      * single item in the filter then function must create the {@code SimpleBuilder}
      * and only call a single {@code with()} method. In this use case the
--- a/src/main/java/org/apache/commons/collections4/bloomfilter/SimpleBloomFilter.java
+++ b/src/main/java/org/apache/commons/collections4/bloomfilter/SimpleBloomFilter.java
@@ -33,30 +33,38 @@ import org.apache.commons.collections4.bloomfilter.hasher.function.Murmur128x64C
 public class SimpleBloomFilter<T> extends BitSetBloomFilter implements BloomFilter {
 
     /**
-     * The function that converts the instance of T to the SimpleBuilder. <p> If the
+     * The function that converts the instance of T to the SimpleBuilder.
+     * <p>
+     * If the
      * object T is to be considered as a single item in the filter then function
      * must create the {@code SimpleBuilder} and only call a single {@code with()}
-     * method.</p> <p> If the object T is to be considered as several items then the
+     * method.
+     * </p> <p>
+     * If the object T is to be considered as several items then the
      * function must create the {@code SimpleBuilder} and call the {@code with()}
      * method once for each item.</p>
      */
     private Function<T, SimpleBuilder> func;
 
     /**
-     * Constructs a SimpleBloomFilter from the shape and function.
+     * Constructs a SimpleBloomFilter from the shape and function containing the
+     * items specified in the hasher.
      *
      * <p> The function {@code func} converts the instance of {@code T} into the
-     * SimpleBuilder. </p><p> If the object {@code T} is to be considered as a
+     * SimpleBuilder.
+     * </p><p>
+     * If the object {@code T} is to be considered as a
      * single item in the filter then function must create the {@code SimpleBuilder}
      * and only call a single {@code with()} method. In this use case the
      * {@code Shape} is generally configured with the number of items equal to the
-     * expected number of {@code T} to be added to the Bloom filter.</p> <p> If the
-     * object {@code T} is to be considered as several items then the function must
+     * expected number of {@code T} to be added to the Bloom filter.
+     * </p> <p>
+     * If the object {@code T} is to be considered as several items then the function must
      * create the {@code SimpleBuilder} and call the {@code with()} method once for
      * each item. In this use case the {@code Shape} is generally configured with
      * the number of items equal to the expected number of {@code T} multiplied by
-     * the number of times {@code with()} is called.</p> </p>
-     *
+     * the number of times {@code with()} is called.
+     * </p>
      * @param hasher the Hasher to use.
      * @param shape  the Shape of the Bloom filter.
      * @param func   a Function to convert T to a SimpleBuilder.
@@ -70,6 +78,20 @@ public class SimpleBloomFilter<T> extends BitSetBloomFilter implements BloomFilt
     /**
      * Constructs a SimpleBloomFilter from the shape and function. This constructor
      * creates an empty Bloom filter.
+     * </p><p>
+     * If the object {@code T} is to be considered as a
+     * single item in the filter then function must create the {@code SimpleBuilder}
+     * and only call a single {@code with()} method. In this use case the
+     * {@code Shape} is generally configured with the number of items equal to the
+     * expected number of {@code T} to be added to the Bloom filter.
+     * </p> <p>
+     * If the object {@code T} is to be considered as several items then the function must
+     * create the {@code SimpleBuilder} and call the {@code with()} method once for
+     * each item. In this use case the {@code Shape} is generally configured with
+     * the number of items equal to the expected number of {@code T} multiplied by
+     * the number of times {@code with()} is called.
+     * </p>
+
      * @param shape the Shape of the Bloom filter.
      * @param func  a Function to convert T to a SimpleBuilder.
      * @see #func
@@ -85,7 +107,7 @@ public class SimpleBloomFilter<T> extends BitSetBloomFilter implements BloomFilt
      * <p>Note: This method should return {@code true} even if no additional bit
      * indexes were enabled. A {@code false} result indicates that this filter is
      * not ensured to contain the {@code data}.
-     *
+     * </p>
      * @param data the data to merge.
      * @return true if the merge was successful
      * @throws IllegalArgumentException if the shape of the other filter does not
--- a/src/main/java/org/apache/commons/collections4/bloomfilter/SimpleBloomFilter.java
+++ b/src/main/java/org/apache/commons/collections4/bloomfilter/SimpleBloomFilter.java
@@ -33,24 +33,34 @@ import org.apache.commons.collections4.bloomfilter.hasher.function.Murmur128x64C
 public class SimpleBloomFilter<T> extends BitSetBloomFilter implements BloomFilter {
 
     /**
-     * The function that converts the instance of T to the SimpleBuilder.
-     * <p>
-     * If the object T is to be considered as a single item in the filter then
-     * function must create the {@code SimpleBuilder} and only call a single {@code with()}
-     * method.</p>
-     * <p>
-     * If the object T is to be considered as several items then the function must
-     * create the {@code SimpleBuilder} and call the {@code with()} method once for each item.</p>
+     * The function that converts the instance of T to the SimpleBuilder. <p> If the
+     * object T is to be considered as a single item in the filter then function
+     * must create the {@code SimpleBuilder} and only call a single {@code with()}
+     * method.</p> <p> If the object T is to be considered as several items then the
+     * function must create the {@code SimpleBuilder} and call the {@code with()}
+     * method once for each item.</p>
      */
     private Function<T, SimpleBuilder> func;
 
     /**
-     * Constructs a SimpleBloomFilter from the shape and function.  This constructor
-     * creates an empty Bloom filter.
+     * Constructs a SimpleBloomFilter from the shape and function.
+     *
+     * <p> The function {@code func} converts the instance of {@code T} into the
+     * SimpleBuilder. </p><p> If the object {@code T} is to be considered as a
+     * single item in the filter then function must create the {@code SimpleBuilder}
+     * and only call a single {@code with()} method. In this use case the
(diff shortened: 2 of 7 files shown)


src/main/java/org/apache/commons/collections4/bloomfilter/SimpleBloomFilter.java
```diff
@@ -0,0 +1,127 @@
...
+     * Returns {@code true} if this filter contains the object.
+     * Specifically this returns {@code true} if this filter is enabled for all bit indexes
+     * identified by the hashing of the object {@code data}.
+     *
+     * @param data the data to check for.
+     * @return true if this filter is enabled for all bits specified by the data object.
+     */
+    public boolean contains( T data ) {
+        return this.contains( this.func.apply( data ).build() );
+    }
```
kinow, line 113: The `merge` and `contains` functions look much simpler for the caller user :+1:

src/main/java/org/apache/commons/collections4/bloomfilter/SimpleBloomFilter.java
```diff
@@ -0,0 +1,127 @@
...
+
+    /**
+     * Constructs a SimpleBloomFilter from the shape, function and a data object.
+     * This constructor creates an Bloom filter populated with the data from the
+     * {@code data} parameter.
+     * @param shape the Shape of the Bloom filter.
+     * @param func a Function to convert T to a SimpleBuilder.
+     * @param data the data object to populate the filter with.
+     * @see #func
```
kinow, line 79: #TodayILeaned. Didn't know we could link with the object :+1:

src/test/java/org/apache/commons/collections4/bloomfilter/hasher/HasherBuilderTest.java
```diff
@@ -108,4 +110,156 @@ static String getExtendedString() {
...
+        }
+    }
+
+    /**
+     * Test that adding a BigInteger into the hasher works correctly
+     */
+    @Test
+    public void withBigIngeterTest() {
+        TestBuilder builder = new TestBuilder();
+        BigInteger[] values = { BigInteger.ZERO, BigInteger.ONE, BigInteger.TWO, BigInteger.TEN,
```
kinow, line 225: @Claudenw it looks like `.TWO` is private in Java 8, only `ONE, `TEN`, `ZERO` are available: https://docs.oracle.com/javase/8/docs/api/java/math/BigInteger.html

There's an error in the Java 8 build in Travis.

garydgregory:
Odd error on Java 8:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:testCompile (default-testCompile) on project commons-collections4: Compilation failure
[ERROR] /home/travis/build/apache/commons-collections/src/test/java/org/apache/commons/collections4/bloomfilter/hasher/HasherBuilderTest.java:[225,76] TWO has private access in java.math.BigInteger
```
See https://travis-ci.org/github/apache/commons-collections/jobs/685354919

Claudenw:
Fixes COLLECTIONS-762

src/main/java/org/apache/commons/collections4/bloomfilter/hasher/Hasher.java
```diff
@@ -103,6 +108,95 @@ default Builder withUnencoded(CharSequence item) {
...
+        /**
+         * Adds a BigDecimal into the hasher.
+         * The scale of the BigDecimal is placed in the buffer followed by the
+         * byte array for the unscaled value.
+         * @param data the BigDecimal to add.
+         * @return a reference to this object
+         */
+        default B with(BigDecimal data) {
+            byte[] value = data.unscaledValue().toByteArray();
+            return with( ByteBuffer.allocate( Integer.BYTES+value.length )
```
dota17, line 196: return with( ByteBuffer.allocate( Integer.BYTES + value.length )

src/test/java/org/apache/commons/collections4/bloomfilter/SimpleBloomFilterMultipleItemTest.java
```diff
@@ -0,0 +1,104 @@
...
+    @Test
+    public void containsTTest() {
+        SimpleBloomFilter<TypedObject> bloomFilter = createEmptyFilter( shape );
+        bloomFilter.merge( new TypedObject() );
+
+        TypedObject t = new TypedObject();
+
+        Assert.assertTrue( bloomFilter.contains(t));
+
+        t.hello="hola";
```
dota17, line 92: t.hello = "hola";

src/main/java/org/apache/commons/collections4/bloomfilter/SimpleBloomFilter.java
```diff
@@ -0,0 +1,127 @@
...
+     */
+    public SimpleBloomFilter(Shape shape, Function<T, SimpleBuilder> func, T data) {
+        this(shape, func);
+        this.merge( data );
```
garydgregory, line 83: No extra spaces please: `this.merge( data );` -> `this.merge(data);`

aherbert: As before, creating a filter and adding data are two separate functions. E.g. there is no constructor for an ArrayList that accepts a single item as well as the capacity. 

This concept is better removed or changed to create a filter with another Bloom filter (i.e. a collection). Here this highlights the dual functionality of the Shape which specifies not only the number of hash functions per item but also contains the number of expected items which is not strictly necessary for the filter to function. Creating from another Bloom filter can copy the other filter shape (indeed it mush match hash function identity and number of hash functions) but what would be the expected number of items?

Claudenw: The purpose of the SimpleBloomFilter was to provide a class that is simple to use and works for a great many users.  It is, in some ways, patterned after the Guava Bloom filter implementation.

The goal is to construct a Bloom filter that accepts objects of type T.  So there is a `merge( T )` and a `contains( T )` method.

Whether or not there is a constructor for an ArrayList that accepts a single item as well as a capacity is immaterial here.  By definition the Bloom filter has a concept of the number of items that will be added.  That concept is mathematically tied to other factors such as the number of hash functions for each item, the number of bits, and the probability of false positives.  Yes the shape has all 4 values.  No, you don't need all 4 values all the time. But, they are mathematically bound together and so make sense to be represented as the Shape.

This constructor is simply a short hand for `new SimpleBloomFilter(Shape shape, Function<T, SimpleBuilder> func).merge( T )` and does not restrict the user from creating another BloomFilter and merging it with this one.

If you are asking for another constructor that takes an existing BloomFilter as a parameter to provide the shape and an initial data load, that could be provided, but is not a reason not to have this constructor.  Is that what you are asking for?

src/main/java/org/apache/commons/collections4/bloomfilter/BitSetBloomFilter.java
```diff
@@ -25,7 +25,7 @@
 import org.apache.commons.collections4.bloomfilter.hasher.StaticHasher;
 
 /**
- * A bloom filter using a Java BitSet to track enabled bits. This is a standard
+ * A bloom filter uses a Java BitSet to track enabled bits. This is a standard
```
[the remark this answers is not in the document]
Claudenw: BitSet is a specific class.  The implementations in this library use BitSet but it is not required.  One could use long[] (without resorting to using the BitSet class) or one could use other BitSet like implementations.

src/main/java/org/apache/commons/collections4/bloomfilter/SimpleBloomFilter.java
aherbert: I am more asking to remove clutter from the API. It does not seem essential to me to have this constructor that saves 1 line of code.

When using a Bloom filter it would be expected to add hundreds or thousands of items. This will be done in a loop or over time by a method. Adding a single item at the creation point is a drop in the ocean against the final result of the saturated filter.

This is where my example of the ArrayList comes in (or any collection for that matter). There are no collections in the JDK that are created with an item except the singleton specialisations. Thus the language do not see this as a common case to satisfy. I'd suggest dropping the constructor that takes an item.

I would not add a constructor that accepts a Bloom filter since that also only saves 1 line of code over just calling merge on the next line.

src/main/java/org/apache/commons/collections4/bloomfilter/BitSetBloomFilter.java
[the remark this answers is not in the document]
aherbert: Perhaps the grammar should be:
```
A bloom filter **that** uses a Java BitSet
```
Anyway, this is a change totally unrelated to the SimpleBloomFilter in this PR. So should be reverted.

Since the backing implementation is largely irrelevant (aside from allowing the determination of memory requirements) a better wording would then be `A Bloom filter using a fixed size array ...` and the entire class renamed to ArrayBloomFilter. So this change should be done separately.

src/test/java/org/apache/commons/collections4/bloomfilter/hasher/HasherBuilderTest.java
```diff
@@ -29,7 +32,8 @@
 
 /**
  * Tests the
- * {@link org.apache.commons.collections4.bloomfilter.hasher.Hasher.Builder Hasher.Builder}.
+ * {@link org.apache.commons.collections4.bloomfilter.hasher.Hasher.Builder
+ * Hasher.Builder}.
```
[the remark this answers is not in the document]
Claudenw: The problem with converting T directly to a byte[] is two fold.

First, to do the conversion to the byte[] you need to know the Shape.

Second, it confuses the architecture.  In the architecture

The Hasher is responsible for
- hashing one or more items.
- being able to replay the hash of the item(s).
- converting the hash(es) into a collection of indexes into a bit vector.

The Shape is responsible for: 
- providing the number of bits that is represented by each item in the Bloom filter, also known as the number of hash functions, or _k_.
- providing the length of the bit vector, also known as _m_.
- providing the number of expected items in the Bloom filter, also known as _n_.
- calculating the probability of false positive results, also known as _p_.

The Bloom filter is responsible for:
- the implementation and management of a logical bit vector of _m_ bits.
- determining if another Bloom filter is contained in this filter. `contains( BloomFilter )`
- determining if the items contained in Hasher, taken as a whole, are contained in this filter. `contains( Hasher )`.  This is equivalent to constructing a second Bloom filter and calling the above but without the overhead.
- merging another bloom filter into this filter. `merge( BloomFilter )`
- merging the items contained in a Hasher, taken as a whole, into this filter. `merge( Hasher )`.  As with the contains method this is equivalent to constructing a Bloom filter and calling the above.

By using a function to go directly from T to byte[], the the function has to know the Shape of the filter that it is going to be used in.  Under the current architecture the Hasher does not care what the Shape is until the collection of indexes for a specific Shape are requested. 

The SimpleHasher was chosen because it defines the underlying hash function (Murmur128x64Cyclic), any Hasher implementation would be an acceptable return value for the Function in the constructor.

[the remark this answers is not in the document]
aherbert: I think I was not clear. The `byte[]` is just the part of the item T that you want to hash. So you have two situations where you convert the item T to either:

1. T -> List<byte[]>
2. T -> byte[]

These are then fed into the DynamicHasher. In case 1 you have a byte[] for each property or part of T you want to add to the Bloom filter. In case 2 you are adding the entire item T as a single entity.

You could group 1 and 2 into the same situation where I concatenate the parts that would make up the List<byte[]> to a single longer array and create a SimpleBuilder containing the result. The idea is the same: a single item T should only add one set of indices to the Bloom filter.

I do not suggest by-passing the Hasher. I suggest making it easier to specify case 2. Thus as a user I can convert my item to a binary form `byte[]` then the SimpleBloomFilter would do the rest of the work to hash the binary form and create a sequence of indexes for the filter shape.

This is why I suggested a `Function<T, byte[]>` version. The `byte[]` would be used internally by the SimpleBloomFilter to create a Hasher and generate indices for the Shape. Essentially this would be:

```java
byte[] b = ...;
Hasher h = new SimpleBuilder().with(b).build();
```

But this could be done using a specialised route that avoids all the object creation, for example using a variant of DynamicHasher that has a single `byte[]` to hash:

```java
byte[] b = ...; // The SimpleBloomFilter converts T to byte[]
HashFunction hf = ...; // From the SimpleBloomFilter
Hasher h = new DynamicHasher(hf, b);
```

src/test/java/org/apache/commons/collections4/bloomfilter/hasher/HasherBuilderTest.java
```diff
@@ -58,10 +62,9 @@ public Builder with(byte[] item) {
...
         final String ascii = "plain";
         final String extended = getExtendedString();
-        for (final String s : new String[] {ascii, extended}) {
-            for (final Charset cs : new Charset[] {
-                StandardCharsets.ISO_8859_1, StandardCharsets.UTF_8, StandardCharsets.UTF_16
-            }) {
+        for (final String s : new String[] { ascii, extended }) {
+            for (final Charset cs : new Charset[] { StandardCharsets.ISO_8859_1, StandardCharsets.UTF_8,
+                StandardCharsets.UTF_16 }) {
                 TestBuilder builder = new TestBuilder();
```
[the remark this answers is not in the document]
Claudenw: I tried to format using the src/conf/checkstyle.xml and the Eclipse checkstyle plugin.  Is that checkstyle.xml up to date or is there a problem with my Eclipse checkstyle configuration?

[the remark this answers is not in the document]
aherbert: If you are using Eclipse to format the entire file then there is a problem with your Eclipse formatting since it changed what was already in the file. There is no formatting file that has been applied to the collections source code. All formatting is done manually and must pass the checkstyle rules. These rules are not as strict as they could be but that is for another discussion.

In general you should only format new additions to existing source. If you reformat the existing source then you are adding noise to the PR.

src/main/java/org/apache/commons/collections4/bloomfilter/SimpleBloomFilter.java
Claudenw: The BloomFilter constructor that  takes a Hasher effectively does the same thing: creates the container and populates it.  Should that also be removed?  Perhaps a Jira ticket should be opened for that change.

src/test/java/org/apache/commons/collections4/bloomfilter/hasher/HasherBuilderTest.java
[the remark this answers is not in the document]
Claudenw: I can not figure out how to get Eclipse to format the line so that the continued array initialization (line 67 above) is not indented an extra 4 space.  Anyone have any idea how to do this?

PR #162, opened 2020-05-17
Removed constructors that set initial values

Claudenw:
fix for COLLECTIONS-763

Removes constructors that initialze the BloomFilter with data via the Hasher argument to the constructor.  The exception is the HasherBloomFilter which by definition wraps a Hasher.


kinow (commented):
I think if it has already been released, then we have to wait for a major version release in order to keep backward compatibility?

PR #163, opened 2020-05-26
refine travis-ci scripts

XenoAmess:


XenoAmess:
EmptyPropertiesTest fails on jdk13.
But....why?
the last commit 5 days ago can pass the build but this cannot?
I don't think I touched somethings that can affect the test result...
Anyway, it seems not related to this pr.

PR #165, opened 2020-06-02
[COLLECTIONS-764] Refine javadoc and fix typos

XenoAmess:
1. refine javadocs.
2. fix typos.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/list/CursorableLinkedList.java
+++ b/src/main/java/org/apache/commons/collections4/list/CursorableLinkedList.java
@@ -279,7 +279,7 @@ public class CursorableLinkedList<E> extends AbstractLinkedList<E> implements Se
     }
 
     /**
-     * Deregisters a cursor from the list to be notified of changes.
+     * Deregister a cursor from the list to be notified of changes.
      *
      * @param cursor  the cursor to deregister
      */
--- a/src/main/java/org/apache/commons/collections4/list/CursorableLinkedList.java
+++ b/src/main/java/org/apache/commons/collections4/list/CursorableLinkedList.java
@@ -279,7 +279,7 @@ public class CursorableLinkedList<E> extends AbstractLinkedList<E> implements Se
     }
 
     /**
-     * Deregister a cursor from the list to be notified of changes.
+     * Deregisters a cursor from the list to be notified of changes.
      *
      * @param cursor  the cursor to deregister
      */
--- a/src/main/java/org/apache/commons/collections4/trie/AbstractPatriciaTrie.java
+++ b/src/main/java/org/apache/commons/collections4/trie/AbstractPatriciaTrie.java
@@ -71,8 +71,8 @@ abstract class AbstractPatriciaTrie<K, V> extends AbstractBitwiseTrie<K, V> {
     }
 
     /**
-     * Constructs a new {@link org.apache.commons.collections4.Trie org.apache.commons.collections4.Trie Trie}
-     * using the given {@link KeyAnalyzer} and initializes the {@link org.apache.commons.collections4.Trie Trie}
+     * Constructs a new {@link org.apache.commons.collections4.Trie}
+     * using the given {@link KeyAnalyzer} and initializes the {@link org.apache.commons.collections4.Trie}
      * with the values from the provided {@link Map}.
      */
     protected AbstractPatriciaTrie(final KeyAnalyzer<? super K> keyAnalyzer,
@@ -103,7 +103,7 @@ abstract class AbstractPatriciaTrie<K, V> extends AbstractBitwiseTrie<K, V> {
     }
 
     /**
-     * A helper method to increment the {@link Trie} size and the modification counter.
+     * A helper method to increment the {@link org.apache.commons.collections4.Trie} size and the modification counter.
      */
     void incrementSize() {
         size++;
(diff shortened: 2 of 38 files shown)


src/main/java/org/apache/commons/collections4/trie/AbstractPatriciaTrie.java
```diff
@@ -341,10 +341,6 @@ public V selectValue(final K key) {
         return entry.getValue();
     }
 
-    /**
-     * This is equivalent to the other {@link #selectR(TrieEntry, int, Object, int, Cursor, Reference)}
-     * method but without its overhead because we're selecting only one best matching Entry from the {@link Trie}.
-     */
```
kinow, line 347: I think this one needs to stay?

XenoAmess:
Several people suggested me not to put everything in one single pr, but split them.
So I listened to their advice, and let things done one by one.

src/main/java/org/apache/commons/collections4/trie/AbstractPatriciaTrie.java
XenoAmess: > I think this one needs to stay?

I delete it because I can never found that function it point to.

kinow: Ah, you are correct! There was another method, but it was removed when the base class was renamed and the code changed a little bit to create this abstract class - https://github.com/apache/commons-collections/commit/83226e162de60bd22551b1a37cc3e513811e1449#diff-821b6e7d45ec0fd696f3cdc8027f029cL303

All good here then. Thanks!

src/main/java/org/apache/commons/collections4/list/CursorableLinkedList.java
```diff
@@ -279,7 +279,7 @@ protected void registerCursor(final Cursor<E> cursor) {
     }
 
     /**
-     * Deregisters a cursor from the list to be notified of changes.
+     * Deregister a cursor from the list to be notified of changes.
```
dota17, line 282: garydgregory on 17 Apr Member
If a method starts with "make...", it usually reads better to start the Javadoc with "Makes..." IMO, It's confusing to have the Javadoc be different, otherwise you could say that all method that start with "set..." also "forces" a value.

src/main/java/org/apache/commons/collections4/trie/PatriciaTrie.java
```diff
@@ -29,7 +29,7 @@
  * all data at the edges of the {@link org.apache.commons.collections4.Trie}
  * (and having empty internal nodes), PATRICIA stores data in every node.
  * This allows for very efficient traversal, insert, delete, predecessor,
- * successor, prefix, range, and {@link #select(Object)}
+ * successor, prefix, range, and {@link #select(String)}
```
dota17, line 32: [ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.1.1:javadoc (default-cli) on project commons-collections4: An error has occurred in Javadoc report generation: 
[ERROR] Exit code: 1 - /home/travis/build/apache/commons-collections/src/main/java/org/apache/commons/collections4/trie/PatriciaTrie.java:32: error: reference not found
[ERROR]  * successor, prefix, range, and {@link #select(String)}

Maybe it's not need to update.

src/main/java/org/apache/commons/collections4/list/CursorableLinkedList.java
XenoAmess: yes you are right.
will revert it.

src/main/java/org/apache/commons/collections4/trie/PatriciaTrie.java
XenoAmess: @dota17 
yes you are correct.
It can pass on one of my machines with jdk8 so I thought it be correct.
but actually according to travis-ci, it can only pass openjdk8 and oraclejdk8, but fails on other jdks.
so I will revert it.
Thank you!

XenoAmess: @dota17
Though I have no idea why it can pass on 8 but not others....
Strange isn't it...
image

XenoAmess:
the travis-ci false pending for several days already.
will try to retrigger it.

wlfbck:
Do these fixes include that for example [CircularFifoQueue](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/queue/CircularFifoQueue.html) has dummydoc for `poll()`, `element()`, `peek()`, `remove()`? Because that stuff is not in the source, so i assume it is generated. But it hides the original javadoc from java.util.Queue, which is very annoying.

XenoAmess:
@wlfbck
> Do these fixes include that for example [CircularFifoQueue](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/queue/CircularFifoQueue.html) has dummydoc for `poll()`, `element()`, `peek()`, `remove()`? Because that stuff is not in the source, so i assume it is generated. But it hides the original javadoc from java.util.Queue, which is very annoying.

Nope.
But I think it shall be able to be solved by adding: 
```
    /**
     * {@inheritDoc}
     */
```

The question you mentioned is planned to be fixed by using maven-javadoc-plugin.
Currently the plan is:
1. wait for me fully recover from a flu.(not yet QAQ)
2. my pr about some basic refinement to maven-javadoc-plugin get reviewed and passed.(still waiting)
3. I will make a pr to fix some bugs in maven-javadoc-plugin's javadoc:fix operation, and see it get passed.
4. I use the javadoc:fix on this repo, and see what doc can be auto fixed.

The reason I don't want to use the javadoc:fix now is it really have some bugs, I mean, quite some.
But that will take lots of times to wait (maven guys a really...busy...)
So if you really want some of the docs be fixed soon, you can apply javadoc:fix on it, and revert all things it messed up, only leave some good changes.(but that is also a time coster).

kinow:
Merged. Thanks @XenoAmess !
Bruno

PR #176, opened 2020-08-21
[COLLECTIONS-765] Update Javadoc of ListIteratorWrapper#remove

TobiasKiecker:
While doing research, we found some counterexamples to the original javadoc and thus think that there was a typo.


kinow (approved):
@TobiasKiecker the new javadocs look good to me. Thanks for the investigation and for updating it.

No issues with binary backward compatibility (https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html#jls-13.4.21).

I'm adding a JIRA so that it's included in the next changelog/release notes.

Thanks!
Bruno

PR #188, opened 2020-10-02
MapBuilder class to decide among various types of Maps

Amita-Pradhan:


src/main/java/org/apache/commons/collections4/MapBuilder.java
```diff
@@ -0,0 +1,115 @@
...
+ *
+ * <pre>{@code
+ * Map builderMap = new MapBuilder().setIterationOrder(MapBuilder.KeyOrder.INSERTION_ORDER).build();
+ * builderMap.put("A", 1);
+ * builderMap.put("X", 24);
+ * builderMap.put("B", 2);
+ * builderMap.put("Y", 26);
+ * }</pre>
+ *
+ * @author Amita Pradhan
```
kinow, line 20: We have dropped author tags (it will be recorded in git, and also in our changelog)

kinow (commented):
I am used to choosing a map based on what I need, so not sure if I would use it immediately.

But looks very interesting at first. Needs a JIRA issue too (which is used for the changelog too).

Will take another look later. But fter a cursory look, code looks good.

Thanks for your contribution!
Bruno

src/main/java/org/apache/commons/collections4/MapBuilder.java
```diff
@@ -0,0 +1,115 @@
...
+ */
+public class MapBuilder {
+
+    private Comparator comparator;
+    private KeyOrder iterationOrder;
+    private boolean synchronizedMap;
+    private boolean immutable;
+    private Map data;
```
garydgregory, line 28: Missing generics.

src/main/java/org/apache/commons/collections4/MapBuilder.java
```diff
@@ -0,0 +1,115 @@
...
+ * builderMap.put("X", 24);
+ * builderMap.put("B", 2);
+ * builderMap.put("Y", 26);
+ * }</pre>
+ *
+ * @author Amita Pradhan
+ */
+public class MapBuilder {
+
+    private Comparator comparator;
```
garydgregory, line 24: Missing generics.

src/main/java/org/apache/commons/collections4/MapBuilder.java
```diff
@@ -0,0 +1,115 @@
...
+        if(immutable) {
+            map = Collections.unmodifiableMap(map);
+        }
+
+        return map;
+    }
+
+    enum KeyOrder
+    {
+        RANDOM, NATURAL_ORDER, INSERTION_ORDER, COMPARATOR_ORDER;
```
garydgregory, line 113: The term RANDOM is not correct since you are not randomizing anything.

src/test/java/org/apache/commons/collections4/MapBuilderTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * Test Cases for Map Builder
+ */
+class MapBuilderTest {
+
+    @Test
+    void setComparator() {
+        // Null Comparator
+        Map myMap = new HashMap();
```
garydgregory, line 20: Missing generics.

src/test/java/org/apache/commons/collections4/MapBuilderTest.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * Test Cases for Map Builder
+ */
+class MapBuilderTest {
```
garydgregory, line 15: IMO, you do not/should not need to test the behavior of Map implementations. Checking the type of the built instance is enough.

src/main/java/org/apache/commons/collections4/MapBuilder.java
```diff
@@ -0,0 +1,115 @@
...
+ */
+public class MapBuilder {
+
+    private Comparator comparator;
+    private KeyOrder iterationOrder;
+    private boolean synchronizedMap;
+    private boolean immutable;
+    private Map data;
+
+    public MapBuilder() {
```
garydgregory, line 30: I only see this useful if you cover all Map implementations in the JDK. It's really the only way IMO to see if this builder is handy or not but I am somewhat doubtful in general. Let's see where this goes from here... For, me it's simpler to use the right class in the right place. For example, how do I build a `SortedMap<Foo, Bar>` and is returned to the call site as a `SortedMap<Foo, Bar>`?

garydgregory (changes requested):

src/main/java/org/apache/commons/collections4/MapBuilder.java
Amita-Pradhan: Done. Removed.

src/main/java/org/apache/commons/collections4/MapBuilder.java
Amita-Pradhan: Done.

src/main/java/org/apache/commons/collections4/MapBuilder.java
Amita-Pradhan: Done

src/main/java/org/apache/commons/collections4/MapBuilder.java
Amita-Pradhan: Yup.. changed it to UNORDERED for now. Please let me know your thoughts.

src/test/java/org/apache/commons/collections4/MapBuilderTest.java
Amita-Pradhan: Done.

src/test/java/org/apache/commons/collections4/MapBuilderTest.java
Amita-Pradhan: I drifted towards functional testing here.Just to clarify checking instance type as MapType expected (LinkedHashMap/TreeMap) etc should be done here ?

src/main/java/org/apache/commons/collections4/MapBuilder.java
Amita-Pradhan: Will try to add as many as possible :-). Wanted to have something like this when I started using Maps and tried to search generateMap in MapUtils for help.

garydgregory: Hi @Amita-Pradhan ,

This file is missing the Apache license header. Note that the GitHub build fails (see Actions tab).

Amita-Pradhan: Done. Fixed the header check and checkstyle errors.

Claudenw:
@kinow , @garydgregory , any reason not to merge this or should we just close the pull request?

garydgregory:
> @kinow , @garydgregory , any reason not to merge this or should we just close the pull request?

If it were just me, I would close it. I feel I am better served knowing what the JDK offers , what Commons Collection offers, and the difference. It feels like a solution in search of a problem.  I don't see using this in code I write FWIW.

PR #190, opened 2020-10-12
[COLLECTIONS-769] Fix flaky UnmodifiableMultiValuedMapTest

tongxin97:
Detailed description: https://issues.apache.org/jira/browse/COLLECTIONS-769

Fix: replace the affected assertions with calls to a helper function that asserts all the added values are still present in the map under test.

### DIFF
--- a/src/test/java/org/apache/commons/collections4/multimap/UnmodifiableMultiValuedMapTest.java
+++ b/src/test/java/org/apache/commons/collections4/multimap/UnmodifiableMultiValuedMapTest.java
@@ -48,6 +48,12 @@ public class UnmodifiableMultiValuedMapTest<K, V> extends AbstractMultiValuedMap
     }
 
     // -----------------------------------------------------------------------
+    private void assertMapContainsAllValues(MultiValuedMap<K, V> map) {
+        assertEquals("[uno, un]", map.get((K) "one").toString());
+        assertEquals("[dos, deux]", map.get((K) "two").toString());
+        assertEquals("[tres, trois]", map.get((K) "three").toString());
+    }
+
     @Override
     public boolean isAddSupported() {
         return false;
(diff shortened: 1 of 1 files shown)


src/test/java/org/apache/commons/collections4/multimap/UnmodifiableMultiValuedMapTest.java
```diff
@@ -48,6 +48,12 @@ public static Test suite() {
     }
 
     // -----------------------------------------------------------------------
+    private void assertMapContainsAllValues(MultiValuedMap<K, V> map) {
```
kinow, line 51: @tongxin97 can you add a comment over this method, just to clarify why we needed it, please?

kinow (approved):

kinow:
I couldn't reproduce the issue, as `Hashmap` always returned the values in order. But I think it may change depending on number of elements in map, JVM implementation, etc. So looks safe to go with this improved version. Merging after the javadoc is added :+1:

src/test/java/org/apache/commons/collections4/multimap/UnmodifiableMultiValuedMapTest.java
tongxin97: Sounds good. Updated.

kinow:
Merged! Thanks @tongxin97 !

tongxin97:
@kinow Thanks! I saw there might be an error on merging [here](https://travis-ci.org/github/apache/commons-collections/jobs/734868363). Just want to highlight this in case it is something that needs to be taken care of.

kinow:
Might be because I squashed the commits when merging. I tested it with `mvn` locally after squashing them, and looks like the `master` branch is building OK. But will keep an eye open in case a similar issue happens. Thanks for the heaps up @tongxin97 :+1:

PR #191, opened 2020-10-15
WIP DO NOT MERGE YET Junit assert throws

nhojpatrick:


src/test/java/org/apache/commons/collections4/IteratorUtilsTest.java
```diff
@@ -436,9 +442,11 @@ public void testArrayListIterator() {
         }
     }
 
-    @Test(expected = NullPointerException.class)
+    @Test
     public void testAsEnumerationNull() {
-        IteratorUtils.asEnumeration(null);
+        final Executable testMethod = () -> IteratorUtils.asEnumeration(null);
+        final NullPointerException thrown = assertThrows(NullPointerException.class, testMethod);
+        assertThat(thrown.getMessage(), is(equalTo("iterator")));
```
garydgregory, line 449: -1: This is over the top IMO and gets too deep into white-box testing; this test now asserts way too much, all that we care is that the NPE is thrown, not the details of the message. This test will break as soon as we reimplement the guts of the method.
```
assertThrows(NullPointerException.class, () -> IteratorUtils.asEnumeration(null));
```
is enough

PR #194, opened 2020-11-09
[COLLECTIONS-771] Fix flaky AbstractMultiValuedMapTest#testToString

tongxin97:
Issue description: https://issues.apache.org/jira/browse/COLLECTIONS-771

Assert `map.toString()` is equal to either one of the two possible string representations:
1. "{A=[X, Y, Z], B=[U, V, W]}"
2. "{B=[U, V, W], A=[X, Y, Z]}"

### DIFF
--- a/src/test/java/org/apache/commons/collections4/multimap/AbstractMultiValuedMapTest.java
+++ b/src/test/java/org/apache/commons/collections4/multimap/AbstractMultiValuedMapTest.java
@@ -675,7 +675,10 @@ public abstract class AbstractMultiValuedMapTest<K, V> extends AbstractObjectTes
         map.put((K) "B", (V) "U");
         map.put((K) "B", (V) "V");
         map.put((K) "B", (V) "W");
-        assertEquals("{A=[X, Y, Z], B=[U, V, W]}", map.toString());
+        assertTrue(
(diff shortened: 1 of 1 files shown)


kinow (approved):
Sounds reasonable to me. We could also assert that it contains each string separately, but I think that would be the case only if we added more keys/values. Good enough for now I think.

Thanks @tongxin97 !

kinow:
Travis CI failure not related. And it happened on JVM 11 only:

```
[ERROR] Tests run: 45, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.041 s <<< FAILURE! - in org.apache.commons.collections4.properties.EmptyPropertiesTest

[ERROR] testSave  Time elapsed: 0.016 s  <<< FAILURE!

org.junit.internal.ArrayComparisonFailure: arrays first differed at element [33]; expected:<52> but was:<51>

	at org.apache.commons.collections4.properties.EmptyPropertiesTest.testSave(EmptyPropertiesTest.java:262)

Caused by: java.lang.AssertionError: expected:<52> but was:<51>

	at org.apache.commons.collections4.properties.EmptyPropertiesTest.testSave(EmptyPropertiesTest.java:262)
```

JVM 11 test on GH action passed. Might be another flaky test.

kinow:
Added change log entry, updated JIRA, merged. Thanks @tongxin97 !

PR #195, opened 2020-11-19
[COLLECTIONS-772] - Add private constructor to ArrayUtils; better Javadocs.

arturobernalg:


src/main/java/org/apache/commons/collections4/IterableUtils.java
```diff
@@ -50,6 +50,11 @@
  */
 public class IterableUtils {
 
+    /**
+     * {@code IterableUtils} should not normally be instantiated.
+     */
+    private IterableUtils() {}
+
```
aherbert, line 57: This is a public class that used to have a public (default) constructor. You have made it private which breaks binary compatibility.

The other class `ArrayUtils` where you added a private constructor is a package private class and so that change does not break binary compatibility which compares the API at public and protected scope.

arturobernalg: Hi @aherbert 

Yep. I know. and that's my intentions. Unify the constructor.

garydgregory: There is nothing planned for 5.0 AFAIK, 4.5 would be next. Creating a 5.0 branch now would be premature IMO and I personally would not want to spend the time maintaining it.

aherbert: Perhaps make the constructor public and mark it as deprecated for removal in 5.0 with a note that the class is not intended to be instantiated. This will clarify the intention and ensure the change does happen when 5.0 is released as all deprecated tags will be addressed on a major version increment.

Or remove this from the PR as binary compatibility is an important goal of commons within major versions, even for something trivial like this where the use cases of instances of this class are non-existent and I think it unlikely that any end user would be creating instances.

arturobernalg: Hi,

I have reverted the changes  of the class IterableUtils. Although I think it is necessary to match all the constructors to private access

kinow: >Perhaps make the constructor public and mark it as deprecated for removal in 5.0 with a note that the class is not intended to be instantiated. This will clarify the intention and ensure the change does happen when 5.0 is released as all deprecated tags will be addressed on a major version increment.

+1 and will remember us that this can be done by 5.0.

src/main/java/org/apache/commons/collections4/IteratorUtils.java
```diff
@@ -123,7 +123,7 @@
     private static final String DEFAULT_TOSTRING_DELIMITER = ", ";
 
     /**
-     * IteratorUtils is not normally instantiated.
+     * {@code IteratorUtils} should not normally be instantiated.
```
garydgregory, line 126: Hi @arturobernalg 
If you want to change the all the comments, I do not think we need to duplicate the class names in each comment, I would simply say:
```
/**
 * Don't allow instances.
 */
```

arturobernalg: OK, changed and rebased on 1 commit

src/main/java/org/apache/commons/collections4/ArrayUtils.java
```diff
@@ -38,6 +38,12 @@
  */
 class ArrayUtils {
 
+
+    /**
+     * Don't allow instances.
+     */
+    private ArrayUtils() {}
```
kinow, line 45: This appears to be the only change that breaks backward compatibility.

kinow (commented):
Looks good to me. I think it cannot be merged until a major release is cut due to `ArrayUtils` constructor being made private (instead of deprecating it, for instance).

kinow (commented):
We cannot merge it right now, I think @arturobernalg, as this would break backward compatibility. Unless I missed something, I think this class had the default public constructor, but now we are replacing by a private one (same issue from `IterableUtils` I think)

src/main/java/org/apache/commons/collections4/IterableUtils.java
arturobernalg: Thank you @kinow

arturobernalg:
HI @kinow 

But the class is a package private class and so that change does not break binary compatibility

src/main/java/org/apache/commons/collections4/ArrayUtils.java
kinow: Ha! Thanks for your last comment @arturobernalg ! I did not see the class was private. I checked with clirr, it looks good.

```
kinow@ranma:~/Development/java/apache/commons-collections$ git log -n 1
commit 0679fa25d199ef6f276a430c43fcdecb1aa7c1c1 (HEAD -> pr-195)
Author: Arturo Bernal <arturobernalg@gmail.com>
Date:   Thu Nov 19 07:45:05 2020 +0100

    COLLECTIONS-772 - Provide private constructor to the Utils Collections
kinow@ranma:~/Development/java/apache/commons-collections$ mvn clirr:check
...
[INFO] --- clirr-maven-plugin:2.8:check (default-cli) @ commons-collections4 ---
[INFO] artifact org.apache.commons:commons-collections4: checking for updates from repo.jenkins-ci.org
[INFO] artifact org.apache.commons:commons-collections4: checking for updates from central
[INFO] Comparing to version: 4.4
[INFO] Succeeded with 0 errors; 0 warnings; and 49 other changes.
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 12.449 s
[INFO] Finished at: 2020-11-21T22:15:37+13:00
[INFO] ------------------------------------------------------------------------

```

kinow (approved):

src/main/java/org/apache/commons/collections4/ArrayUtils.java
arturobernalg: Thank you @kinow

garydgregory: Note that we need to switch to JApiCmp...

garydgregory:
@arturobernalg 
In the future, please provide a description for the PR, at least with a link to a JIRA, if a JIRA exists for a PR.

PR #205, opened 2021-01-13
[COLLECTIONS-761] - IteratorUtils.toList() will return empty list when size of iterator is 1

arturobernalg:
Add test with javadoc pointer to Jira and prove that it not reproducible

### DIFF
--- a/src/test/java/org/apache/commons/collections4/IteratorUtilsTest.java
+++ b/src/test/java/org/apache/commons/collections4/IteratorUtilsTest.java
@@ -47,9 +47,7 @@ import java.util.Set;
 import java.util.LinkedHashSet;
 
 import org.apache.commons.collections4.iterators.*;
-import org.apache.commons.collections4.list.GrowthList;
 import org.apache.commons.collections4.map.EntrySetToMapIteratorAdapter;
-import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.w3c.dom.Node;
--- a/src/test/java/org/apache/commons/collections4/IteratorUtilsTest.java
+++ b/src/test/java/org/apache/commons/collections4/IteratorUtilsTest.java
@@ -34,6 +34,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.ListIterator;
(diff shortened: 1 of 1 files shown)


src/test/java/org/apache/commons/collections4/IteratorUtilsTest.java
```diff
@@ -46,7 +47,9 @@
 import java.util.LinkedHashSet;
 
 import org.apache.commons.collections4.iterators.*;
+import org.apache.commons.collections4.list.GrowthList;
 import org.apache.commons.collections4.map.EntrySetToMapIteratorAdapter;
+import org.junit.Assert;
```
kinow, line 52: Is this import being used?

src/test/java/org/apache/commons/collections4/IteratorUtilsTest.java
```diff
@@ -46,7 +47,9 @@
 import java.util.LinkedHashSet;
 
 import org.apache.commons.collections4.iterators.*;
+import org.apache.commons.collections4.list.GrowthList;
```
kinow, line 50: Is this import being used?

kinow (commented):
The test code is simple, but at the moment it's not failing. From what I understood from the JIRA linked, it was supposed to test something that is not working? Unless we are adding the test for coverage?

src/test/java/org/apache/commons/collections4/IteratorUtilsTest.java
arturobernalg: True. Fix It

src/test/java/org/apache/commons/collections4/IteratorUtilsTest.java
arturobernalg: True. Fix It

arturobernalg:
> The test code is simple, but at the moment it's not failing. From what I understood from the JIRA linked, it was supposed to test something that is not working? Unless we are adding the test for coverage?

Hi @kinow 

First that all Happy New Years. I tried to prove the Jira and was not possible. I think it's working correctly. For some reason @garydgregory  suggested adding the test. I guess its related with my last question about the regression

kinow:
Hi @arturobernalg , happy 2021!

> I tried to prove the Jira and was not possible. I think it's working correctly.

From the JIRA description, it sounded to me like we would get an empty list when it was not desired. So that was the issue/problem we would need to fix for the next release.

> For some reason @garydgregory suggested adding the test. I guess its related with my last question about the regression

I think what Gary meant, is that along what I explained above, that if we indeed have an issue, it is really good if the person that created the JIRA issue can provide a unit test, and/or some small code snippet to show the developers what needs to be fixed.

For example, if you have a JIRA-1234 issue, where it says "Serializing an empty list results in an Exception", then it would be nice if the person reporting could include some code or unit test like

```java
@Test
public void testEmptyListSerializationIssue() {
    // NOTE: we are not testing assertThrows(() -> {...}) here, as the test is for "how should the API behave", and
    //             not "how the API is behaving now"
    UtilCode.serialize(new SomeCollectionsClass();
}
```

Hope it helps
Bruno

arturobernalg:
Hi @kinow 

thank you very much. I understand now. 
I close the PR  and we wait for the person who reported the Jira send a test

PR #210, opened 2021-01-26
[COLLECTIONS-781] - INDEX_NOT_FOUND Constant

arturobernalg:
The idea it's create a constant that represent that the Index to search it's not found

### DIFF
--- a/src/main/java/org/apache/commons/collections4/ListUtils.java
+++ b/src/main/java/org/apache/commons/collections4/ListUtils.java
@@ -42,7 +42,6 @@ import org.apache.commons.collections4.sequence.SequencesComparator;
  * @since 1.0
  */
 public class ListUtils {
-
     /**
      * Don't allow instances.
      */
@@ -539,7 +538,7 @@ public class ListUtils {
                 }
             }
         }
-        return -1;
+        return CollectionUtils.INDEX_NOT_FOUND;
     }
 
     //-----------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


sebbASF:
Why not have a single shared constant, rather than 1 per class?

arturobernalg:
> Why not have a single shared constant, rather than 1 per class?

HI @sebbASF 

I agree with you that we should have  a single shared constant.  What do you think about have this constant in CollectionUtils class?

arturobernalg:
Changed

src/main/java/org/apache/commons/collections4/ListUtils.java
```diff
@@ -539,7 +538,7 @@ public static int hashCodeForList(final Collection<?> list) {
                 }
             }
         }
-        return -1;
+        return ArrayUtils.INDEX_NOT_FOUND;
```
garydgregory, line 541: Hi @arturobernalg 

I think you can replace `ArrayUtils.INDEX_NOT_FOUND` with `CollectionUtils.INDEX_NOT_FOUND` and delete `ArrayUtils.INDEX_NOT_FOUND` since it is package private. Just make sure you use a better comment for `CollectionUtils.INDEX_NOT_FOUND` like the one for `ArrayUtils.INDEX_NOT_FOUND`.

arturobernalg: Changed @garydgregory

PR #215, opened 2021-02-08
Add test cases for indexOf and contains method of ArrayUtils class

HiFromAjay:
The test cases are adapted from https://github.com/apache/commons-lang/blob/master/src/test/java/org/apache/commons/lang3/ArrayUtilsTest.java


kinow (commented):
I'm not sure how to review this one. The tests are simple, nothing wrong with the code. But I don't recall why we copied this class from Lang, instead of adding a dependency that's shaded (as we did in some other component I think?).

If there are tests in Lang, also not sure if we need/should duplicate them here. :thinking: 

Thanks for the PR anyway @HiFromAjay :+1:

Claudenw:
@kinow Any reason not to merge this?  Or should we just close the pull request?

PR #216, opened 2021-02-26
feature/COLLECTIONS-783 Checks if any value in the given Collections is null or empty.

arturobernalg:
Checks if any value in the given Collections is null or empty. Return true if there is at least one null value in the Collection, and return false if all the values are non-null and not empty.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/CollectionUtils.java
+++ b/src/main/java/org/apache/commons/collections4/CollectionUtils.java
@@ -1559,6 +1559,40 @@ public class CollectionUtils {
         return coll == null || coll.isEmpty();
     }
 
+    //-----------------------------------------------------------------------
+    /**
+     * Checks if any value in the given Collections is {@code null} or empty.
+     *
+     * <p>
+     * Null returns true.
+     * </p>
+     *
+     * <pre>
+     * CollectionUtils.isAnyEmpty(*)                    = false
+     * CollectionUtils.isAnyEmpty(*, *)                 = false
+     * CollectionUtils.isAnyEmpty(*,EMPTY_COLLECTION)   = true
+     * CollectionUtils.isAnyEmpty(null)                 = true
+     * CollectionUtils.isAnyEmpty(null, null)           = true
+     * CollectionUtils.isAnyEmpty(null, *)              = true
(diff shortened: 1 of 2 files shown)


kinow (commented):
Code looks OK. I think we just need to consider whether we want a new method for this, or if existing ways to use the code for this are better alternatives (easier maintenance having less methods).

```java
List<String> l = Arrays.asList("a", "b", "c", "");

boolean b = CollectionUtils.containsAny(l, new Object[] { null, "" }); // or
boolean b = IterableUtils.matchesAny(l, (e) -> e == null || "".equals(e));
```

This is similar to [lang]'s `isAnyEmpty`, but instead of a chain of characters, we check the array of objects. But probably we don't want a `isAnyBlank` (`"     "` is blank, not empty).

And this one would need a JIRA issue @arturobernalg but we can wait for others to chime in and take a look at the new code first before creating one if you'd like.

Thanks for the PR's!
Bruno

arturobernalg:
> Code looks OK. I think we just need to consider whether we want a new method for this, or if existing ways to use the code for this are better alternatives (easier maintenance having less methods).
> 
> ```java
> List<String> l = Arrays.asList("a", "b", "c", "");
> 
> boolean b = CollectionUtils.containsAny(l, new Object[] { null, "" }); // or
> boolean b = IterableUtils.matchesAny(l, (e) -> e == null || "".equals(e));
> ```
> 
> This is similar to [lang]'s `isAnyEmpty`, but instead of a chain of characters, we check the array of objects. But probably we don't want a `isAnyBlank` (`" "` is blank, not empty).
> 
> And this one would need a JIRA issue @arturobernalg but we can wait for others to chime in and take a look at the new code first before creating one if you'd like.
> 
> Thanks for the PR's!
> Bruno

Hi @kinow 

Jira created COLLECTIONS-783.

I will check your comment and analyze what can we reuse and what will the  best alternative. Thank you very much

arturobernalg:
CollectionUtils.containsAny  check  bjects.requireNonNull and throw NPE. IMO this will  break the binary.  I keep checking `IterableUtils.matchesAny`

arturobernalg:
HI @kinow 

We can use `CollectionUtils.containsAny(collectionE, new Object[] { null, "" })`

Close the PR

TY

PR #221, opened 2021-03-25
Minor improvement

arturobernalg:
- Use lambda and method reference
- Simplify conditions
- inline variable
- Simplify boolean conditions


arturobernalg:
sorry for the number of modify files. I didn't realize that there were so many

PR #229, opened 2021-04-21
COLLECTIONS-789 - Redundant 'if' statement

arturobernalg:
* Simplified to a single return statement.


src/test/java/org/apache/commons/collections4/CollectionUtilsTest.java
```diff
@@ -1348,10 +1348,7 @@ public void testHashCode() {
         final Equator<Integer> e = new Equator<Integer>() {
             @Override
             public boolean equate(final Integer o1, final Integer o2) {
-                if (o1 % 2 == 0 ^ o2 % 2 == 0) {
-                    return false;
-                }
-                return true;
+                return !(o1 % 2 == 0 ^ o2 % 2 == 0);
```
LarsBodewig, line 1351: Newbie myself, but isn't that equivalent to `(o1 % 2 == 0) == (o2 % 2 == 0)`? I have no idea what's easier to read though.

PR #237, opened 2021-05-21
COLLECTIONS-786: fix PassiveExpiringMap documentation

vandrewskis:
`containsKey` does not access entire map, while
`containsValue` does


kinow (approved):
Hi @vandrewskis 

Thanks for the PR. I agree with the issue description and with the patch here. Both `containsValue` and `entrySet` call `removeAllExpired(now());` which iterates all elements as the Javadoc says. While `containsKey` does not.

`size` also calls that, but it's part of the "etc" :-) so no need to include.

+1

kinow:
Added changelog, updated JIRA fixVersion, merged. Thanks!!

vandrewskis:
@kinow , thanks for review and merge 👍

PR #242, opened 2021-06-18
feat(MapUtils): Add empty string judgment

kingmao666:
Add empty string judgment


kinow (changes requested):
I think the current behaviour, described in the javadocs, only guarantees that the method is null-safe. It doesn't say anything about empty strings. Eventually someone could request that blank (i.e. "    ") strings are also ignored. But IMO, these cases are better handled in the user application, leaving this method to return the `.toString()` of an object in the map/key.

If the user doesn't want `""` or `"   "` strings, s/he can filter those values later. Also, I'm not sure if the `Object answer` is a String, in which case it wouldn't make sense to have the `answer != ""` check too.

garydgregory:
-1 I agree with @kinow. This would also create a bad precedent where empty strings are treated differently in a single method and also likely to break existing applications.

PR #244, opened 2021-07-28
Adds functionality to build a java Map from collections of keys and values

hasandiwan:


garydgregory:
-1 code does not even compile.

LarsBodewig:
Additionally:
1. How do the `keys` and `values` correspond with each other if the keys are an unsorted `Set`?
2. If `indexOf` existed, how should `values.indexOf(k)` return a value different from `k`?

garydgregory:
Closing: No reply.

PR #247, opened 2021-08-11
Shallow cloning on travis

YunLemon:
Travis CI provide a way to shallow clone a repository. This has the obvious benefit of speed, since you only need to download a small number of commits.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,3 +38,6 @@ script:
 after_success:
   # jacoco will run in the main script. Include the profile to submit to coveralls.
   - mvn -V jacoco:report coveralls:report -Ptravis-jacoco
+
+git:
+  depth: 1


garydgregory:
We might as well keep the default (50) as there could be side effects: https://stackoverflow.com/questions/31278233/disadvantages-of-shallow-cloning-on-travis-and-other-ci-services

YunLemon:
Thank you for your reply! I have read the answers of this question. 
I think there are two main problems with using shallow clones.
1. Push too fast. In daily development, can we really push so fast? As there exists a new commit, why do we still care about the result of an old commit?
2. Restart an old build. Do we really have this need?

By the way, what do you think is the best value for depth?

garydgregory:
> Thank you for your reply! I have read the answers of this question.
> I think there are two main problems with using shallow clones.
> 
> 1. Push too fast. In daily development, can we really push so fast? As there exists a new commit, why do we still care about the result of an old commit?
> 2. Restart an old build. Do we really have this need?
> 
> By the way, what do you think is the best value for depth?

My view is to follow the configuration by exception mindset and only override a default if you must. In this case, we do not need to. In addition, keeping the default keeps the configuration simpler and easier to maintain.

YunLemon:
Thank you! Mr. Gregory. I understand. I really like your book _JUnit in Action_ and _Java Persistence with Hibernate_.
To be honest, I am a graduate student and doing some research on the config problems of Travis CI.
I also see that this project define jobs that are allowed to fail. As official document says: [https://docs.travis-ci.com/user/customizing-the-build#fast-finishing](url). Do you think we should set it true to get faster feedback.

PR #251, opened 2021-09-08
hamcrest v2.2

nhojpatrick:


kinow (commented):
I didn't understand the doubble `hamcrest` artifact dependency.

image

It seems to me that we have two entries for `org.hamcrest:hamcrest:2.2`, scoped for `test`, pretty much identical except that in one entry we exclude `hamcrest-core`? I expanded as in the screenshot to see if it was in a different profile, but these dependencies are instead redundant I think?

garydgregory:
Clsoing, let Dependabot do it.

nhojpatrick:
Just reviewed, it was dodge merge/rebase cause the duplication. Also dependabot won't upgrade, as the artifactId change.

PR #255, opened 2021-09-23
COLLECTIONS-796  SetUniqueList.createSetBasedOnList doesn't add list elements to return value Export

cbkurz:
* There seems no immediate reason why the addAll was removed in commit b1c45ac691d46a8c609f2534d2adfa59c0599527
* Tests added
* Functionality implemented again
* removed typo in UnmodifiableListIterator.unmodifiableListIterator (previous 'UnmodifiableListIterator.umodifiableListIterator')

### DIFF
--- a/src/main/java/org/apache/commons/collections4/iterators/UnmodifiableListIterator.java
+++ b/src/main/java/org/apache/commons/collections4/iterators/UnmodifiableListIterator.java
@@ -43,7 +43,7 @@ public final class UnmodifiableListIterator<E> implements ListIterator<E>, Unmod
      * @return a new unmodifiable list iterator
      * @throws NullPointerException if the iterator is null
      */
-    public static <E> ListIterator<E> umodifiableListIterator(final ListIterator<? extends E> iterator) {
+    public static <E> ListIterator<E> unmodifiableListIterator(final ListIterator<? extends E> iterator) {
         Objects.requireNonNull(iterator, "iterator");
         if (iterator instanceof Unmodifiable) {
             @SuppressWarnings("unchecked") // safe to upcast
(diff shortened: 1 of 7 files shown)


src/main/java/org/apache/commons/collections4/iterators/UnmodifiableListIterator.java
```diff
@@ -53,6 +53,14 @@
         return new UnmodifiableListIterator<>(iterator);
     }
 
+    /**
+     * use "org.apache.commons.collections4.iterators.UnmodifiableListIterator#unmodifiableListIterator" instead.
+     */
+    @Deprecated
+    public static <E> ListIterator<E> umodifiableListIterator(final ListIterator<? extends E> iterator) {
```
kinow, line 60: :+1: thanks!

src/main/java/org/apache/commons/collections4/iterators/UnmodifiableListIterator.java
```diff
@@ -53,6 +53,14 @@
         return new UnmodifiableListIterator<>(iterator);
     }
 
+    /**
+     * use "org.apache.commons.collections4.iterators.UnmodifiableListIterator#unmodifiableListIterator" instead.
```
kinow, line 57: Would need the @deprecated javadoc here too to help IDE's I think?

kinow (approved):
Change looks good, but not sure why `addAll` was removed, @garydgregory any thoughts on this?

src/main/java/org/apache/commons/collections4/iterators/UnmodifiableListIterator.java
cbkurz: Hi thanks for the suggestion, I added a commit for this.

kinow:
Merged! Thanks @cbkurz !

PR #256, opened 2021-09-28
SortedProperties should sort entrySet() as well as keys()

berry120:
At present the `SortedProperties` class overrides & sorts in the `keys()` method, but not the `entrySet()` method. This leads to inconsistencies depending on how the properties are retrieved (for example, when calling `write()` to write the properties to a file, Java 8 uses `keyset()` but Java 11+ uses `entrySet()` - same for other methods like `list()`.)

This PR guarantees the same ordering on `keys()` as `entrySet()`, removing this inconsistency.

(First contribution to this repository, so please feel free to point me in the correct direction if something is wrong!)


garydgregory:
Thank you @berry120

PR #259, opened 2021-10-09
Remove unnecessary Casting

arturobernalg:


kinow:
Travis CI error is not related (not to selt: make it uniform the maven targets executed in GH actions and Travis CI)

PR #265, opened 2021-11-04
[COLLECTIONS-800] Adding partitionBalanced(List,int) method

ClaudioConsolmagno:
See [COLLECTIONS-800](https://issues.apache.org/jira/projects/COLLECTIONS/issues/COLLECTIONS-800)

Implementation Notes:
- Given that the `Partition` class is a private static class not used by anything else I think it's ok to just use isBalanced as a boolean instead of, say, an enum or having a separate brand new private class for this new method.
- Added several comments to clarify the logic, let me know if you think that's too much.
- I didn't know how far I could go when refactoring existing tests so I only half changed them for `partition()` method. 
- Similarly, existing tests all around seem fairly simple but I thought I should have good coverage and test different scenarios for partitionBalanced() hence added the whole `forEach` test cases (I'm a fan of Spock and data driven testing).

### DIFF
--- a/src/main/java/org/apache/commons/collections4/ListUtils.java
+++ b/src/main/java/org/apache/commons/collections4/ListUtils.java
@@ -101,10 +101,12 @@ public class ListUtils {
     private static class Partition<T> extends AbstractList<List<T>> {
         private final List<T> list;
         private final int size;
+        private final boolean isBalanced;
 
-        private Partition(final List<T> list, final int size) {
+        private Partition(final List<T> list, final int size, final boolean isBalanced) {
             this.list = list;
             this.size = size;
+            this.isBalanced = isBalanced;
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/collections4/ListUtils.java
```diff
@@ -117,9 +119,25 @@ private Partition(final List<T> list, final int size) {
...
+                // evenly distribute partitions
+                currentPartitionSize = (int) Math.ceil((double) list.size() / (double) listSize);
+                start = index * currentPartitionSize;
+                // remainder of above is the threshold for which indices greater than will have one less element
+                final int threshold = (list.size() % listSize);
+                // when currentPartitionSize is 1 or threshold is 0 there's nothing to balance
+                // when index hasn't crossed the threshold we don't need balancing yet
+                if (currentPartitionSize > 1 && threshold > 0 && index >= threshold) {
+                    start -= (index - threshold); // adjust start as partitions before threshold have one less element
+                    currentPartitionSize--; // currentPartitionSize is decremented as threshold is crossed
```
LarsBodewig, line 134: Newbie myself, but I wonder if that could be written more simply:
```java
evenlyDistributedSize = ...
if (index < threshold) { // lists before threshold should contain +1
    start = index * (evenlyDistributedSize + 1);
    end = start + (evenlyDistributedSize + 1);
} else {
    start = index * evenlyDistributedSize + threshold; // offset other starting positions by threshold
    end = start + evenlyDistributedSize;
}
```
...or similar (Note: this is untested). The if condition and subtractions look quite complicated in my opinion.

src/main/java/org/apache/commons/collections4/ListUtils.java
```diff
@@ -486,13 +504,52 @@ public static String longestCommonSubsequence(final CharSequence charSequenceA,
...
-        return new Partition<>(list, size);
+        return new Partition<>(list, size, false);
+    }
+
+    /**
+     * Returns consecutive {@link List#subList(int, int) sublists} of a
+     * list, partitioned in a way to balance entries across all sublists. For example,
+     * partitioning a list containing {@code [a, b, c, d, e]} with a partition
+     * size of 3 yields {@code [[a, b, c], [d, e]]} -- an outer list containing
+     * two inner lists of three and two elements, all in the original order. Partitioning
```
LarsBodewig, line 522: I see why you would use the example from `ListUtils.partition`, however this could be confusing IMO if you start comparing what method to use and see no difference so far. I would opt for a more clear example here.

PR #267, opened 2021-11-27
add removeNullElements method in ListUtils

ramananravi:


src/test/java/org/apache/commons/collections4/ListUtilsTest.java
```diff
@@ -500,4 +500,45 @@ public void testSubtractNullElement() {
 
         assertEquals(expected, result);
     }
+    
+    @Test
+    public void testremoveNullElements() {
```
kinow, line 505: The test name would have to be corrected to `testRemoveNullElements` (upper case `R` in remove)

src/main/java/org/apache/commons/collections4/ListUtils.java
```diff
@@ -744,6 +745,23 @@ public static String longestCommonSubsequence(final CharSequence charSequenceA,
...
+     * @throws NullPointerException if the input list is null
+     * @throws IllegalArgumentException if the input list is empty
+     */
+    public static <E> List<E> removeNullElements(final List<E> list) {
+    	Objects.requireNonNull(list, "input list must not be null");
+        if (list.size() == 0) {
+            throw new IllegalArgumentException("input list must not be empty");
+        }
+        return list.stream().filter(Objects::nonNull).collect(Collectors.toList());
+    }
```
kinow, line 763: Not sure if that would be worth adding this method. I think at the moment users can use `ListUtils.removeAll(list, Arrays.astList(null))` (not tested)? While `removeNullElements` could be convenient for some users, we would have to consider the maintenance in commons-collections, and cases where users want to remove not only `null`, but maybe also empty strings, invalid values, etc. Let's wait to hear others :+1: thanks!

src/test/java/org/apache/commons/collections4/ListUtilsTest.java
ramananravi: Yes, that was a typo unnoticed. Thanks a lot for pointing it out. Fixed it now.

src/main/java/org/apache/commons/collections4/ListUtils.java
ramananravi: Yes, that's right. ListUtils.removeAll(list, Arrays.astList(null)) can be used to achieve this, but this seems like a workaround (as we will have to explicitly create a list with just null element and pass it as an input). While on the other hand, thought removeNullElements could be handy for the said purpose. Please let me know in case of any comments. Thanks.

garydgregory: I agree with @kinow that it does not seem worth it. The argument that Bruno's example requires a list to be created seems beside the point, because the PR creates a stream, a lambda, a filter and so on. In my experience, I usually end up with putting a non-null filter in the middle of stream processing which ends up best since I don't have to go back and forth between streams and collections.

ramananravi: Agreed @garydgregory. The very fact that I had to put not-null checks in the middle of iterating a list every time, made me think if there could be a handy util to solve this (so that once done my list would become null-free). Again, I wouldn't say it's a must-have but thought if this could be a good-to-have util. On a side note, any alternate approach suggestions to avoid the said stream-lambda-collector pipeline? Just to know if there's a better way to do the same. Thanks in advance.

Claudenw: How about `ListUtils.select( list, Object::nonNull )` ??

ramananravi: Thanks a lot @Claudenw. One quick follow-up question. Should we consider replacing list.stream().filter(Objects::nonNull).collect(Collectors.toList()) with select( list, Object::nonNull ) or are we saying removeNullElements method itself isn't needed as the same can be achieved using select( list, Object::nonNull ) ? This would help me decide on whether I will have to incorporate the said comments or close this PR altogether. Thanks in advance.

Claudenw: My take was the "removeNullElements" method was not needed because there was a simple solution already available.

ramananravi: Thanks for the suggestion @Claudenw. The only reason I added removeNullElements was to have this util handy (to abstract null checks similar to defaultIfNull util). But if all of us agree this isn't needed, I shall go ahead and close this PR. Thanks again.

src/main/java/org/apache/commons/collections4/ListUtils.java
```diff
@@ -744,6 +745,23 @@ public static String longestCommonSubsequence(final CharSequence charSequenceA,
...
+     *
+     * @param      <E> the element type
+     * @param list the input list, must not be null and must not be empty
+     * @return a new list by removing all null elements from the given list
+     * @throws NullPointerException     if the input list is null
+     * @throws IllegalArgumentException if the input list is empty
+     */
+    public static <E> List<E> removeNullElements(final List<E> list) {
+        Objects.requireNonNull(list, "input list must not be null");
+        if (list.size() == 0) {
```
garydgregory, line 759: This should not be an error IMO. Who cares if nothing was done to the input list after all? 
I still don't see the value, for my money that is.

ramananravi: Thanks for the suggestion, fixed the same now.

Claudenw:
@garydgregory @kinow @ramananravi Are we agreed to close this pull request with "not accepted" or some such comment?

PR #272, opened 2022-01-28
Simplify conditions and avoid extra computations.

arturobernalg:


Claudenw (commented):
LGTM though MultiValueMap is depricated

src/main/java/org/apache/commons/collections4/collection/SynchronizedCollection.java
```diff
@@ -213,7 +213,7 @@ public boolean equals(final Object object) {
             if (object == this) {
                 return true;
             }
-            return object == this || decorated().equals(object);
+            return decorated().equals(object);
```
Claudenw, line 216: This should be a ternary.

There is also an error in the original code in that comparing two difference SynchronizedCollection instances on the same decorated collection will yield a `false` when it should be `true`.

`object.decorated() == this.decorated()`

src/main/java/org/apache/commons/collections4/comparators/FixedOrderComparator.java
```diff
@@ -273,12 +273,11 @@ public boolean equals(final Object object) {
...
             final FixedOrderComparator<?> comp = (FixedOrderComparator<?>) object;
-            return (null == map ? null == comp.map : map.equals(comp.map)) &&
-                   (null == unknownObjectBehavior ? null == comp.unknownObjectBehavior :
-                        unknownObjectBehavior == comp.unknownObjectBehavior &&
-                        counter == comp.counter &&
-                        isLocked == comp.isLocked &&
-                        unknownObjectBehavior == comp.unknownObjectBehavior);
+            return map.equals(comp.map)
+                    && (null == unknownObjectBehavior ? null == comp.unknownObjectBehavior :
+                    unknownObjectBehavior == comp.unknownObjectBehavior
```
Claudenw, line 278: This should probably call `Objects.equals( unknownObjectBehavior, comp.unknownObjectBehavior )`

Claudenw (changes requested):
It feels like all the places where you changed

```
        if (paramTypes == null && args != null
            || paramTypes != null && args == null
            || paramTypes != null && args != null && paramTypes.length != args.length) {
            throw new IllegalArgumentException("Parameter types must match the arguments");
        }
```
to 
```
        if (paramTypes == null && args != null
            || paramTypes != null && args == null
            || paramTypes != null && paramTypes.length != args.length) {
            throw new IllegalArgumentException("Parameter types must match the arguments");
        }
```

Should be a convenience method, though I don't know where to put it.  Leave the code as you have changed it, or move it if you can find a reasonable place to put it.

PR #274, opened 2022-01-29
Use Java 8 and method reference.

arturobernalg:


aherbert:
Change looks good. Can you drop the update to the bloomfilter package (unreleased code). This package is being rewritten and is currently under review in PR #258.

PR #279, opened 2022-02-14
COLLECTIONS-604

Kaammill:
Hi, 
I was thinking about the more uniform safe-null methods in CollectionUtils from jira task: COLLECTIONS-604. In my opinion it would be nice to have one way of treating null-safe methods.  I decided to try wit the approach that the methods that already return Boolean values could be null-save.  I would appreciate all kind of feedback on this task :)

### DIFF
--- a/src/main/java/org/apache/commons/collections4/CollectionUtils.java
+++ b/src/main/java/org/apache/commons/collections4/CollectionUtils.java
@@ -419,16 +419,14 @@ public class CollectionUtils {
      * cost of an additional space complexity O(n).
      * </p>
      *
-     * @param coll1  the first collection, must not be null
-     * @param coll2  the second collection, must not be null
      * @return {@code true} iff the intersection of the collections has the same cardinality
      *   as the set of unique elements from the second collection
-     * @throws NullPointerException if coll1 or coll2 is null
+     * @returns true if first and second collection is null
      * @since 4.0
      */
     public static boolean containsAll(final Collection<?> coll1, final Collection<?> coll2) {
-        Objects.requireNonNull(coll1, "coll1");
-        Objects.requireNonNull(coll2, "coll2");
+        if (coll1 == null && coll2 == null) return true;
+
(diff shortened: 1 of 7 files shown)


src/main/java/org/apache/commons/collections4/CollectionUtils.java
```diff
@@ -419,16 +419,14 @@ private CollectionUtils() {}
...
      *   as the set of unique elements from the second collection
-     * @throws NullPointerException if coll1 or coll2 is null
+     * @returns true if first and second collection is null
      * @since 4.0
      */
     public static boolean containsAll(final Collection<?> coll1, final Collection<?> coll2) {
-        Objects.requireNonNull(coll1, "coll1");
-        Objects.requireNonNull(coll2, "coll2");
+        if (coll1 == null && coll2 == null) return true;
+
```
Simulant87, line 429: The code below will still throw a now no longer documented and expected NullPointerException if coll1 if not null but coll2 is null.
It's the same thing for the change in the method below, and possible more, I haven't read through the whole PR.

Claudenw (changes requested):
I think you need to think about set operations.

what does `containsAny(x,y)` mean if 

- y is null?  If this is a set operation then all X contain the null set.
- x is null? If this is a set operation then the null set does not contain any other set, except perhaps the null set.

I think the same questions need to be answered for all the collections questions.

I would recommend updating the javadoc for each method so that it is clear what the various null values return.

Claudenw:
@Kaammill The more I think about this change the more I think that there is a significant issue.
Changing the code to not throw an NPE when the NPE was documented in the earlier release is (I think) a breaking change and would require a new major version.

Alternatively, a new set of classes that handle Null pointer arguments differently might be valuable.  However, this is a discussion to have on the dev mailing list.

Are you still interested in making these changes?

garydgregory:
Closing: No reply from OP in over a year.

PR #281, opened 2022-03-01
COLLECTIONS-777 JUnit v5 assertThrows no1

nhojpatrick:
Next batch of JUnit v5 upgrades.

### DIFF
--- a/src/test/java/org/apache/commons/collections4/trie/PatriciaTrieTest.java
+++ b/src/test/java/org/apache/commons/collections4/trie/PatriciaTrieTest.java
@@ -58,7 +58,6 @@ public class PatriciaTrieTest<V> extends AbstractSortedMapTest<String, V> {
         return false;
     }
 
-
     public void testPrefixMap() {
         final PatriciaTrie<String> trie = new PatriciaTrie<>();
 
@@ -202,10 +201,10 @@ public class PatriciaTrieTest<V> extends AbstractSortedMapTest<String, V> {
         iterator = map.keySet().iterator();
         map.put("Amber", "Amber");
         Assertions.assertEquals(3, map.size());
-        try {
-            iterator.next();
-            Assertions.fail("CME expected");
-        } catch(final ConcurrentModificationException expected) {}
+
+        final Iterator<String> iterator1 = iterator;
+        Assertions.assertThrows(ConcurrentModificationException.class, () -> iterator1.next());
+
         Assertions.assertEquals("Amber", map.firstKey());
         Assertions.assertEquals("Ammun", map.lastKey());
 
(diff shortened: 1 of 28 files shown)


kinow (commented):
Change looks good to me, great job @nhojpatrick. The commits must be squashed, since they all seem to be related and wouldn't help much having these many commits instead of a single one :+1: 

Thanks
Bruno

nhojpatrick:
Yeah, I'm doing 1 file per commit so easier if i need to rebase, but yes should be squashed when PR merged.

nhojpatrick:
Was unaware of COLLECTIONS-777 but yes that would cover this work. I've squashed myself and prefixed commit with COLLECTIONS-777.

src/test/java/org/apache/commons/collections4/trie/PatriciaTrieTest.java
```diff
@@ -243,28 +242,26 @@ public void testPrefixMap() {
...
-        try {
-            final Object o = map.lastKey();
-            Assertions.fail("got a last key: " + o);
-        } catch(final NoSuchElementException nsee) {}
+
+        final SortedMap<String, String> map3 = map;
+        Assertions.assertThrows(NoSuchElementException.class, () -> map3.firstKey());
+
+        final SortedMap<String, String> map4 = map;
+        Assertions.assertThrows(NoSuchElementException.class, () -> map4.lastKey());
```
kinow, line 263: Maybe keep the comments above. I don't know what first or last key represent in this case, but I assume it could be useful to see that in the test failure output (I guess that's why the original developer put them there).

src/test/java/org/apache/commons/collections4/trie/PatriciaTrieTest.java
```diff
@@ -274,14 +271,13 @@ public void testPrefixMap() {
...
-        try {
-            final Object o = map.lastKey();
-            Assertions.fail("got a last key: " + o);
-        } catch(final NoSuchElementException nsee) {}
+
+        final SortedMap<String, String> map5 = map;
+        Assertions.assertThrows(NoSuchElementException.class, () -> map5.firstKey());
+
+        final SortedMap<String, String> map6 = map;
+        Assertions.assertThrows(NoSuchElementException.class, () -> map6.lastKey());
```
kinow, line 279: Ditto here.

kinow (commented):
There are a few cases here too, as in #282, that I think we better keep the failure messages, in case this is helpful to other devs. Messages like "Should have thrown NPE" are not necessary since it's clear from the context. I had a look at the test method names and javadocs, but didn't find useful replacements for the comments :+1:

src/test/java/org/apache/commons/collections4/trie/PatriciaTrieTest.java
nhojpatrick: I keep getting told to drop the failure message or not check the exception message, so assumed that is the apache commons standard. I can change, but it's slightly confusing as each apache commons pr review appears to have different criterial or standards that I need to follow.

nhojpatrick: done, in next push

src/test/java/org/apache/commons/collections4/trie/PatriciaTrieTest.java
nhojpatrick: ditto

garydgregory:
Is this one ready?

nhojpatrick:
@garydgregory ready for final review.

Or I can merge #282 and #283 and #284 into this branch if needed. But as 20-30 files per pr, I didn't want to wait until everything is done as not sure how long it will take.

PR #306, opened 2022-05-01
[COLLECTIONS-812] fix flaky test due to timestamp comment inserted in…

samabcde:
… Properties methods.

A comment line is inserted after the comment given when using [Properties#save(OutputStream, String)](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Properties.html#save(java.io.OutputStream,java.lang.String)), [Properties#store(OutputStream, String)](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Properties.html#store(java.io.OutputStream,java.lang.String)) and [Properties#store(Writer, String)](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Properties.html#store(java.io.Writer,java.lang.String))
e.g.`#Sat Apr 30 22:00:30 HKT 2022`. Hence the test will fail when actual, expected content are not written in same second. To fix the problem, comparison will ignore this comment

- Additional to #305, also fix `testStoreToPrintWriter()` and `testSetProperty()` which are also using `Properties#store`

- Remove unnecessary `throws IOException` 

@kinow, should we keep using `Properties#save` for compatibility, or just use `Properties#store` and rename the method to `testStore`?

### DIFF
--- a/src/test/java/org/apache/commons/collections4/properties/EmptyPropertiesTest.java
+++ b/src/test/java/org/apache/commons/collections4/properties/EmptyPropertiesTest.java
@@ -29,8 +29,11 @@ import static org.junit.jupiter.api.Assertions.fail;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.OutputStream;
 import java.io.PrintStream;
 import java.io.PrintWriter;
+import java.io.UnsupportedEncodingException;
+import java.io.Writer;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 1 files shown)


kinow (commented):
Superseded by https://github.com/apache/commons-collections/pull/305 , I think :-)

samabcde:
@kinow , see your fix after I completed this, so I will close this PR. But please help to fix `testStoreToPrintWriter()` and `testSetProperty()` which are also using `Properties#store` and hence are also flaky.

PR #308, opened 2022-05-09
COLLECTIONS-770 When creating an iterator chain containing other iterator chains flatten the resulting internal object.

stevebosman-oc:
Contributed on behalf of the @opencastsoftware open source team

I am slightly wary of opening up UnmodifiableIterator, but this seems to solve the performance degradation of nested IteratorChains seen in COLLECTIONS-770

### DIFF
--- a/src/test/java/org/apache/commons/collections4/iterators/IteratorChainTest.java
+++ b/src/test/java/org/apache/commons/collections4/iterators/IteratorChainTest.java
@@ -35,12 +35,16 @@ import org.junit.jupiter.api.Test;
 public class IteratorChainTest extends AbstractIteratorTest<String> {
 
     protected String[] testArray = {
-        "One", "Two", "Three", "Four", "Five", "Six"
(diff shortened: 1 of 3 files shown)


src/test/java/org/apache/commons/collections4/iterators/IteratorChainTest.java
```diff
@@ -170,4 +177,60 @@ public void testEmptyChain() {
...
+            assertEquals( "Iteration value is correct", testValue, iterValue );
+        }
+
+        assertFalse("Iterator should now be empty", iteratorChainOfChains.hasNext());
+
+        try {
+            iteratorChainOfChains.next();
+        } catch (final Exception e) {
+            assertEquals("NoSuchElementException must be thrown", e.getClass(), NoSuchElementException.class);
+        }
```
Claudenw, line 221: Replace this with `assertThrows()`

PR #313, opened 2022-06-15
Remove redundant local variable.

arturobernalg:


kinow (changes requested):
In both cases I found the code before the change easier to read and understand. In the end there might be some bytecode optimization by the JVM so not sure there is really much to gain by changing it.

aherbert:
Certainly in the case of the BloomFilter Shape, the code has been written as a direct translation of the known formulas. Running it through the `javap` decompiler the code is 32 bytes in java byte code and so will be inlined anyway (it is under the 35 byte default). Once the bytecode has been optimised in the JVM I expected there will be no difference with this code revision given that the call to log1p will take longer than the rest of the code put together.

arturobernalg:
Ok, your right. I close the PR

PR #314, opened 2022-06-16
Move counting long predicate

Claudenw:
Fix [COLLECTIONS-826](https://issues.apache.org/jira/browse/COLLECTIONS-826)

### DIFF
--- a/src/main/java/org/apache/commons/collections4/bloomfilter/CountingLongPredicate.java
+++ b/src/main/java/org/apache/commons/collections4/bloomfilter/CountingLongPredicate.java
@@ -26,9 +26,9 @@ import java.util.function.LongPredicate;
  *
  */
 class CountingLongPredicate implements LongPredicate {
-    int idx = 0;
-    final long[] ary;
-    final LongBiPredicate func;
+    private int idx = 0;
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/collections4/bloomfilter/CountingLongPredicate.java
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * A long predicate that applies the test func to each member of the @{code ary} in sequence for each call to @{code test()}.
+ * if the @{code ary} is exhausted, the subsequent calls to to @{code test} are executed with a zero value.
+ * If the calls to @{code test} do not exhaust the @{code ary} the @{code forEachRemaining} method can be called to
+ * execute the @code{text} with a zero value for each remaining @{code idx} value.
+ *
+ */
+class CountingLongPredicate implements LongPredicate {
+    int idx = 0;
```
garydgregory, line 29: Why are these not all private?

Claudenw: because I was not paying attention. It should be fixed now.

PR #316, opened 2022-06-29
COLLECTIONS-823: Modified ArrayCountingBloomFilter.forEachBitMap to be more efficient

Claudenw:
Resolves https://issues.apache.org/jira/projects/COLLECTIONS/issues/COLLECTIONS-823

Modifies `ArrayCountingBloomFilter.forEachBitmap()` to be more efficient. by not creating an array of longs but building each long and calling the consumer in turn.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/bloomfilter/ArrayCountingBloomFilter.java
+++ b/src/main/java/org/apache/commons/collections4/bloomfilter/ArrayCountingBloomFilter.java
@@ -234,7 +234,7 @@ public final class ArrayCountingBloomFilter implements CountingBloomFilter {
         int blocksm1 = BitMap.numberOfBitMaps(shape.getNumberOfBits()) - 1;
         int i = 0;
         long value;
-        // must break final block separate as the the number of bits may not fall on the long boundary
+        // must break final block separate as the number of bits may not fall on the long boundary
         for (int j = 0; j < blocksm1; j++) {
             value = 0;
             for (int k = 0; k < Long.SIZE; k++) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/commons/collections4/bloomfilter/ArrayCountingBloomFilter.java
```diff
@@ -231,7 +231,29 @@ public boolean forEachIndex(IntPredicate consumer) {
     @Override
     public boolean forEachBitMap(LongPredicate consumer) {
         Objects.requireNonNull(consumer, "consumer");
-        return BitMapProducer.fromIndexProducer(this, shape.getNumberOfBits()).forEachBitMap(consumer);
+        int blocksm1 = BitMap.numberOfBitMaps(shape.getNumberOfBits()) - 1;
```
aherbert, line 234: Although it is the same number I would prefer `counts.length` here and not `shape.getNumberOfBits()` since it is the total length of the counts array that is being iterated over.

aherbert, line 234: `final int blocksm1`

aherbert (changes requested):
Changes look OK. Just a few minor fixes required.

src/main/java/org/apache/commons/collections4/bloomfilter/ArrayCountingBloomFilter.java
```diff
@@ -231,7 +231,29 @@ public boolean forEachIndex(IntPredicate consumer) {
...
+        int i = 0;
+        long value;
+        // must break final block separate as the the number of bits may not fall on the long boundary
```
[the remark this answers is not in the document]
Claudenw: done

src/main/java/org/apache/commons/collections4/bloomfilter/ArrayCountingBloomFilter.java
Claudenw: fixed

PR #322, opened 2022-07-15
do refactoring

lexasub:


garydgregory:
Hi @lexasub 
The Bloom filter package is being actively developed ATM, so this PR is clashing with on-going work. Sorry about that.

PR #328, opened 2022-08-09
Collections 827: Add tests using or, and and xor with different length filters.

Claudenw:
Added tests between shapes of different length for "or", "and" and "xor" operations.


src/test/java/org/apache/commons/collections4/bloomfilter/SetOperationsTest.java
```diff
@@ -222,12 +222,26 @@ public final void testOrCardinality() {
...
+
+        filter1 = new SparseBloomFilter(shape, IndexProducer.fromIndexArray(new int[] { 1, 63 }));
+        filter2 = new SparseBloomFilter(shape2, IndexProducer.fromIndexArray(new int[] { 5, 64, 169 }));
         assertEquals(5, SetOperations.orCardinality(filter1, filter2));
         assertEquals(5, SetOperations.orCardinality(filter2, filter1));
+
+        filter1 = new SparseBloomFilter(shape, IndexProducer.fromIndexArray(new int[] { 5, 63 }));
+        filter2 = new SparseBloomFilter(shape2, IndexProducer.fromIndexArray(new int[] { 5, 64, 169 }));
+        assertEquals(4, SetOperations.orCardinality(filter1, filter2));
+        assertEquals(4, SetOperations.orCardinality(filter2, filter1));
```
aherbert, line 244: I think we can reduce the amount of duplicate code with a custom assertion:
```Java
        assertSymmetricOperation(4, SetOperations::orCardinality, filter1, filter2);
    }

    private static void assertSymmetricOperation(int expected, ToIntBiFunction<BloomFilter, BloomFilter> operation,
            BloomFilter filter1, BloomFilter filter2) {
        assertEquals(expected, operation.applyAsInt(filter1, filter2), "op(filter1, filter2)");
        assertEquals(expected, operation.applyAsInt(filter2, filter1), "op(filter2, filter1)");
    }
```

This ensures you only have one place where you specify the expected value and where filter 1 and 2 are swapped as arguments to the assertion.

PR #330, opened 2022-08-22
Implemented clear() for COLLECTIONS-831

Claudenw:
- Defined BloomFilter.clear() to be a reset to unpopulated state.
- Modified BloomFilter implementations.
- Added test case.

Fix for COLLECTIONS-831

### DIFF
--- a/src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBloomFilterTest.java
+++ b/src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBloomFilterTest.java
@@ -183,7 +183,7 @@ public abstract class AbstractBloomFilterTest<T extends BloomFilter> {
         BloomFilter bf1 = createFilter(getTestShape(), from1);
         assertTrue( bf1.cardinality()>0 );
         bf1.clear();
-        assertTrue( bf1.cardinality()==0);
+        assertEquals( 0, bf1.cardinality());
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBloomFilterTest.java
```diff
@@ -177,6 +177,14 @@ public final void testContains() {
         assertFalse(bf1.contains(bf4));
         assertTrue(bf4.contains(bf1));
     }
+    
+    @Test
+    public void testClear() {
+        BloomFilter bf1 = createFilter(getTestShape(), from1);
+        assertTrue( bf1.cardinality()>0 );
+        bf1.clear();
+        assertTrue( bf1.cardinality()==0);
```
kinow, line 186: `assertEquals(0, bf1.cardinality())`

kinow (commented):
@Claudenw https://issues.apache.org/jira/browse/CASSANDRA-831 ? Is that the correct ticket?

kinow (approved):

kinow:
Looks good to me. One suggestion for the unit test assertion, but not a blocker IMO. Thanks!

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBloomFilterTest.java
```diff
@@ -177,6 +177,14 @@ public final void testContains() {
         assertFalse(bf1.contains(bf4));
         assertTrue(bf4.contains(bf1));
     }
+    
+    @Test
+    public void testClear() {
+        BloomFilter bf1 = createFilter(getTestShape(), from1);
+        assertTrue( bf1.cardinality()>0 );
```
aherbert, line 184: This should use the opposite of assertEquals:
```Java
assertNotEquals(0, bf1.cardinality())
```

aherbert:
The updates have introduced some code duplications. Also you did not update to use `assertNotEquals` in the test for the non-empty filter.

aherbert:
@Claudenw after a lot of commits I think the files are back to the state where we started. So the change to use assertEquals is no longer present, and no change to use assertNotEquals.

aherbert:
Closed by 5a31023eae4bf2bd621fa8613699f54842109402

PR #331, opened 2022-08-23
Collections 763: Remove BloomFilter constructors that create initial entry

Claudenw:
Removed constructors that took initial entry.

Added `merge()` methods for IndexProducer and BitMapProducer.

Added `CountingBloomFilter.remove()` methods for IndexProducer and BitMapProducer.
Moved implementations of `ArrayCountingBloomFilter.merge()` to default implementations in `CountingBloomFilter`
Moved implementations of `ArrayCOuntingBloomFilter.remove()` to default implementations in `CountingBloomFilter`

Updated tests accordingly
Added merge() tests.
Added remove() tests.


src/main/java/org/apache/commons/collections4/bloomfilter/CountingBloomFilter.java
```diff
@@ -77,6 +79,92 @@ public interface CountingBloomFilter extends BloomFilter, BitCountProducer {
...
+     *
+     * <p>This method will return {@code true} if the filter is valid after the operation.</p>
+     *
+     * @param other the other Bloom filter
+     * @return {@code true} if the removal was successful and the state is valid
+     * @see #isValid()
+     * @see #add(BitCountProducer)
+     */
+    default boolean merge(final BloomFilter other) {
+        Objects.requireNonNull(other, "other");
```
aherbert, line 98: Could this be changed to:
```Java
return merge((IndexProducer) other);
```

src/main/java/org/apache/commons/collections4/bloomfilter/CountingBloomFilter.java
```diff
@@ -77,6 +79,92 @@ public interface CountingBloomFilter extends BloomFilter, BitCountProducer {
...
+     *
+     * <p>This method will return {@code true} if the filter is valid after the operation.</p>
+     *
+     * @param bitMapProducer the BitMapProducer
+     * @return {@code true} if the removal was successful and the state is valid
+     * @see #isValid()
+     * @see #add(BitCountProducer)
+     */
+    default boolean merge(final BitMapProducer bitMapProducer) {
+        return merge(IndexProducer.fromBitMapProducer(bitMapProducer));
```
aherbert, line 165: `Objects.requireNonNull(indexProducer, "bitMapProducer");` ?

Just to get the correct exception message rather than the message from the IndexProducer.

src/main/java/org/apache/commons/collections4/bloomfilter/CountingBloomFilter.java
```diff
@@ -110,7 +201,52 @@ public interface CountingBloomFilter extends BloomFilter, BitCountProducer {
      * @see #isValid()
      * @see #subtract(BitCountProducer)
      */
-    boolean remove(Hasher hasher);
+    default boolean remove(final Hasher hasher) {
+        Objects.requireNonNull(hasher, "hasher");
+        return subtract(BitCountProducer.from(hasher.uniqueIndices(getShape())));
```
aherbert, line 206: No catch of IOOBE and wrapping to IAE

src/main/java/org/apache/commons/collections4/bloomfilter/CountingBloomFilter.java
```diff
@@ -110,7 +201,52 @@ public interface CountingBloomFilter extends BloomFilter, BitCountProducer {
...
+     *
+     * <p>This method will return {@code true} if the filter is valid after the operation.</p>
+     *
+     * @param bitMapProducer the BitMapProducer to provide the indexes
+     * @return {@code true} if the removal was successful and the state is valid
+     * @see #isValid()
+     * @see #subtract(BitCountProducer)
+     */
+    default boolean remove(final BitMapProducer bitMapProducer) {
+        return remove(IndexProducer.fromBitMapProducer(bitMapProducer));
```
aherbert, line 248: This double wraps the input: `BitMapProducer -> IndexProducer -> BitCountProducer`. This could be a future optimisation to go direct: `BitMapProducer -> BitCountProducer`. The method could be done in an internal helper class.

src/main/java/org/apache/commons/collections4/bloomfilter/SimpleBloomFilter.java
```diff
@@ -133,12 +81,9 @@ public SimpleBloomFilter copy() {
...
-     * @param indexProducer the IndexProducer to merge from.
-     * @throws IllegalArgumentException if producer sends illegal value.
-     */
-    private void merge(IndexProducer indexProducer) {
+    @Override
+    public boolean merge(IndexProducer indexProducer) {
+        Objects.requireNonNull(indexProducer, "indexProducer");
         indexProducer.forEachIndex(idx -> {
             if (idx < 0 || idx >= shape.getNumberOfBits()) {
                 throw new IllegalArgumentException(String.format(
```
aherbert, line 89: I note that here the index range uses a closed range [0, x-1]. Other exception message use an open range on the upper bound [0, x).

src/main/java/org/apache/commons/collections4/bloomfilter/SparseBloomFilter.java
```diff
@@ -140,12 +78,9 @@ private boolean add(int idx) {
...
 
-    /**
-     * Performs a merge using an IndexProducer.
-     * @param indexProducer the IndexProducer to merge from.
-     * @throws IllegalArgumentException if producer sends illegal value.
-     */
-    private void merge(IndexProducer indexProducer) {
+    @Override
+    public boolean merge(IndexProducer indexProducer) {
+        Objects.requireNonNull(indexProducer, "indexProducer");
```
aherbert, line 83: Note:

This merge method (and others such as remove/add/subtract) throw exceptions. However the filter will be left in a corrupted state if some of the indices have been processed before the exception. Implementing these methods as an atomic transaction would be slower and more involved. So I suggest that the non-atomic nature of the bulk updates to a filter are commented, perhaps at the package level, as a design decision to achieve maximum performance under correct usage.

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBloomFilterTest.java
```diff
@@ -186,8 +179,10 @@ public final void testContains() {
     @Test
     public final void testEstimateIntersection() {
 
-        final BloomFilter bf = createFilter(getTestShape(), from1);
-        final BloomFilter bf2 = createFilter(getTestShape(), bigHasher);
+        final BloomFilter bf = createEmptyFilter(getTestShape());
+        bf.merge(from1);
+        final BloomFilter bf2 = createEmptyFilter(getTestShape());
+        bf2.merge(from1);
```
aherbert, line 185: `bf2` should merge `bigHasher`

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBloomFilterTest.java
```diff
@@ -62,49 +62,34 @@ protected final Shape getTestShape() {
...
     protected abstract T createEmptyFilter(Shape shape);
 
-    /**
-     * Create the BloomFilter implementation we are testing.
-     *
-     * @param shape the shape of the filter.
-     * @param hasher the hasher to use to create the filter.
-     * @return a BloomFilter implementation.
-     */
-    protected abstract T createFilter(Shape shape, Hasher hasher);
```
aherbert, line 72: Why not maintain these methods as non-abstract, e.g.
```Java
protected T createFilter(Shape shape, Hasher hasher) {
    T f = createEmptyFilter(Shape shape);
    f.merge(hasher);
    return f;
}
```

Then you do not have to change all the existing tests to create and then merge the argument.

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractHasherTest.java
```diff
@@ -97,8 +97,8 @@ public void testUniqueIndex() {
         List<Integer> full = Arrays.stream(producer.asIndexArray()).boxed().collect(Collectors.toList());
         producer = hasher.uniqueIndices(shape);
         List<Integer> unique = Arrays.stream(producer.asIndexArray()).boxed().collect(Collectors.toList());
-        assertTrue( full.size() > unique.size() );
-        Set<Integer> set = new HashSet<Integer>( unique );
-        assertEquals( set.size(), unique.size() );
+        assertTrue(full.size() > unique.size());
+        Set<Integer> set = new HashSet<Integer>(unique);
```
aherbert, line 101: Note: You can use the diamond operator `<>` in generics so you do not require the `<Integer>`

src/test/java/org/apache/commons/collections4/bloomfilter/DefaultBloomFilterTest.java
```diff
@@ -196,6 +196,7 @@ public AbstractDefaultBloomFilter copy() {
             result.indices.addAll(indices);
             return result;
         }
+
```
aherbert, line 199: Remove the extra line

src/test/java/org/apache/commons/collections4/bloomfilter/SetOperationsTest.java
```diff
@@ -49,52 +49,63 @@ private static void assertSymmetricOperation(double expected, ToDoubleBiFunction
         assertEquals(expected, operation.applyAsDouble(filter2, filter1), "op(filter2, filter1)");
     }
 
+    private BloomFilter forTest(Shape shape, Hasher hasher) {
```
aherbert, line 52: I would name these `createFilter` to be self documenting.

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBloomFilterTest.java
Claudenw: If this is a common pattern then why are we removing it from the interface?  I think the addition of the 2 merge functions is a good one, but I am still not convinced that the move from a simple constructor to 2 methods is a good one.  I think the convenience constructors should remain.

aherbert: This is a common pattern for _testing_.

Out of curiosity I would ask you to search your use cases in your own production code to see how often you create a filter and then add only item immediately.

I would think this is equivalent to having an ArrayList constructor that adds 1 item. This is not in the JDK. There are better alternatives for such a case such as a singleton list or using Arrays.asList.

On top of this the item you are adding will require some source of indices (e.g. a Hasher, another filter, etc). So there will already be some lines of code to create that from the actual Object, or collection of Objects, that are being used to initialise the filter. 

If you find use cases for this then we can add a more suitable functionality in later releases, or perhaps this exact functionality. For now the purpose is to reduce the initial public API to a manageable release and then receive feedback from the community on what features to enhance.

src/main/java/org/apache/commons/collections4/bloomfilter/CountingBloomFilter.java
```diff
@@ -77,6 +79,92 @@ public interface CountingBloomFilter extends BloomFilter, BitCountProducer {
...
+     * @see #isValid()
+     * @see #add(BitCountProducer)
+     */
+    default boolean merge(final Hasher hasher) {
+        Objects.requireNonNull(hasher, "hasher");
+        try {
+            return add(BitCountProducer.from(hasher.uniqueIndices(getShape())));
+        } catch (IndexOutOfBoundsException e) {
+            throw new IllegalArgumentException(
+                    String.format("Filter only accepts values in the [0,%d) range", getShape().getNumberOfBits()));
```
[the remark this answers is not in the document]
Claudenw: fixed

src/main/java/org/apache/commons/collections4/bloomfilter/CountingBloomFilter.java
```diff
@@ -77,6 +79,92 @@ public interface CountingBloomFilter extends BloomFilter, BitCountProducer {
...
+     * @see #isValid()
+     * @see #add(BitCountProducer)
+     */
+    default boolean merge(final IndexProducer indexProducer) {
+        Objects.requireNonNull(indexProducer, "producer");
+        try {
+            return add(BitCountProducer.from(indexProducer));
+        } catch (IndexOutOfBoundsException e) {
+            throw new IllegalArgumentException(
+                    String.format("Filter only accepts values in the [0,%d) range", getShape().getNumberOfBits()));
```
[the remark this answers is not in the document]
Claudenw: fixed

src/main/java/org/apache/commons/collections4/bloomfilter/CountingBloomFilter.java
Claudenw: fixed

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBloomFilterTest.java
Claudenw: fixed

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBloomFilterTest.java
Claudenw: changed

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractHasherTest.java
Claudenw: changed

src/test/java/org/apache/commons/collections4/bloomfilter/DefaultBloomFilterTest.java
Claudenw: removed

src/test/java/org/apache/commons/collections4/bloomfilter/SetOperationsTest.java
Claudenw: done

pom.xml
```diff
@@ -588,9 +588,9 @@
     <commons.jira.pid>12310465</commons.jira.pid>
     <!-- The RC version used in the staging repository URL. -->
     <commons.rc.version>RC1</commons.rc.version>
-    <checkstyle.version>3.1.2</checkstyle.version>
+    <checkstyle.version>3.2.0</checkstyle.version>
```
aherbert, line 591: The changes to pom.xml and changes.xml should not be here.

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBloomFilterTest.java
```diff
@@ -93,16 +94,30 @@ protected final Shape getTestShape() {
...
 
     @Test
-    public void testConstructWitBitMapProducer() {
+    public void testMergeWithHasher() {
+        // value too large
+        final BloomFilter f = createEmptyFilter(getTestShape());
+        f.merge(from1);
+        int[] idx = f.asIndexArray();
+        assertEquals(getTestShape().getNumberOfHashFunctions(), idx.length);
+        for (int i=0; i<idx.length; i++) {
```
aherbert, line 114: Whitespace:
```Java
for (int i = 0; i < idx.length; i++) {
    assertEquals(i + 1, idx[i]);
```

src/test/java/org/apache/commons/collections4/bloomfilter/SparseBloomFilterTest.java
```diff
@@ -33,66 +33,31 @@ protected SparseBloomFilter createEmptyFilter(final Shape shape) {
 
     @Override
     protected SparseBloomFilter createFilter(final Shape shape, final Hasher hasher) {
-        return new SparseBloomFilter(shape, hasher);
```
aherbert, line 36: Since these methods just do the same as `createEmptyFilter(Shape)` then `merge(source)` I do not understand why they have to be overridden. The implementation can be in the Abstract test:
```Java
protected BloomFilter createFilter(final Shape shape, final Hasher hasher) {
    BloomFilter bf = createEmptyFilter(shape);
    bf.merge(hasher);
    return bf;
}
```
Is it because the return argument type is incorrect? Then this can be fixed with:
```Java
protected SparseBloomFilter createFilter(final Shape shape, final Hasher hasher) {
    return (SparseBloomFilter) super.createFilter(shape, hasher);
}
```
Doing it this way ensures all filters are using the same create and merge method to populate the new filter.

src/test/java/org/apache/commons/collections4/bloomfilter/SparseBloomFilterTest.java
```diff
@@ -33,66 +33,31 @@ protected SparseBloomFilter createEmptyFilter(final Shape shape) {
...
-                return new SparseBloomFilter(new SimpleBloomFilter(shape, hasher));
-            }
-        };
-        executeNestedTest(nestedTest);
-    }
 
     @Test
     public void testBitMapProducerEdgeCases() {
         int[] values = { 1, 2, 3, 4, 5, 6, 7, 8, 9, 65, 66, 67, 68, 69, 70, 71 };
-        BloomFilter bf = createFilter(getTestShape(), IndexProducer.fromIndexArray(values));
```
aherbert, line 95: Why not use `createFilter` here. You have already reinstated the method. You should roll back the changes throughout the test suite where you previously dropped the method.

Your current diff seems to have too many changes. I think you should:

- branch from master
- copy all the modified classes in `main/java/...` over the existing classes
- modify only the 'createFilter' methods in the test classes and any other location that previously used the removed constructors.
- set your upstream to this PR branch in your github and force push the changes. That should clear all your current git history and just introduce the desired changes.

src/main/java/org/apache/commons/collections4/bloomfilter/BloomFilter.java
```diff
@@ -60,6 +60,11 @@ public interface BloomFilter extends IndexProducer, BitMapProducer {
      */
     Shape getShape();
 
+    /**
```
aherbert, line 63: Why is the clear method being added here? Your PR branch should be rebased on the current master as this change should not be in this PR.

src/test/java/org/apache/commons/collections4/properties/EmptyPropertiesTest.java
```diff
@@ -31,16 +31,39 @@
...
 import java.util.HashMap;
 import java.util.Properties;
-
 import org.apache.commons.io.input.NullReader;
 import org.apache.commons.lang3.ArrayUtils;
 import org.junit.jupiter.api.Test;
 
 public class EmptyPropertiesTest {
 
+    /**
```
aherbert, line 44: You should not be including any changes to this package.

aherbert (changes requested):
Hi @Claudenw 

I think you have quite a lot of extra changes in your diff. Since this is a simple change you should just start again from the current master. You can then copy all the modified class files from `main/java` over and modify the appropriate test files to handle the lack of constructors. This should be mainly done by a change to the AbstractBloomFilterTest to use `createEmptyFilter` and `merge` in each of the `createFilter` methods.

If you do this locally on a new branch you can set the upstream to your github remote associated with this PR. A force push will remove your git history from this PR and update with a single atomic commit with the changes. I.e. you do not need a new PR, just a new diff pushed into this one.

When you are done you should be able to view the diff under the [Files changed](https://github.com/apache/commons-collections/pull/331/files) tab in the Github UI. This is what I am using to see the changes.

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBloomFilterTest.java
```diff
@@ -316,20 +345,20 @@ public final void testMerge() {
...
-                Shape.fromKM(getTestShape().getNumberOfHashFunctions(), 3 * Long.SIZE),
-                new IncrementingHasher(Long.SIZE * 2, 1));
+        BloomFilter bf6 = new SparseBloomFilter(
+                Shape.fromKM(getTestShape().getNumberOfHashFunctions(), 3 * Long.SIZE));
+        bf6.merge(new IncrementingHasher(Long.SIZE * 2, 1));
         assertThrows(IllegalArgumentException.class, () -> bf1.merge(bf6));
     }
 
-    private static void assertIndexProducerConstructor(Shape shape, int[] values, int[] expected) {
+    private void assertIndexProducerConstructor(Shape shape, int[] values, int[] expected) {
```
aherbert, line 359: This method name is now incorrect. Perhaps `assertIndexProducerMerge`

src/test/java/org/apache/commons/collections4/bloomfilter/DefaultBloomFilterTest.java
```diff
@@ -163,28 +123,36 @@ public boolean merge(BloomFilter other) {
                             String.format("Value in list %s is less than 0", indices.first()));
                 }
             }
+        }
+        @Override
```
aherbert, line 127: Requires an extra blank line

src/test/java/org/apache/commons/collections4/bloomfilter/EnhancedDoubleHasherTest.java
```diff
@@ -43,39 +43,39 @@ protected int getHasherSize(Hasher hasher) {
     @Test
     public void testByteConstructor() {
         // single value become increment.
-        EnhancedDoubleHasher hasher = new EnhancedDoubleHasher( new byte[] { 1 } );
-        assertEquals( 0, hasher.getInitial() );
-        assertEquals( 0x01_00_00_00_00_00_00_00L, hasher.getIncrement() );
+        EnhancedDoubleHasher hasher = new EnhancedDoubleHasher(new byte[] { 1 });
```
aherbert, line 46: These formatting changes should not be part of this PR. I have applied formatting to master so you can rebase and should pick up the changes.

src/test/java/org/apache/commons/collections4/bloomfilter/SetOperationsTest.java
```diff
@@ -49,52 +49,63 @@ private static void assertSymmetricOperation(double expected, ToDoubleBiFunction
         assertEquals(expected, operation.applyAsDouble(filter2, filter1), "op(filter2, filter1)");
     }
 
+    private BloomFilter createTest(Shape shape, Hasher hasher) {
```
aherbert, line 52: These should be more appropriately named, e.g. `createFilter`

src/test/java/org/apache/commons/collections4/bloomfilter/SparseBloomFilterTest.java
aherbert: This comment has not been addressed as all the `createEmptyFilter` followed by `merge` does not seem required.

aherbert (changes requested):
That looks better. There are still a few things to fix. Thanks.

src/main/java/org/apache/commons/collections4/bloomfilter/SparseBloomFilter.java
Claudenw: added paragraph in package-info.

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBloomFilterTest.java
Claudenw: fixed

src/test/java/org/apache/commons/collections4/bloomfilter/DefaultBloomFilterTest.java
Claudenw: fixed

src/main/java/org/apache/commons/collections4/bloomfilter/SimpleBloomFilter.java
Claudenw: fixed

src/test/java/org/apache/commons/collections4/bloomfilter/SparseBloomFilterTest.java
Claudenw: fixed

src/test/java/org/apache/commons/collections4/bloomfilter/DefaultBloomFilterTest.java
```diff
@@ -163,28 +123,30 @@ public boolean merge(BloomFilter other) {
...
                 }
             }
-            return true;
         }
 
         @Override
-        public int cardinality() {
-            return indices.size();
+        public boolean merge(IndexProducer indexProducer) {
+            boolean result = indexProducer.forEachIndex((x) -> {
```
aherbert, line 130: You do not need parentheses around the `(x)` for single argument lambda functions.

aherbert (changes requested):
Hi @Claudenw 

The changes look good.

I noticed a few typos and some whitespace issues.

There `merge(Hasher)` method of BloomFilter has code but the similar `remove(Hasher)` calls another method. The merge should delegate as well.

You have added some new tests. These use a linear series of bits. I think the test should not use a series just to ensure that the methods can interleave on and off bits during translations.

aherbert:
@Claudenw I merged this locally and was doing some minor code clean-up. It seems that there is a flaky test:

DefaultIndexProducerTest.testAsIndexArray

In AbstractIndexProducerTest.testAsIndexArray the producer is written to a list using asIndexArray. Then the producer forEach is called to remove entries from the list. However the default implementation of IndexProducer.asIndexArray removes duplicates (as it uses a BitSet to collate the indices). So the List<Integer> can have less entries than the number produced by forEach (which contains the duplicates).

You are generating 10 values in the range [0, 512). The likelihood of no duplicates is (512)(511)(510)...(504)(503) / 512^10 = 0.915. So the probability of a duplicate is around 8.5%. If you checkout this current PR branch and run the DefaultIndexProducerTest repeatedly you will need to do it about 10 times to see a failure. This is low enough that the 4 CI builds may not hit a failure.

Q. Should asIndexArray eliminate duplicates? If this is the correct functionality then I can fix this test to handle the duplicates in forEach and complete the merge.

Claudenw:
there is a method in defaultindexprovidertest that will remove dups from
the original array and sort it so that assert array equals can work.

In other tests I ran the main test in a loop 5k times to try to hit such
problems.  I must have missed this one

aherbert:
I have found the index sorting and duplicate elimination routines. I know how to fix the test. The question is whether the test is indicating a problem. I think it is.

The default implementation of:
- IndexProducer.forEach _will_ return duplicates
- IndexProducer.asIndexArray _will not_ return duplicates

Is this the functionality we wish to support? If so then it should be documented. Currently there is no indication in the javadoc that asIndexArray is either sorted or distinct.

I can fix the test; I can add the documentation; but should this be done?

Looking at the overrides for asIndexArray:

- ArrayCountingBloomFilter: _will not_ return duplicates
- EnhancedDoubleHasher: _will_ return duplicates
- HasherCollection: _will_ return duplicates

Note that the two hashers have a code comment that they **need** to return duplicates.

So we have a contradictory functionality. BloomFilters and a default IndexProducer will produce sorted distinct indices for asIndexArray. Hashers will be unsorted and may contain duplicates. The HasherCollection can output a complete jumble of indices which make no practical sense. It could be modified to make the indices from each hasher unique, then duplicates will indicate multiple hashers produced the same index. However this is not supported in the forEach method and would be a performance overhead. Currently the HasherCollection thus acts as if one item but it is built from multiple hashers. This should be in the documentation to prevent the mistake that the collection will act as if it represents multiple items.

So is the suggested fix:

- document that the **default** asIndexArray method will output distinct indices (current behaviour, but no contract for overrides)
- update the test in DefaultIndexProducerTest.testAsIndexArray to remove duplicates from the expected array. This will then enforce the **default** behaviour that the indices are distinct

Note: this does not mandate that asIndexArray must **always** return no duplicates, only that the default implementation does.

Or:

- document the asIndexArray method to state that indices may be duplicates (current behaviour)
- update the test in DefaultIndexProducerTest.testAsIndexArray to be robust to duplicates

I would favour the second approach as it avoids a commitment to certain functionality and highlights that duplicates are possible. This has minimum impact on the rest of the code. I am happy to make the second change now and finish this merge.

Claudenw:
IndexProducers may return duplicates and make no order guarantees.  (there used to be an order guarantee but we removed that).

Hasher based IndexProducers, by their nature, generally return unordered and possible duplicate values.  There is a hasher method to produce an IndexProducer that guaranteed uniqueness.

BloomFilter based IndexProducers, by their nature, generally return ordered and unique values, though I can think of implementations where the order may not be true, we don't have one.

The default implementation of IndexProducer uses BitSet in its implementation to simplify the code to produce the index list.  So the uniqueness is an artefact of the implementation.  If you have a fast implementation that can take the forEachIndex() and convert it to an array without imposing the uniqueness constraint then please implement that.

In short I concur with your assessment.

The HasherCollection is intended to simplify creation of some filters.  In practice it is that same as calling functions with hasher arguments once for each hasher in the collection.  Due to the difference between classes of Bloom filters (e.g. standard, counting, stable) there are times when the duplicates are required.

HasherCollections work well in distributed systems where an object is represented in a Bloom filter by hashes of multiple properties.  A for query across the systems a HasherCollection is constructed and passed to the endpoints.  The endpoints can then build filters based on the shape at the endpoint.  So the specific back end systems do not have to agree on shape, but do agree on Hash algorithm.

aherbert:
I have merged this in commit: 9a58c1bbdf7c65e71b33ab1ba406f1de5a8191df

I have made some further modifications. I updated the IndexProducer to document that the forEach and asIndexArray that `Indices ordering and uniqueness is not guaranteed`.

I have refactored the IndexProducer test so that the distinct and ordering behaviour can be asserted. Please check the current master. I found one bug which is currently disabled in BitCountProducerFromIndexProducerTest. If you create a BitCountProducer from an IndexProducer that has duplicates then the counts are not correct. Can you have a look and consider how to fix this, either by documentation that duplicate indices are ignored, or by actually correcting the counts.

Since an IndexProducer makes no guarantee on duplicates it may be fair to just state that the BitCountProducer will create a count of 1 even in the event of duplicates.

PR #335, opened 2022-09-10
[Collections-834][bloom filters] bit count producer operation is not clearly defined

Claudenw:
fix for COLLECTIONS-384

Document bit count producer operation in the BitCountProducer.java file.

Add unit tests for all BitCountProducers.


Claudenw:
@aherbert Described the BitCountProducer contract in the BitCountProducer javadoc, and added tests to show that they work as expected.  I hope I have not made too much of a hash of it.

src/test/java/org/apache/commons/collections4/bloomfilter/BitCountProducerFromDefaultIndexProducerTest.java
```diff
@@ -41,24 +37,19 @@ protected int getBehaviour() {
...
+        return new int[][]{{0, 1}, {63, 1}, {1, 1}, {1, 1}, {64, 1}, {127, 1}, {128, 1}};
+    }
 
-        producer.forEachCount((i, v) -> {
-            m.put(i, v);
-            return true;
-        });
+    @Override
+    protected int[] getExpectedIndex() {
+        return expected;
```
aherbert, line 47: If the Abstract test is modified as suggested to separately test behaviour and the expected indices then this can `return data`.

src/test/java/org/apache/commons/collections4/bloomfilter/BitCountProducerFromIntArrayTest.java
```diff
@@ -0,0 +1,54 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.collections4.bloomfilter;
+
+public class BitCountProducerFromIntArrayTest extends AbstractBitCountProducerTest {
+
+    int[] data = {6, 8, 1, 2, 4, 4, 5};
+    int[] expected = {1, 2, 4, 5, 6, 8};
```
aherbert, line 22: Updating the abstract test removes the requirement for `expected`

src/test/java/org/apache/commons/collections4/bloomfilter/BitCountProducerFromHasherTest.java
```diff
@@ -0,0 +1,47 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.collections4.bloomfilter;
+
+public class BitCountProducerFromHasherTest extends AbstractBitCountProducerTest {
+
+    @Override
+    protected BitCountProducer createProducer() {
+        return BitCountProducer.from(new IncrementingHasher(0, 1).indices(Shape.fromKM(17, 72)));
```
aherbert, line 23: I would update the test to use something other than IncrementingHasher(0, 1), perhaps IncrementingHasher(3, 1), or IncrementingHasher(3, 2). This would check a non-sequential set of indices.

src/test/java/org/apache/commons/collections4/bloomfilter/BitCountProducerFromSparseBloomFilterTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+    protected Shape shape = Shape.fromKM(17, 72);
+
+    @Override
+    protected BitCountProducer createProducer() {
+        Hasher hasher = new IncrementingHasher(0, 1);
+        BloomFilter bf = new SparseBloomFilter(shape);
+        bf.merge(hasher);
+        return BitCountProducer.from(bf);
+
```
aherbert, line 29: Remove empty line

aherbert (changes requested):
Hi @Claudenw,

Thanks for all the new test data. I had failed to spot previously that the expected indices from an IndexProducer are not actually checked. I think these changes address the issue with the disabled test in BitCountProducerFromIndexProducerTest and strengthen the test suite.

You test the expected indices using an array that mandates the order. However we already test the order behaviour separately. I suggested a change to only use the expected indices to detect correct/incorrect actual indices. The remaining behaviour tests will detect ordering and duplication problems.

I provided an implementation for AbstractBitCountProducerTest that matches the AbstractIndexProducerTest. This checks consistency between the indices output from the two interface forEach methods. Currently it includes new behaviour flags for forEachCount. However the test suite passes on the assumption the forEachIndex and forEachCount have the same behaviour. So maybe we can drop them. They could be added in the future if an implementation is created with different behaviour for the two methods.

I would change some of the tests from using IncrementingHasher(0, 1) to something else. This avoids a sequence from 0 which is simplistic test data. The data is easy to create with a stream:
```Java
    // For: IncrementingHasher(3, 2) and Shape.fromKM(17, 72)
    IntStream.iterate(3, i -> i + 2).limit(17).toArray();
```
For multiple hashers you can collate them:
```Java
    // For: IncrementingHasher(3, 2), IncrementingHasher(0, 1), and Shape.fromKM(17, 72)
    IntStream.concat(
        IntStream.iterate(3, i -> i + 2).limit(17),
        IntStream.iterate(0, i -> i + 1).limit(17)
    ).distinct().toArray();
```

src/main/java/org/apache/commons/collections4/bloomfilter/BitCountProducer.java
```diff
@@ -47,7 +60,7 @@ default boolean forEachIndex(IntPredicate predicate) {
 
     /**
      * Creates a BitCountProducer from an IndexProducer.  The resulting
-     * producer will count each enabled bit once.
+     * producer will return every index from the IndexProducer with a count of 1.
```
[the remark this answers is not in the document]
Claudenw: added with reference to class javaodc

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractIndexProducerTest.java
```diff
@@ -93,6 +92,12 @@ int[] toArray() {
      */
     protected abstract int getBehaviour();
 
+    protected abstract int[] getExpectedIndex();
```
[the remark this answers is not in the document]
Claudenw: added

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractIndexProducerTest.java
```diff
@@ -93,6 +92,12 @@ int[] toArray() {
...
+
+    protected int[] getExpectedForEach() {
```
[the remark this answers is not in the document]
Claudenw: removed

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractIndexProducerTest.java
```diff
@@ -93,6 +92,12 @@ int[] toArray() {
...
+        return getExpectedIndex();
+    }
+
     @Test
     public final void testForEachIndex() {
```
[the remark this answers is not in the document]
Claudenw: renamed

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractIndexProducerTest.java
```diff
@@ -144,12 +148,13 @@ public final void testBehaviourAsIndexArray() {
             long count = Arrays.stream(actual).distinct().count();
             Assertions.assertEquals(count, actual.length);
         }
+        int[] expected = getExpectedIndex();
```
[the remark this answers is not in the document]
Claudenw: moved to testAsIndexArray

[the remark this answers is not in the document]
Claudenw: removed

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractIndexProducerTest.java
```diff
@@ -161,6 +166,8 @@ public final void testBehaviourForEach() {
             long count = Arrays.stream(actual).distinct().count();
             Assertions.assertEquals(count, actual.length);
         }
+        int[] expected = getExpectedForEach();
```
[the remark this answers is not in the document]
Claudenw: Moved to its own testForEachIndex

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractIndexProducerTest.java
```diff
@@ -149,7 +181,6 @@ public final void testBehaviourAsIndexArray() {
     @Test
     public final void testBehaviourForEach() {
         int flags = getBehaviour();
-        Assumptions.assumeTrue((flags & (FOR_EACH_ORDERED | FOR_EACH_DISTINCT)) != 0);
```
aherbert, line 152: Reinstate the Assumptions

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBitCountProducerTest.java
```diff
@@ -16,35 +16,29 @@
...
+     * @return an array of integer pairs comprising the index and the expected count for the index.
      */
-    public static BitCountConsumer TRUE_CONSUMER = new BitCountConsumer() {
-
-        @Override
-        public boolean test(int index, int count) {
-            return true;
-        }
-    };
+    protected abstract int[][] getExpectedBitCount();
```
aherbert, line 41: This method may not need to be abstract. In most cases where ordering and duplicates are not mandated then this implementation will be fine:
```Java
    protected int[][] getExpectedBitCount() {
        return Arrays.stream(getExpectedIndices()).mapToObj(x -> new int[] {x, 1}).toArray(int[][]::new);
    }
```

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractIndexProducerTest.java
```diff
@@ -93,8 +98,36 @@ int[] toArray() {
      */
     protected abstract int getBehaviour();
 
+    /**
+     * Creates an array of expected indices.
+     * @return an array of expected indices.
+     */
+    protected abstract int[] getExpectedIndices();
```
aherbert, line 105: Note that this method requires assumptions. The expected indices are dependent on the producer returned from createProducer, which is ultimately dependent on the shape of the Bloom filter.

src/test/java/org/apache/commons/collections4/bloomfilter/IndexProducerFromIntArrayTest.java
```diff
@@ -18,19 +18,32 @@
...
         // Delegates to the default asIndexArray which is distinct and ordered
         return AS_ARRAY_DISTINCT | AS_ARRAY_ORDERED;
     }
+
+    @Override
+    protected int[] getExpectedIndices() {
+        return expected;
+    }
+
+//    @Override
```
aherbert, line 45: Remove commented out code

src/test/java/org/apache/commons/collections4/bloomfilter/BitCountProducerFromIntArrayTest.java
```diff
@@ -0,0 +1,49 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.collections4.bloomfilter;
+
+public class BitCountProducerFromIntArrayTest extends AbstractBitCountProducerTest {
+
+    int[] data = {6, 8, 1, 2, 4, 4, 5};
+
```
aherbert, line 22: Remove extra line

aherbert (changes requested):
I suggested some changes to the AbstractBitCountProducerTest to assert on the behaviour flags separately from the bit count totals. Can you add these? Please see [this note](https://github.com/apache/commons-collections/pull/335#discussion_r972942396).

src/test/java/org/apache/commons/collections4/bloomfilter/BitCountProducerFromIntArrayTest.java
Claudenw: fixed

src/test/java/org/apache/commons/collections4/bloomfilter/IndexProducerFromIntArrayTest.java
Claudenw: fixed

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractIndexProducerTest.java
Claudenw: fixed

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractIndexProducerTest.java
Claudenw: fixed

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractIndexProducerTest.java
```diff
@@ -93,8 +98,36 @@ int[] toArray() {
...
+
+    @Test
+    public final void testAsIndexArrayValues() {
+        List<Integer> lst = new ArrayList<>();
```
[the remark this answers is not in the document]
Claudenw: fixed

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractIndexProducerTest.java
```diff
@@ -43,7 +48,7 @@ public abstract class AbstractIndexProducerTest {
     /**
      * An expandable list of int values.
      */
-    private static class IntList {
+    protected static class IntList {
```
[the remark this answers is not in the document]
Claudenw: Must be protected for AbstractBitCountProducer usage

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBitCountProducerTest.java
```diff
@@ -16,35 +16,29 @@
...
 
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import org.apache.commons.collections4.bloomfilter.BitCountProducer.BitCountConsumer;
 import org.junit.jupiter.api.Test;
 
 public abstract class AbstractBitCountProducerTest extends AbstractIndexProducerTest {
```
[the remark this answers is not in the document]
Claudenw: I don't know what happened.  I have, I think, fixed it now.

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractIndexProducerTest.java
```diff
@@ -134,7 +167,6 @@ public final void testConsistency() {
     @Test
     public final void testBehaviourAsIndexArray() {
         int flags = getBehaviour();
-        Assumptions.assumeTrue((flags & (AS_ARRAY_ORDERED | AS_ARRAY_DISTINCT)) != 0);
```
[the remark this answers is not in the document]
Claudenw: fixed

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractIndexProducerTest.java
```diff
@@ -93,8 +98,36 @@ int[] toArray() {
...
+        List<Integer> lst = new ArrayList<>();
+        Arrays.stream(createProducer().asIndexArray()).boxed().forEach( lst::add );
+        for (int i : getExpectedIndices()) {
+            assertTrue( lst.contains(i), "Missing "+i );
+        }
+    }
+
     @Test
     public final void testForEachIndex() {
+        //IndexProducer producer = createProducer();
```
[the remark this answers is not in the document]
Claudenw: fixed

src/test/java/org/apache/commons/collections4/bloomfilter/BitCountProducerFromSparseBloomFilterTest.java
Claudenw: fixed

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractIndexProducerTest.java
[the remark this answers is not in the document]
aherbert: When I checked you had not pushed the entire AbstractBitCountProducerTest and so it was not (yet) used in that class. I see the changes now. Let me know when you have fixed everything and I'll review again.

aherbert (changes requested):
Hi @Claudenw 

I've not looked at everything here. I found an issue where the update to the documentation of `BitCountProducer` is not correct (given the current implementations). See my test that is checking the count of indices output from the IndexProducer. It fails for `ArrayCountingBloomFilter`.

I don't think we want to change the implementations. We should change the documentation to be less strict. You can then remove some of the updates to the tests as we do not need `int[] getExpectedForEach`. It is only overridden in `BitCountProducerFromHasherTest` anyway.

Since you have not changed any of `src/main` besides javadoc I can only assume this was to pass the test that I think is not required if we are less strict with the interface definition.

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractIndexProducerTest.java
```diff
@@ -134,8 +176,8 @@ public final void testConsistency() {
     @Test
     public final void testBehaviourAsIndexArray() {
         int flags = getBehaviour();
-        Assumptions.assumeTrue((flags & (AS_ARRAY_ORDERED | AS_ARRAY_DISTINCT)) != 0);
         int[] actual = createProducer().asIndexArray();
+        assumeTrue((flags & (AS_ARRAY_ORDERED | AS_ARRAY_DISTINCT)) != 0);
```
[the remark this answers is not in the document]
Claudenw: fixed

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractIndexProducerTest.java
```diff
@@ -149,7 +191,6 @@ public final void testBehaviourAsIndexArray() {
     @Test
     public final void testBehaviourForEach() {
         int flags = getBehaviour();
-        Assumptions.assumeTrue((flags & (FOR_EACH_ORDERED | FOR_EACH_DISTINCT)) != 0);
```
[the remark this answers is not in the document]
Claudenw: fixed

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBitCountProducerTest.java
```diff
@@ -60,23 +80,90 @@ public boolean test(int index, int count) {
...
+            return true;
+            }
+        );
+        assertEquals(expected, actual);
+    }
+
+    @Test
+    public final void testBehaviourForEachCount() {
+        int flags = getBehaviour();
+        IntList list = new IntList();
```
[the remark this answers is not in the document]
Claudenw: fixed

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBitCountProducerTest.java
```diff
@@ -60,23 +80,90 @@ public boolean test(int index, int count) {
...
+        int[] actual = list.toArray();
+        if ((flags & FOR_EACH_COUNT_ORDERED) != 0) {
+            int[] expected = Arrays.stream(actual).sorted().toArray();
+            Assertions.assertArrayEquals(expected, actual);
+        }
+        if ((flags & FOR_EACH_COUNT_DISTINCT) != 0) {
+            long count = Arrays.stream(actual).distinct().count();
+            Assertions.assertEquals(count, actual.length);
         }
+        int[] expected = getExpectedForEach();
```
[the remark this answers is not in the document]
Claudenw: At first I thought you were wrong, but now I see we are in violent agreement and my documentation is poorly written! 

Given the example at the bottom of the BitCountProducer interface documentation.

```
 * An IndexProducer that generates the values [1,2,3,1,5,6] can be represented by a BitCountProducer
 * that generates [(1,2),(2,1),(3,1),(5,1)(6,1)] or [(1,1),(2,1),(3,1),(1,1),(5,1),(6,1)] where the
 * entries may be in any order.
```

How would you rephrase:

```
 * A BitCountProducer may return duplicate indices and may be unordered.
 *
 * The guarantees are:
 * <ul>
 * <li>that for every unique value produced by the IndexProducer there will be at least one
 * index in the BitCountProducer.</li>
 * <li>that the total count of a specific value produced by the IndexProducer will equal the
 * total of the counts in the BitCountProducer for that index.</li>
 * </ul>
```

I have removed the `getExpectedForEach()` from the tests.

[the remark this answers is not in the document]
aherbert: I think this is what we have:

```
 * Defines a mapping of index to counts.
 *
 * <p>Note that a BitCountProducer may return duplicate indices and may be unordered.
 *
 * <p>Implementations must guarantee that:
 *
 * <ul>
 * <li>The mapping of index to counts is the combined sum of counts at each index.
 * <li>For every unique value produced by the IndexProducer there will be at least one matching
 * index and count produced by the BitCountProducer.
 * <li>The BitCountProducer will not generate indices that are not output by the IndexProducer.
 * </ul>
 *
 * <p>Note that implementations that do not output duplicate indices for BitCountProducer and
 * do for IndexProducer, or vice versa, are consistent if the distinct indices from each are
 * the same.
 *
 * <p>For example the mapping [(1,2),(2,3),(3,1)] can be output with many combinations including:
 * <pre>
 * [(1,2),(2,3),(3,1)]
 * [(1,1),(1,1),(2,1),(2,1),(2,1),(3,1)]
 * [(1,1),(3,1),(1,1),(2,1),(2,1),(2,1)]
 * [(3,1),(1,1),(2,2),(1,1),(2,1)]
 * ...
 * </pre>
```

I am not sure we need to spell it out with the example. In this case perhaps less is more.

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBitCountProducerTest.java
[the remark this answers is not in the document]
aherbert: As before, assumeTrue should be in the line after you get the flags.

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBitCountProducerTest.java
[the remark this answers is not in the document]
Claudenw: Replaced javadoc text with this code..

src/main/java/org/apache/commons/collections4/bloomfilter/BitCountProducer.java
```diff
@@ -47,7 +72,7 @@ default boolean forEachIndex(IntPredicate predicate) {
...
```
aherbert, line 75: I was considering why do we need to do this: convert an IndexProducer to a BitCountProducer. It would be to add it to a CountingBloomFilter. In this case the conversion does not remove duplicate indices. This means that the BitCountProducer can increment the count of an index more than 1 when adding a single IndexProducer.

This has the following consequences:
- The count at each index of a CountingBloomFilter does not represent the number of individual hashes added to the filter that had that index enabled. The count is bounded by the number of hashes as a lower limit, i.e. it may be higher.
- Removal of an IndexProducer from a counting bloom filter must use the same method to construct the BitCountProducer as the addition in order to use the exact same index->count mapping.

I am fine with this behaviour. Filtering an IndexProducer to unique indices is extra work. The `IndexFilter` class helps with this but requires a `Shape`. So it cannot be done in the BitCountProducer interface which has no Shape; other mechanisms to filter the IndexProducer without a bound on the index would be inefficient.

So perhaps we require a bit of extra documentation here to serve as a user-beware warning:
```
 * <p>Note that the BitCountProducer does not remove duplicates. Any use of the
 * BitCountProducer to create an aggregate mapping of index to counts, such as a
 * CountingBloomFilter, should use the same BitCountProducer in both add and
 * subtract operations to maintain consistency.
```
I believe this is consistent with the new wording in the javadoc header for the BitCountProducer interface.

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBitCountProducerTest.java
```diff
@@ -16,35 +16,46 @@
...
+
+import org.apache.commons.collections4.bag.TreeBag;
 import org.apache.commons.collections4.bloomfilter.BitCountProducer.BitCountConsumer;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
 public abstract class AbstractBitCountProducerTest extends AbstractIndexProducerTest {
 
+    /** Flag to indicate the {@link BitCountProducer#forEachCount(BitCountConsumer)} is ordered. */
+    protected static final int FOR_EACH_COUNT_ORDERED = 0x10;
```
aherbert, line 36: In the interest of simplicity, can these be dropped?

I previously stated that we should only have them if the behaviour of a BitCountProducer will be different from an IndexProducer. From what I can see in the tests these are used in 3 tests, and only when `FOR_EACH_ORDERED` and `FOR_EACH_DISTINCT` from the AbstractIndexProducerTest are also used. Dropping these constants will reduce maintenance and ensure the flags are correctly set for BitCount and Index producers.

You can add a note in the test to state that the same flags are reused.

I think the simplest way to do this is to update the flags to be private:
```Java
/** Flag to indicate the {@link BitCountProducer#forEachCount(BitCountConsumer)} is ordered.
 * This flag currently reuses the value from IndexProducer. At present no implementations
 * exhibit different behaviour for the two interfaces. This may change in the future and
 * an explicit flag value will be required. */
private static final int FOR_EACH_COUNT_ORDERED = FOR_EACH_ORDERED;
/** Flag to indicate the {@link BitCountProducer#forEachCount(BitCountConsumer)} is distinct.
 * This flag currently reuses the value from IndexProducer. At present no implementations
 * exhibit different behaviour for the two interfaces. This may change in the future and
 * an explicit flag value will be required. */
private static final int FOR_EACH_COUNT_DISTINCT = FOR_EACH_DISTINCT;
```
An alternative would be to have a flag for unspecified behaviour. This must be explicitly set by implementing tests:
```
FOR_EACH_COUNT_UNSPECIFIED = 0x10;
FOR_EACH_COUNT_ORDERED = 0x20;
FOR_EACH_COUNT_DISTINCT = 0x40;
```
The abstract test can then check if the behaviour is unspecified. If true then the other flags must be unset. If false then at least one of the other flags must be set. I think this is more work and not required, especially since we are testing the behaviour that is not a part of the contract given in the javadoc of the main implementations. This is more of a behaviour sense check of the implementations.

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBitCountProducerTest.java
```diff
@@ -60,23 +71,89 @@ public boolean test(int index, int count) {
...
+        assertEquals(expected, actual);
+    }
 
-        assertFalse(createProducer().forEachCount(FALSE_CONSUMER), "non-empty should be false");
-        assertTrue(createProducer().forEachCount(TRUE_CONSUMER), "non-empty should be true");
-        if (supportsEmpty()) {
-            assertTrue(createEmptyProducer().forEachCount(FALSE_CONSUMER), "empty should be true");
-            assertTrue(createEmptyProducer().forEachCount(TRUE_CONSUMER), "empty should be true");
+    @Test
+    public final void testBehaviourForEachCount() {
```
aherbert, line 128: Add a javadoc note:
```Java
/**
 * Test the behaviour of {@link BitCountProducer#forEachCount(BitCountConsumer)} with respect
 * to ordered and distinct indices. Currently the behaviour is assumed to be the same as
 * {@link IndexProducer#forEachIndex(java.util.function.IntPredicate)}.
 */
```

aherbert (commented):
Hi @Claudenw 

This is looking good. I suggested a bit more text for the BitCountProducer interface. There is also a simplification we can make in the AbstractBitCountProducerTest to lower maintenance by hiding the new flag constants (for now).

src/main/java/org/apache/commons/collections4/bloomfilter/BitCountProducer.java
Claudenw: added

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBitCountProducerTest.java
Claudenw: added

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBitCountProducerTest.java
Claudenw: I'm not sure I like this, but I can't think of a good counter argument.  I could do the hand waving and say IndexProducer and BitCountProducer produce different orders but I can't see a reasonable example in any possible implementation I can come up with.  So, changes made.

Claudenw:
I discovered issues with the HasherCollection and Hasher generated IndexProducer and BitCountProducers created from those IndexProducers.  Additional tests and corrections included in this latest push.

aherbert:
Last commit was on Oct 21st. Is this correct or do you need to push again?

Claudenw:
Let me see what happened.  In pushed this AM.  (or so I thought)

Claudenw:
> Last commit was on Oct 21st. Is this correct or do you need to push again?

Found the issue on my system.  Should be updated now.

src/main/java/org/apache/commons/collections4/bloomfilter/BitCountProducer.java
```diff
@@ -57,12 +88,22 @@ static BitCountProducer from(IndexProducer idx) {
             public boolean forEachCount(BitCountConsumer consumer) {
                 return idx.forEachIndex(i -> consumer.test(i, 1));
             }
+
+            @Override
+            public int[] asIndexArray() {
```
aherbert, line 93: I presume this change is required otherwise the test suite fails.

src/main/java/org/apache/commons/collections4/bloomfilter/HasherCollection.java
```diff
@@ -106,6 +120,23 @@ public boolean forEachIndex(IntPredicate consumer) {
...
+     * <p>This is equivalent to an IndexProducer created from a Bloom filter that comprises all
+     * the contained hashers.</p>
+     *
+     * @param shape the shape of the desired Bloom filter.
+     * @return the iterator of integers
+     */
+    public IndexProducer absoluteUniqueIndices(final Shape shape) {
+        return consumer -> {
+            Objects.requireNonNull(consumer, "consumer");
+            return uniqueIndices(shape).forEachIndex(IndexFilter.create(shape, consumer));
```
aherbert, line 136: IIUC this can exceed the capacity of the IndexFilter. The uniqueIndices(shape) will output up to k indices for each hasher in the collection (eliminating duplicates). Thus if we have two hashers with shape(k=5, n=1000) we could output 10 indices from the combined unique indices. Then we create an IndexFilter for shape which could be optimised only to check up to k=5 indices.

This throws an index out of bounds exception from IndexFilter.ArrayTracker.

```Java
@Test
public void testAbsoluteUniqueIndices() {
    int[] actual = new HasherCollection(
        new IncrementingHasher(1, 1),
        new IncrementingHasher(10, 1)
    ).absoluteUniqueIndices(Shape.fromKM(5, 1000)).asIndexArray();
    int[] expected = IntStream.concat(
            IntStream.range(1, 1 + 5),
            IntStream.range(10, 10 + 5)
        ).toArray();
    Assertions.assertArrayEquals(expected, actual);
}
```

When I update HasherCollection absoluteUniqueIndices to use a fake shape that forces the IndexTracker.BitMapTracker implementation then the above tests passes:
```Java
public IndexProducer absoluteUniqueIndices(final Shape shape) {
    return consumer -> {
        Objects.requireNonNull(consumer, "consumer");
        return uniqueIndices(shape).forEachIndex(IndexFilter.create(
                Shape.fromKM(shape.getNumberOfBits(), shape.getNumberOfBits()), consumer));
    };
}
```

I think that this can be easily solved by having a new package-private factory constructor in IndexFilter to choose the implementation that can handle exceeding the number of hash functions in the shape:

```Java
static IntPredicate create(int numberOfBits, IntPredicate consumer) {
    // Choose the BitMapTracker implementation to handle any number of hash functions
```

Or you could create a new Shape as:
```Java
public IndexProducer absoluteUniqueIndices(final Shape shape) {
    return consumer -> {
        Objects.requireNonNull(consumer, "consumer");
        return uniqueIndices(shape).forEachIndex(IndexFilter.create(
                Shape.fromKM(shape.getNumberOfHashFunctions() * hashers.size(),
                             shape.getNumberOfBits()), consumer));
    };
}
```
Perhaps the later is simpler.

src/main/java/org/apache/commons/collections4/bloomfilter/HasherCollection.java
```diff
@@ -141,29 +172,16 @@ public boolean forEachIndex(IntPredicate consumer) {
...
-                return lst.get(0);
-            }
-            int[] result = new int[count[0]];
-            int offset = 0;
-            for (int[] ary : lst) {
-                System.arraycopy(ary, 0, result, offset, ary.length);
-                offset += ary.length;
-            }
-            return result;
+            int[] result = new int[shape.getNumberOfHashFunctions()*hashers.size()];
```
aherbert, line 175: This is a much better implementation.

Can you add whitespace around the multiply ` * `.

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractIndexProducerTest.java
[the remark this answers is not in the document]
aherbert: You have not removed the redundant part of this test. The distinctness is tested in `testBehaviourAsIndexArray`.

src/test/java/org/apache/commons/collections4/bloomfilter/BitCountProducerFromAbsoluteUniqueHasherCollectionTest.java
```diff
@@ -0,0 +1,50 @@
...
+package org.apache.commons.collections4.bloomfilter;
+
+
+public class BitCountProducerFromAbsoluteUniqueHasherCollectionTest extends AbstractBitCountProducerTest {
+
+    @Override
+    protected BitCountProducer createProducer() {
+        // hasher has collisions and wraps
+        return BitCountProducer.from(new HasherCollection(
+                new IncrementingHasher(1, 1),
```
aherbert, line 26: This tests a saturation of the number of bits. We should really have a test for non saturation, no collisions just to make sure that works as well.

src/test/java/org/apache/commons/collections4/bloomfilter/IndexProducerFromAbsoluteUniqueHasherCollectionTest2.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.collections4.bloomfilter;
+
+public class IndexProducerFromAbsoluteUniqueHasherCollectionTest2 extends AbstractIndexProducerTest {
```
aherbert, line 19: Remove the 2 from the name suffix

src/test/java/org/apache/commons/collections4/bloomfilter/BitCountProducerFromAbsoluteUniqueHasherCollectionTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Override
+    protected BitCountProducer createEmptyProducer() {
+        return BitCountProducer.from(new HasherCollection().uniqueIndices(Shape.fromKM(11, 10)));
+    }
+
+    @Override
+    protected int getBehaviour() {
+        // Hasher allows duplicates and may be unordered
+        return 0;
```
aherbert, line 38: Should this have flags for distinct?

src/test/java/org/apache/commons/collections4/bloomfilter/BitCountProducerFromUniqueHasherCollectionTest.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.collections4.bloomfilter;
+
+
```
aherbert, line 19: Remove empty line

src/test/java/org/apache/commons/collections4/bloomfilter/BitCountProducerFromUniqueHasherTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @Override
+    protected BitCountProducer createEmptyProducer() {
+        return BitCountProducer.from(NullHasher.INSTANCE.indices(Shape.fromKM(17, 72)));
+    }
+
+    @Override
+    protected int getBehaviour() {
+        // Hasher allows duplicates and may be unordered
+        return AS_ARRAY_DISTINCT;
```
aherbert, line 35: Update the comment as this does not allow duplicates

Q. Is forEach also distinct?

src/test/java/org/apache/commons/collections4/bloomfilter/BitCountProducerFromUniqueHasherCollectionTest.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    protected BitCountProducer createEmptyProducer() {
+        return BitCountProducer.from(NullHasher.INSTANCE.uniqueIndices(Shape.fromKM(11, 10)));
+    }
+
+    @Override
+    protected int getBehaviour() {
+        // Hasher allows duplicates and may be unordered
```
aherbert, line 37: `// HasherCollection allows ...`

src/test/java/org/apache/commons/collections4/bloomfilter/EnhancedDoubleHasherTest.java
```diff
@@ -35,6 +36,12 @@ protected Hasher createEmptyHasher() {
         return NullHasher.INSTANCE;
     }
 
+
```
aherbert, line 39: Remove extra line

src/test/java/org/apache/commons/collections4/bloomfilter/IndexProducerFromAbsoluteUniqueHasherCollectionTest2.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Override
+    protected int[] getExpectedIndices() {
+        return new int[]{1, 2, 3, 4, 5, 6, 7, 8, 9, 0};
+    }
+
+    @Override
+    protected int getBehaviour() {
+        // HasherCollection allows duplicates and may be unordered
```
aherbert, line 41: Update comment as duplicates are not allowed from absoluteUniqueIndices.

Q. Should forEach also be distinct?

src/test/java/org/apache/commons/collections4/bloomfilter/UniqueIndexProducerFromHasherCollectionTest.java
```diff
@@ -36,4 +36,11 @@ protected int getBehaviour() {
         // index from each hasher. The result is there may still be duplicates.
         return 0;
     }
+
+
```
aherbert, line 40: Remove extra line

aherbert (changes requested):
This looks good. However there is a bug in the new HasherCollection.absoluteUniqueIndices method due to how the IndexFilter is created. I've suggested two quick fixes to get around this.

src/test/java/org/apache/commons/collections4/bloomfilter/UniqueIndexProducerFromHasherCollectionTest.java
Claudenw: Done

src/test/java/org/apache/commons/collections4/bloomfilter/IndexProducerFromAbsoluteUniqueHasherCollectionTest2.java
Claudenw: Updataed

src/test/java/org/apache/commons/collections4/bloomfilter/EnhancedDoubleHasherTest.java
Claudenw: fixed

src/test/java/org/apache/commons/collections4/bloomfilter/BitCountProducerFromUniqueHasherTest.java
Claudenw: yes, fixed

src/test/java/org/apache/commons/collections4/bloomfilter/BitCountProducerFromUniqueHasherCollectionTest.java
Claudenw: fixed

src/test/java/org/apache/commons/collections4/bloomfilter/BitCountProducerFromUniqueHasherCollectionTest.java
Claudenw: fixed

src/test/java/org/apache/commons/collections4/bloomfilter/BitCountProducerFromAbsoluteUniqueHasherCollectionTest.java
Claudenw: Yes

src/test/java/org/apache/commons/collections4/bloomfilter/IndexProducerFromAbsoluteUniqueHasherCollectionTest2.java
Claudenw: fixed

src/test/java/org/apache/commons/collections4/bloomfilter/BitCountProducerFromAbsoluteUniqueHasherCollectionTest.java
Claudenw: Changed

src/main/java/org/apache/commons/collections4/bloomfilter/HasherCollection.java
Claudenw: fixed

src/main/java/org/apache/commons/collections4/bloomfilter/HasherCollection.java
Claudenw: implemented later version with a check for hashers.size() == 0

src/main/java/org/apache/commons/collections4/bloomfilter/HasherCollection.java
```diff
@@ -90,6 +90,20 @@ public IndexProducer indices(final Shape shape) {
...
+    /**
+     * Creates an IndexProducer of comprising the unique indices from each of the contained
+     * hashers.
+     *
+     * <p>This method may return duplicates if the collection of unique values from each of the contained
+     * hashers contain duplicates.  This is equivalent to creating Bloom filters for each contained hasher
+     * and returning concatenated IndexProducer from each.</p>
+     *
+     * <p>A BitCountProducer generated from this IndexProducer is equivalent to a BitCountProducer from a
+     * counting Bloom filter that was constructed from the contained hashers.<p>
```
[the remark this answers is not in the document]
Claudenw: fixed

src/main/java/org/apache/commons/collections4/bloomfilter/HasherCollection.java
```diff
@@ -90,6 +90,20 @@ public IndexProducer indices(final Shape shape) {
         return new HasherCollectionIndexProducer(shape);
     }
 
+    /**
+     * Creates an IndexProducer of comprising the unique indices from each of the contained
+     * hashers.
+     *
+     * <p>This method may return duplicates if the collection of unique values from each of the contained
+     * hashers contain duplicates.  This is equivalent to creating Bloom filters for each contained hasher
+     * and returning concatenated IndexProducer from each.</p>
```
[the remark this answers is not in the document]
Claudenw: fixed

src/main/java/org/apache/commons/collections4/bloomfilter/HasherCollection.java
```diff
@@ -90,6 +90,20 @@ public IndexProducer indices(final Shape shape) {
         return new HasherCollectionIndexProducer(shape);
     }
 
+    /**
+     * Creates an IndexProducer of comprising the unique indices from each of the contained
```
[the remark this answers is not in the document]
Claudenw: done

src/main/java/org/apache/commons/collections4/bloomfilter/Hasher.java
```diff
@@ -49,7 +49,8 @@ public interface Hasher {
      * Creates an IndexProducer of unique indices for this hasher based on the Shape.
      *
      * <p>This is like the `indices(Shape)` method except that it adds the guarantee that no
-     * duplicate values will be returned</p>
+     * duplicate values will be returned.  The indices produced are equivalent those returned
```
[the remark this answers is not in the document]
Claudenw: fixed

Claudenw:
In all cases where `AS_ARRAY_ORDERED` is set `FOR_EACH_ORDERED` is also set, and the `FOR_EACH_COUNT_ORDERED` is therefore set as well.

In all cases where `AS_ARRAY_DISTINCT` is set `FOR_EACH_DISTINCT` is also set, and the `FOR_EACH_COUNT_DISTINCT` is therefore set as well.

I propose collapsing this into 2 statics:
`INDICES_ORDERED` : replaces `AS_ARRAY_ORDERED`, `FOR_EACH_ORDERED`, and `FOR_EACH_COUNT_ORDERED`.
`INDICES_DISTINCT`: replaces `AS_ARRAY_DISTINCT`, `FOR_EACH_DISTINCT` and `FOR_EACH_COUNT_DISTINCT`.

By "replaces" I mean remove all instances of the replaced constants (not make them equivalent).  
Add documentation that states the testing assumes the ORDERED and DISTINCT properties carry across `asArray()`, `forEachIndex()` and `forEachCount()`.

Also, since the `AbstractBitCountProducerTest ` extends `AbstractIndexProducerTest` we can remove all `IndexProducerFrom_X_Test` implementations where there is a `BitCountProducerFrom_X_Test`

@aherbert unless you have an objection, I'll proceed with these changes.

aherbert:
Hi @Claudenw, consolidating all the behaviour flags is a good idea. It will avoid future tests failing to set them correctly. I'll wait until you have done that for a review.

Note: I have reservations about the HasherCollection and the requirement for an `absoluteUniqueIndices` method. In summary:

1. I think it should be resolved after this PR
2. Since the HasherCollection violates the IndexProducer contract by outputting more than k indices, we may need to revise the contract and javadoc as being able to output up to a multiple of k indices for a Shape. This would be helped by a size-type method in the IndexProducer interface to state what count of k will be output.
3. The method `absoluteUniqueIndices` may be better served by a HasherSet, or some other naming variant, that ensures all indices output by the IndexProducer are unique.

Claudenw:
@aherbert The changes are in.  Have a look please

src/test/java/org/apache/commons/collections4/bloomfilter/DefaultIndexProducerTest.java
```diff
@@ -47,8 +52,7 @@ public boolean forEachIndex(IntPredicate predicate) {
 
     @Override
     protected int getBehaviour() {
-        // The default method streams a BitSet so is distinct and ordered.
```
aherbert, line 50: Note: This test has removed some testing of the default methods in IndexProducer.

It should be named `IndexProducerFromIndexArrayTest`. It is specifically targeting the default implementation of the IndexProducer when created from an index array. In that case the `asIndexArray()` method returns a clone of the original indices used to create it.

A `DefaultIndexProducerTest` would test the default implementation of `asIndexArray()` and look like this.

```Java
public class DefaultIndexProducerTest extends AbstractIndexProducerTest {

    /** Make forEachIndex unordered and contain duplicates. */
    private int[] values = {10, 1, 10, 1};

    @Override
    protected int[] getExpectedIndices() {
        return values;
    }

    @Override
    protected IndexProducer createProducer() {
        return new IndexProducer() {
            @Override
            public boolean forEachIndex(IntPredicate predicate) {
                Objects.requireNonNull(predicate);
                for (int i : values) {
                    if (!predicate.test(i)) {
                        return false;
                    }
                }
                return true;
            }
        };
    }

    @Override
    protected IndexProducer createEmptyProducer() {
        return new IndexProducer() {
            @Override
            public boolean forEachIndex(IntPredicate predicate) {
                Objects.requireNonNull(predicate);
                return true;
            }
        };
    }

    @Override
    protected int getBehaviour() {
        // The default method streams a BitSet so is distinct and ordered.
        // However the forEachIndex may not be distinct and ordered and
        // the test cannot distinguish the two cases.
        return 0;
    }
}
```

Note that this highlights a problem with consolidating all the behaviour flags. Currently the test suite cannot be used to test the asIndexArray behaviour is different from the forEachIndex behaviour.

I think we should revert the consolidation of the flags. However to avoid the duplication of flag setting through the various test cases we can change the AbstractIndexProducerTest test to do this:
```Java
public abstract class AbstractIndexProducerTest {

    protected static final int ORDERED = 0x1;
    protected static final int DISTINCT = 0x2;

    protected abstract int getAsIndexArrayBehaviour();

    protected int getForEachIndexBehaviour() {
        return getAsIndexArrayBehaviour();
    }

}
```
And AbtractBitCountProducerTest:
```Java
public abstract class AbstractBitCountProducerTest extends AbstractIndexProducerTest {

    protected int getForEachCountBehaviour() {
        return getAsIndexArrayBehaviour();
    }

}
```

Then get the correct flag in the relevant test. It provides the ability to override each test behaviour individually but will default to the same behaviour if not explicitly changed.

Doing this change (DefaultIndexProducerTest and IndexProducerFromIndexArrayTest) should remove the requirement for the additional test `testFromIndexArray`. The method `testFromBitMapProducer` is already covered in IndexProducerFromBitMapProducerTest.

src/test/java/org/apache/commons/collections4/bloomfilter/IndexProducerFromBitmapProducerTest.java
```diff
@@ -49,10 +49,15 @@ protected IndexProducer createProducer() {
...
+    @Override
+    protected int[] getExpectedIndices() {
+        return new int[]{0, 65, 128, 129};
+    }
+
     @Override
     protected int getBehaviour() {
         // Bit maps will be distinct. Conversion to indices should be ordered.
-        return FOR_EACH_DISTINCT | FOR_EACH_ORDERED | AS_ARRAY_DISTINCT | AS_ARRAY_ORDERED;
+        return INDICES_DISTINCT | INDICES_ORDERED | INDICES_DISTINCT | INDICES_ORDERED;
```
aherbert, line 60: Duplicates

aherbert (changes requested):
Hi @Claudenw 

I think the consolidation of the ordered and distinct flags has reduced the flexibility of the test suite. It also led to you duplicating the flags in all test classes that override getBehaviour().

However the previous flexibility can be reinstated with my suggestion to have a method to get the behaviour for each of: asIndexArray; forEachIndex; forEachCount. By default this would return the same behaviour for all of them. But it can be configured to change and allow testing different behaviour for the methods such as that done in the default implementation of IndexProducer.

src/test/java/org/apache/commons/collections4/bloomfilter/DefaultIndexProducerTest.java
Claudenw: Corrected.  I also added `DefaultBitCountProducerTest` and removed duplicate test `UniqueIndexProducerFromHashCollectionTest` which was effectively a duplicate of `BitCountProducerFromUniqueHasherCollectionTest`

src/test/java/org/apache/commons/collections4/bloomfilter/DefaultBitCountProducerTest.java
```diff
@@ -0,0 +1,94 @@
...
+                    if (!predicate.test(i)) {
+                        return false;
+                    }
+                }
+                return true;
+            }
+
+            @Override
+            public boolean forEachCount(BitCountConsumer consumer) {
+                int[] vals = values.clone();
```
aherbert, line 50: There are no comments as to why you are doing this. Was it for testing?

I suggest leaving forEachCount to not be ordered and match the forEachIndex order. This should test the default implementations of BitCountProducer, not add in extra tests for convenience.

src/test/java/org/apache/commons/collections4/bloomfilter/DefaultBitCountProducerTest.java
```diff
@@ -0,0 +1,94 @@
...
+            public boolean forEachCount(BitCountConsumer consumer) {
+                return true;
+            }
+        };
+    }
+
+    @Override
+    protected int getAsIndexArrayBehaviour() {
+        // The default method streams a BitSet so is distinct and ordered.
+        // However the forEachIndex may not be distinct and ordered and
```
aherbert, line 80: This comment is outdated. The comment should be moved to `getForEachIndexBehaviour` as:
```Java
// The forEachIndex output is intentionally unordered and contains duplicates
```

If you wish to test forEachCount can use different flags then add some comments:
```Java
// Force to be ordered
int[] vals = values.clone();
Arrays.sort(vals);

// ...

// The forEachCount output is intentionally ordered and contains duplicates
```

src/test/java/org/apache/commons/collections4/bloomfilter/DefaultIndexProducerTest.java
```diff
@@ -36,22 +38,41 @@ protected int[] getExpectedIndices() {
...
                 return true;
             }
         };
     }
 
     @Override
-    protected int getBehaviour() {
+    protected int getAsIndexArrayBehaviour() {
+        // The default method streams a BitSet so is distinct and ordered.
+        // However the forEachIndex may not be distinct and ordered and
```
aherbert, line 69: Again can you move this comment to `getForEachIndexBehaviour` as:
```Java
// The forEachIndex output is intentionally unordered and contains duplicates
```

aherbert (changes requested):
Hi @Claudenw 

Thanks for the changes. They look good. I requested a few comment changes to indicate what the tests are intentionally trying to test.

src/test/java/org/apache/commons/collections4/bloomfilter/DefaultBitCountProducerTest.java
Claudenw: I should have reverted that.  It was to test that the forEachCount tests used the proper behaviour.

PR #336, opened 2022-09-13
Fix flaky test failure in SynchronizedBagTest#testCollectionToArray2

Partha-SUST16:
### What is the purpose of this PR

- This PR fixes the flaky test  `org.apache.commons.collections4.bag.SynchronizedBagTest.testCollectionToArray2`
- The mentioned test may fail or pass without changes made to the source code when it is run in different JVMs due to `HashMap`'s non-deterministic iteration order.

### Why the test fails
- As the `SynchronizedBagTest` uses `HashBag` as its `collection` which uses `HashMap` internally and as per `Java 8` documentation, the ordering of `HashMap` is not constant, so assigning the same hashmap as an array to different objects can change the ordering. So when comparing these two arrays with `assertEquals` the test may fail as their ordering can be different.

### Reproduce the test failure
- Run the test with 'NonDex' maven plugin. The command to recreate the flaky test failure is 
  ` mvn -pl . edu.illinois:nondex-maven-plugin:1.1.2:nondex -debug -Dtest=org.apache.commons.collections4.bag.SynchronizedBagTest#testCollectionToArray2`

- Fixing the flaky test now may prevent flaky test failures in future Java versions.

### Expected result
- The tests should run successfully when run with 'NonDex' maven with the same command for recreating the flaky test.

### Actual Result
- We get the failure:
    `[ERROR] testCollectionToArray2  Time elapsed: 0.002 s  <<< FAILURE!
junit.framework.AssertionFailedError: toArrays should be equal expected:<[10, 11, Thirteen, null, Nine, 16, 14, Eight, , Three, 15, 12, 6.0, Seven, 4, One, 2, 5.0]> but was:<[4, 12, , 10, 16, One, Thirteen, Nine, null, 2, 5.0, 6.0, 15, Seven, 14, Eight, Three, 11]>
`
### Fix
- Create a `HashSet` from both arrays named `a` and `array` and check if they both are equal in `testCollectionToArray2`

### DIFF
--- a/src/test/java/org/apache/commons/collections4/bag/SynchronizedBagTest.java
+++ b/src/test/java/org/apache/commons/collections4/bag/SynchronizedBagTest.java
@@ -46,6 +46,10 @@ public class SynchronizedBagTest<T> extends AbstractBagTest<T> {
         return "4";
     }
 
+    @Override
+    protected int getIterationBehaviour(){
+        return UNORDERED;
+    }
 //    public void testCreate() throws Exception {
 //        Bag<T> bag = makeObject();
 //        writeExternalFormToDisk((java.io.Serializable) bag, "src/test/resources/data/test/SynchronizedBag.emptyCollection.version4.obj");
--- a/src/test/java/org/apache/commons/collections4/bag/SynchronizedBagTest.java
+++ b/src/test/java/org/apache/commons/collections4/bag/SynchronizedBagTest.java
@@ -50,6 +50,7 @@ public class SynchronizedBagTest<T> extends AbstractBagTest<T> {
     protected int getIterationBehaviour(){
         return UNORDERED;
     }
+
 //    public void testCreate() throws Exception {
 //        Bag<T> bag = makeObject();
 //        writeExternalFormToDisk((java.io.Serializable) bag, "src/test/resources/data/test/SynchronizedBag.emptyCollection.version4.obj");
--- a/src/test/java/org/apache/commons/collections4/collection/AbstractCollectionTest.java
+++ b/src/test/java/org/apache/commons/collections4/collection/AbstractCollectionTest.java
@@ -1096,7 +1096,7 @@ public abstract class AbstractCollectionTest<E> extends AbstractObjectTest {
         array = getCollection().toArray(new Object[0]);
         a = getCollection().toArray();
         assertEquals("toArrays should be equal",
-                     Arrays.asList(array), Arrays.asList(a));
+                     new HashSet<>(Arrays.asList(array)), new HashSet<>(Arrays.asList(a)));
 
         // Figure out if they're all the same class
         // TODO: It'd be nicer to detect a common superclass
--- a/src/test/java/org/apache/commons/collections4/collection/AbstractCollectionTest.java
+++ b/src/test/java/org/apache/commons/collections4/collection/AbstractCollectionTest.java
@@ -71,10 +71,10 @@ import org.junit.jupiter.api.Test;
  * </ul>
  * <p>
  * <b>Indicate Collection Iteration behaviour</b>
- * </p>
+ * <p>
  * Override these if your collection makes no ordering guarantees for the iterator
  * <ul>
- *  <li>{@link #getIterationBehaviour()}</li>
+ * <li>{@link #getIterationBehaviour()}</li>
  * </ul>
  * <p>
  * <b>Fixture Methods</b>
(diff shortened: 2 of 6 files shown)


aherbert:
This fix may result in hiding implementation errors.

This class is used to test many collections including lists. The toArray contract states:
```
"If this collection makes any guarantees as to what order its elements are
returned by its iterator, this method must return the elements in the
same order."
```
So by ignoring the order returned by toArray the test will not detect errors in implementations that should respect the iterator order.

Some of the collections may support duplicates, e.g. lists. By using a HashSet to collate the arrays before comparison you ignore the order and may also ignore duplicates (depending on hashCode/equals implementation of stored objects).

Note that if all items in the list are the same object class then line 1119 in the same test method performs another assertEquals using two lists built from the output of toArray. It is interesting to note that your fix and verification using NonDex does not detect this as flaky; the conclusion being that many collections tested by this method store objects of more than one class in the reference collection (and thus do not hit this line).

Ideally the AbstractCollectionTest requires a behaviour flag for this test to indicate if the collection under test has a toArray method that respects the iterator order. Then assertEquals can be used with a list, otherwise a loop over the output should compare each item in 1 array is also present somewhere in the other array, see AbstractCollectionTest.verify.

Partha-SUST16:
Thanks @aherbert  for your response. I agree that my fix before indeed may hide implementation errors. I have just tried another fix. Namely, changing the use of HashMap to a LinkedHashMap inside HashBag, which addresses the non-determinism in the iteration order and should not introduce implementation errors. If I am mistaken or if you prefer a different fix, please do let me know

aherbert:
Changing the implementation to pass a test is the wrong solution. This compromises the implementation. You should change the test. Either:

- The toArray result is expected to match the iterator order. Add both to a list and use assertEquals.
- The toArray result is expected to output the same items as the iterator. Add both to a set and use assertEquals.

The test should have a behaviour flag to be overridden by implementing test cases:

```Java
/**
 * Flag to indicate the collection makes no ordering guarantees for the iterator. If this is not used
 * then the behaviour is assumed to be ordered and the output order of the iterator is matched by
 * the toArray method.
 */
protected static final int UNORDERED = 0x1;

/**
 * Return a flag specifying the behaviour of the collection.
 * This is used to change the assertions used by specific tests.
 *
 * @return the behaviour
 */
protected abstract int getBehaviour();
```
Update the test to handle both cases:
```Java
if ((getBehaviour() & UNORDERED) != 0) {
    // Not expected to be ordered
} else {
    // Order must match
}
```

Partha-SUST16:
Thanks @aherbert for pointing in the right direction. 
To fix the flaky test I have now introduced `getIterationBehaviour()` method and in our test file that contains the flaky test I have overridden this method and returns `UNORDERED` flag.
I did not make `getIterationBehaviour()` abstract as this will result in implementing this method in all subclass of `AbstractCollectionTest`.  So I made this a normal method with returning `ORDERED` as default implementation.
This fix passes all the normal tests along with the flaky test that was identified.
Please let me know your thinking on this.

src/test/java/org/apache/commons/collections4/bag/SynchronizedBagTest.java
```diff
@@ -46,6 +46,10 @@ public String getCompatibilityVersion() {
         return "4";
     }
 
+    @Override
+    protected int getIterationBehaviour(){
+        return UNORDERED;
+    }
```
aherbert, line 52: Add an empty line after this method.

src/test/java/org/apache/commons/collections4/collection/AbstractCollectionTest.java
```diff
@@ -70,6 +70,13 @@
  * <li>{@link #isFailFastSupported()}
  * </ul>
  * <p>
+ * <b>Indicate Collection Iteration behaviour</b>
+ * </p>
+ * Override these if your collection makes no ordering guarantees for the iterator
```
aherbert, line 75: Requires a `<p>` tag before `Override`. Looking at the rest of the javadoc the paragraphs are not closed. So the `</p>` on line 74 should change to a `<p>`.

aherbert (changes requested):
This looks good. Just a few minor changes required.

Partha-SUST16:
Hello @aherbert, I have updated the pull request with the file change you mentioned.

aherbert:
This looks fine. However you have a checkstyle issue to correct. Please see the CI build failure for JDK 11. You can run the build locally to test using the default maven goal: `mvn`

Partha-SUST16:
Sorry for not checking this earlier. Updated the code with spacing fix and now all checks have passed @aherbert

aherbert:
@Partha-SUST16 I updated the test for unordered arrays in commit c35d8c6fd1ae28680c9912acf031173db389b80c.

Matching the length and then matching items in a Set will not detect a count mismatch of duplicates. This test now explicitly matches each item once and only once. Please raise another PR if you find any more issues. Thanks for the contribution.

PR #340, opened 2022-09-24
COLLECTIONS-814 CollectionUtils.removeAll() not throwing NPE

angusdev:
[COLLECTIONS-814](https://issues.apache.org/jira/browse/COLLECTIONS-814)
CollectionUtils.removeAll() not throwing proper NullPointerException(NPE) if the first parameter is empty

Add null check in ListUtil.removeAll()


angusdev:
ping?

kinow (approved):
Thanks for bumping this PR @angusdev . I had a read of the JIRA issue, and confirmed the Javadocs state either parameter being null should result in a NPE, but that was not the case. Thanks for including a unit test. I think this can be merged now. I will squash the commits and prepare the change log so this can be merged.

kinow:
Merged. Thanks @angusdev !

PR #341, opened 2022-09-25
Fix flaky test occurred in 'CollectionBagTest.testCollectionToArray2'

Partha-SUST16:
### What is the purpose of this PR

- This PR fixes the flaky test  `org.apache.commons.collections4.bag.CollectionBagTest.testCollectionToArray2`
- The mentioned test may fail or pass without changes made to the source code when it is run in different JVMs due to `HashMap`'s non-deterministic iteration order.

### Why the test fails
- As the `CollectionBagTest` uses `HashBag` as its `collection` which uses `HashMap` internally and as per `Java 8` documentation, the ordering of `HashMap` is not constant, so assigning the same hashmap as an array to different objects can change the ordering. So when comparing these two arrays with `assertEquals` the test may fail as their ordering can be different.

### Reproduce the test failure
- Run the test with 'NonDex' maven plugin. The command to recreate the flaky test failure is 
  ` mvn -pl . edu.illinois:nondex-maven-plugin:1.1.2:nondex -debug -Dtest=org.apache.commons.collections4.bag.CollectionBagTest#testCollectionToArray2`

- Fixing the flaky test now may prevent flaky test failures in future Java versions.

### Expected result
- The tests should run successfully when run with 'NonDex' maven with the same command for recreating the flaky test.

### Actual Result
- We get the failure:
    `[ERROR] org.apache.commons.collections4.bag.CollectionBagTest.testCollectionToArray2  Time elapsed: 0.025 s  <<< FAILURE! junit.framework.AssertionFailedError: toArrays should be equal expected:<[10, 6.0, 11, 4, Seven, Eight, 5.0, 15, 2, 12, 16, Thirteen, null, , Three, Nine, One, One, 14]> but was:<[, null, Seven, 2, 5.0, 12, 10, 11, 14, 16, One, One, Nine, Eight, Thirteen, 15, 4, 6.0, Three]>
`
### Fix
- Override the `getIterationBehaviour()' method from `AbstractCollectionTest` and return `UNORDERED` in `CollectionBagTest`.

PR #353, opened 2022-10-28
[COLLECTIONS-836] fix flaky test DualHashBidiMapTest

anantdahiya8:
In test : `org.apache.commons.collections4.bidimap.DualHashBidiMapTest#testBidiKeySetValuesOrder`
The order of keyset and valueSet can differ because it uses `HashMap` which is non deterministic so its order cannot be determined.

Replacing `HashMap` with `LinkedHashMap` so the order of keySet and valueSet remains same.

### DIFF
--- a/src/test/java/org/apache/commons/collections4/bidimap/AbstractBidiMapTest.java
+++ b/src/test/java/org/apache/commons/collections4/bidimap/AbstractBidiMapTest.java
@@ -27,8 +27,10 @@ import java.util.Set;
 import org.apache.commons.collections4.BidiMap;
 import org.apache.commons.collections4.BulkTest;
 import org.apache.commons.collections4.MapIterator;
+import org.apache.commons.collections4.collection.AbstractCollectionTest;
 import org.apache.commons.collections4.iterators.AbstractMapIteratorTest;
 import org.apache.commons.collections4.map.AbstractIterableMapTest;
+import org.junit.jupiter.api.Assumptions;
 import org.junit.jupiter.api.Test;
 
 /**
@@ -265,6 +267,8 @@ public abstract class AbstractBidiMapTest<K, V> extends AbstractIterableMapTest<
 
     @Test
     public void testBidiKeySetValuesOrder() {
+        // Skip if collection is unordered
+        Assumptions.assumeFalse((getIterationBehaviour() & AbstractCollectionTest.UNORDERED) != 0);
         resetFull();
         final Iterator<K> keys = map.keySet().iterator();
         final Iterator<V> values = map.values().iterator();


src/main/java/org/apache/commons/collections4/bidimap/DualHashBidiMap.java
```diff
@@ -22,6 +22,7 @@
 import java.io.Serializable;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.LinkedMap;
```
kinow, line 25: `s/LinkedMap/LinkedHashMap`

Make sure you read `CONTRIBUTING.md` too as it contains useful information for contributing code.

kinow (changes requested):
CI build is broken due to wrong import.

src/main/java/org/apache/commons/collections4/bidimap/DualHashBidiMap.java
anantdahiya8: Yes, thanks

anantdahiya8:
> CI build is broken due to wrong import.

Updated

aherbert:
Note that there already is a `DualLinkedHashBidiMap` object. So this change is converting the `DualHashBidiMap` to a `DualLinkedHashBidiMap`. This is not the solution.

-1

As with previous PRs related to a similar issue, we should change the test rather than the implementation. See #341, #336.

If the iteration order of the DualHashBidiMap is not specified, then do not change it to pass an arbitrary test. To determine how flaky this test is you could try using the `nondex` plugin. See the other linked PRs for examples.

A simple fix can be to change `testBidiKeySetValuesOrder` to check the `getIterationBehaviour()` flag and skip this test:

```Java
@Test
public void testBidiKeySetValuesOrder() {
    Assume.assumeTrue((getIterationBehaviour() & AbstractCollectionTest.UNORDERED) == 0);
    ...
```

anantdahiya8:
> 

Yup, this is what I was thinking. Thanks for the reference, will make the change

anantdahiya8:
@aherbert Updated

src/test/java/org/apache/commons/collections4/bidimap/AbstractBidiMapTest.java
```diff
@@ -265,16 +266,18 @@ private void removeValue(final BidiMap<?, ?> map, final Object value) {
...
-        final Iterator<K> keys = map.keySet().iterator();
-        final Iterator<V> values = map.values().iterator();
-        while (keys.hasNext() && values.hasNext()) {
-            final K key = keys.next();
-            final V value = values.next();
-            assertSame(map.get(key), value);
-        }
-        assertFalse(keys.hasNext());
-        assertFalse(values.hasNext());
+        if((getIterationBehaviour() & AbstractCollectionTest.UNORDERED) == 0){
```
aherbert, line 269: Using an if statement does not allow reporting that the test has been skipped. Can you revert and use the previous suggestion to skip based on an assumption (Junit5):
```Java
public void testBidiKeySetValuesOrder() {
    // Skip if collection is unordered
    Assumptions.assumeFalse((getIterationBehaviour() & AbstractCollectionTest.UNORDERED) != 0);
```

anantdahiya8: @aherbert Got it. Updated

aherbert:
This is failing CI due to a configuration error in the POM that somehow brings in the wrong junit runner for this junit5 test and the TestAbortedException causes an error and not an ignored test. Since we have no junit4 test fixtures I have updated master to remove junit-jupiter-vintage.

If you rebase and push this should work.

anantdahiya8:
@aherbert Thanks. Updated

kinow (approved):

PR #357, opened 2022-11-06
COLLECTIONS-830: Added Hasher.isEmpty() and appropriate tests

Claudenw:
Adds isEmpty() method to Hasher and implements appropriate tests.

Fixes COLLECTIONS-830

### DIFF
--- a/src/main/java/org/apache/commons/collections4/bloomfilter/Hasher.java
+++ b/src/main/java/org/apache/commons/collections4/bloomfilter/Hasher.java
@@ -65,7 +65,7 @@ public interface Hasher {
     /**
      * Returns {@code true} if the hasher is empty and will return no values.
      *
-     * By default this method returns {@code false}.
+     * <p>By default this method returns {@code false}.</p>
      *
      * @return {@code true} if the hasher is empty and will return no values.
      */
--- a/src/main/java/org/apache/commons/collections4/bloomfilter/Hasher.java
+++ b/src/main/java/org/apache/commons/collections4/bloomfilter/Hasher.java
@@ -64,7 +64,12 @@ public interface Hasher {
 
     /**
      * Returns {@code true} if the hasher is empty and will return no values.
+     *
+     * By default this method returns {@code false}.
+     *
      * @return {@code true} if the hasher is empty and will return no values.
      */
-    boolean isEmpty();
+    default boolean isEmpty() {
+        return false;
+    }
 }
--- a/src/main/java/org/apache/commons/collections4/bloomfilter/Hasher.java
+++ b/src/main/java/org/apache/commons/collections4/bloomfilter/Hasher.java
@@ -61,4 +61,10 @@ public interface Hasher {
             return Hasher.this.indices(shape).forEachIndex(IndexFilter.create(shape, consumer));
         };
     }
+
+    /**
+     * Returns {@code true} if the hasher is empty and will return no values.
+     * @return {@code true} if the hasher is empty and will return no values.
+     */
+    boolean isEmpty();
 }
--- a/src/main/java/org/apache/commons/collections4/bloomfilter/HasherCollection.java
+++ b/src/main/java/org/apache/commons/collections4/bloomfilter/HasherCollection.java
@@ -66,6 +66,16 @@ public class HasherCollection implements Hasher {
         this(Arrays.asList(hashers));
     }
 
+    @Override
+    public boolean isEmpty() {
+        for (Hasher h : hashers) {
+            if (!h.isEmpty()) {
+                return false;
+            }
+        }
+        return true;
+    }
+
     /**
      * Adds a hasher to the collection.
      * @param hasher The hasher to add.
--- a/src/test/java/org/apache/commons/collections4/bloomfilter/AbstractHasherTest.java
+++ b/src/test/java/org/apache/commons/collections4/bloomfilter/AbstractHasherTest.java
@@ -19,6 +19,7 @@ package org.apache.commons.collections4.bloomfilter;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.CsvSource;
 
@@ -85,4 +86,7 @@ public abstract class AbstractHasherTest extends AbstractIndexProducerTest {
         });
         assertEquals(1, count[0], "did not exit early");
     }
+
+    @Test
+    public abstract void testIsEmpty();
 }
--- a/src/test/java/org/apache/commons/collections4/bloomfilter/AbstractHasherTest.java
+++ b/src/test/java/org/apache/commons/collections4/bloomfilter/AbstractHasherTest.java
@@ -17,6 +17,7 @@
 package org.apache.commons.collections4.bloomfilter;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import org.junit.jupiter.api.Test;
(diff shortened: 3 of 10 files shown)


src/main/java/org/apache/commons/collections4/bloomfilter/HasherCollection.java
```diff
@@ -66,6 +66,16 @@ public HasherCollection(Hasher... hashers) {
...
 
+    @Override
+    public boolean isEmpty() {
+        for (Hasher h : hashers) {
+            if (!h.isEmpty()) {
+                return false;
+            }
+        }
+        return true;
+    }
```
kinow, line 77: Bit rusty with Java & streams, but I **think** this is equivalent to:

```java
return hashers.stream().noneMatch(Hasher::isEmpty);
```

src/main/java/org/apache/commons/collections4/bloomfilter/HasherCollection.java
```diff
@@ -66,6 +66,16 @@ public HasherCollection(Hasher... hashers) {
...
     }
 
+    @Override
+    public boolean isEmpty() {
+        for (Hasher h : hashers) {
+            if (!h.isEmpty()) {
+                return false;
+            }
+        }
+        return true;
```
kinow, line 76: Bit rusty with Java & streams, but I **think** this is equivalent to:

```java
return hashers.stream().noneMatch(Hasher::isEmpty);
```

kinow: Out of the loop, so can't comment on the addition of that method, but the rest of the code looks Ok :+1:

src/test/java/org/apache/commons/collections4/bloomfilter/EnhancedDoubleHasherTest.java
```diff
@@ -52,6 +53,11 @@ protected int getHasherSize(Hasher hasher) {
         return 1;
     }
 
+    @Override
+    public void testIsEmpty() {
+        assertFalse( createHasher().isEmpty() );
```
garydgregory, line 58: No extra spaces around parens.

garydgregory:
Hm, `HasherCollection` is not a `Collection`, instead it is a `List` wrapper, so I suggest it be renamed `HasherList` and that it implements `Iterable<Hasher>`. WDYT? It seems to heavy to implement `List` itself.

aherbert (changes requested):
I would rather have `isEmpty` be default and the Hasher interface can remain a functional interface.

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractHasherTest.java
```diff
@@ -85,4 +86,7 @@ public void testHashing(int k, int m) {
         });
         assertEquals(1, count[0], "did not exit early");
     }
+
+    @Test
+    public abstract void testIsEmpty();
```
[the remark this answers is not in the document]
Claudenw: changed

src/main/java/org/apache/commons/collections4/bloomfilter/Hasher.java
```diff
@@ -61,4 +61,10 @@ default IndexProducer uniqueIndices(Shape shape) {
             return Hasher.this.indices(shape).forEachIndex(IndexFilter.create(shape, consumer));
         };
     }
+
+    /**
+     * Returns {@code true} if the hasher is empty and will return no values.
+     * @return {@code true} if the hasher is empty and will return no values.
+     */
+    boolean isEmpty();
```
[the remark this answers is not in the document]
Claudenw: changed

src/test/java/org/apache/commons/collections4/bloomfilter/EnhancedDoubleHasherTest.java
Claudenw: fixed

src/main/java/org/apache/commons/collections4/bloomfilter/HasherCollection.java
Claudenw: Leaving as non-stream method to avoid the stream object creation overhead.

src/main/java/org/apache/commons/collections4/bloomfilter/Hasher.java
```diff
@@ -61,4 +61,15 @@ default IndexProducer uniqueIndices(Shape shape) {
             return Hasher.this.indices(shape).forEachIndex(IndexFilter.create(shape, consumer));
         };
     }
+
+    /**
+     * Returns {@code true} if the hasher is empty and will return no values.
+     *
+     * By default this method returns {@code false}.
```
aherbert, line 68: Needs a `<p>` tag. Change to `<p>The default implementation returns ... </p>`

aherbert (changes requested):
Minor formatting change but otherwise looks OK.

I am curious as to why you would have an empty hasher. Is this not a programming error? Basically it means you are testing if `nothing` is present in a filter, or adding `nothing` to a filter.

Claudenw:
> I am curious as to why you would have an empty hasher. Is this not a programming error? Basically it means you are testing if `nothing` is present in a filter, or adding `nothing` to a filter.

First I need to define a "reference" Bloom filter.  Consider the properties of some object.  If you can index and search a collection of Bloom filters then you can take the properties of an object, create a Bloom filter from the values, and use that as a reference to the object.  Now you can search the collection of filters by creating a Bloom filter that contains only those values that you want to locate.  This usage of Bloom filters is really only available once the indexed collection is possible.  The indexed collection is called a multidimensional Bloom filter -- actually any collection of Bloom filters is a multidimensional Bloom filter.

In the case of a searching multidimensional Bloom filter in a client-server environment where:

- the client and the server agree on the underlying hashing algorithm and the use of the EnhancedDoubleHasher.  
- the Bloom filter being generated is not a standard "gateway" Bloom filter but a "reference" filter.
- The client creates the hashers for the filter and places them into a collection which it then passes to the server.
- The server deserializes the hashers into a HasherCollection.
- If the hasher is empty the server can return the entire collection or can return an error.
- otherwise the server performs the search building the internal Bloom filters as required by the multidimensional Bloom filter algorithm.

Notice that the empty hasher allows a short circuit of the expensive search.

aherbert:
IIUC the use of `isEmpty` is to prevent a client sending an empty something to a server which would waste its time, or use it as a signal. The former seems like an error by the client (or a denial of service attack); the later can be achieved by other means, for example sending empty/signalling bytes before deserialisation on the server side. This would avoid having to create the (empty) HasherCollection altogether.

It seems like quite an edge case scenario. If you have to use an empty Hasher as a signal then it can be done on the server side using:
```Java
static boolean isEmpty(Hasher h) {
    boolean[] flag = {true};
    // Shape can be precomputed for this purpose
    h.indices(Shape.fromKM(1, 1)).forEachIndex(i -> flag[0] = false);
    return flag[0];
}
```
If you have to deserialise the input bytes to a Hasher then just do it as a customisation of your choice of Hasher with an `isEmpty` method. If you are using an EnhancedDoubleHasher then note that this cannot be "empty" even if the initial and increment are zero given that Shape numberOfHash functions is always `> 0` so at least 1 index will be output.

In general any Hasher should output indices. If you created a hasher that contains nothing, this itself is something that should generate output indices. The Hasher simply outputs indices based on a source, and ideally the output sequence is fairly unique among possible input sources.

Claudenw:
Perhaps the HasherCollection does not belong in the library.  It may make more sense to withdraw this pull request, close the COLLECTIONS-830 as won't fix and create another pull request to remove the HasherCollection.  I know there were other issues that @aherbert  had with that implementation anyway.

aherbert:
Removal of HasherCollection would eliminate all the possible confusion around its use given that it has behaviour that can be, at least in part, either: no items; a single item (Hasher); collection of items (List<Hasher>); or a bitwise combination of items (BloomFilter).

I suggest the functionality can be matured outside of the core functionality in Collections and then integrated with all the appropriate code for the use case(s).

Claudenw:
HassherCollection to be removed from code base

PR #359, opened 2022-11-07
add function CollectionUtils.sizeIsNotEmpty

XenoAmess:
as it is really useful sometimes.


aherbert:
Adding:
```Java
CollectionUtils.sizeIsNotEmpty(object)
```
to avoid the existing:
```Java
!CollectionUtils.sizeIsEmpty(object)
```
seems like code bloat to me.

garydgregory:
I agree with @aherbert 
-1

PR #360, opened 2022-11-08
Fix flaky test in CollectionUtilsTest.Java

yannizhou05:
The test below in `CollectionUtilsTest.java` was found flaky because of the ordering of result HashMap. The orders of elements in the HashMap are not the same as every time the test being called. In this case, LinkedHashMap was applied to ensure the orders do not cause the flakiness. This code change is trying to fix the [flaky tests](https://docs.gitlab.com/ee/development/testing_guide/flaky_tests.html) mentioned above, because they sometimes fail (as the picture showed) and sometimes pass. The failure could be reproduced by the commands above by using the tool of [NonDex](https://github.com/TestingResearchIllinois/NonDex). The code change is to make sure the tests will always pass in this case.

- `org.apache.commons.collections4.CollectionUtilsTest.get`

The test failures could be reproduced by

1. `mvn install -pl . -am -DskipTests`
2.  run tests with NonDex 
    `mvn -pl . edu.illinois:nondex-maven-plugin:1.1.2:nondex -Dtest=org.apache.commons.collections4.CollectionUtilsTest#get`

### DIFF
--- a/src/test/java/org/apache/commons/collections4/CollectionUtilsTest.java
+++ b/src/test/java/org/apache/commons/collections4/CollectionUtilsTest.java
@@ -687,7 +687,9 @@ public class CollectionUtilsTest extends MockTestCase {
         assertEquals(2, CollectionUtils.get((Object) collectionA, 2));
         assertEquals(2, CollectionUtils.get((Object) collectionA.iterator(), 2));
         final Map<Integer, Integer> map = CollectionUtils.getCardinalityMap(collectionA);
-        assertEquals(map.entrySet().iterator().next(), CollectionUtils.get((Object) map, 0));
+        Map<Integer, Integer> linkedMap = new LinkedHashMap<>();
+        linkedMap.putAll(map);
+        assertEquals(linkedMap.entrySet().iterator().next(), CollectionUtils.get((Object) linkedMap, 0));
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


aherbert:
I believe this is fixing a non-issue artificially introduced by the `nondex` plugin. The HashMap has an unspecified iteration order, but it is a deterministic data structure. If it is not modified then the iterator will return the same order.

This is different from other changes identified by `nondex` (see #341, #353) where the test assumptions about ordered iteration were false.

However I think the suggested change is harmless and it will prevent others identifying the same issue in the future using the `nondex` plugin.

src/test/java/org/apache/commons/collections4/CollectionUtilsTest.java
```diff
@@ -687,7 +687,9 @@ public void get() {
         assertEquals(2, CollectionUtils.get((Object) collectionA, 2));
         assertEquals(2, CollectionUtils.get((Object) collectionA.iterator(), 2));
         final Map<Integer, Integer> map = CollectionUtils.getCardinalityMap(collectionA);
-        assertEquals(map.entrySet().iterator().next(), CollectionUtils.get((Object) map, 0));
+        Map<Integer, Integer> linkedMap = new LinkedHashMap<>();
```
aherbert, line 690: Add a comment to explain why this is done:
```Java
// Test assumes a defined iteration order so convert to a LinkedHashMap
```
Please use final:
```Java
final Map<Integer, Integer> linkedMap = ...
```

yannizhou05: Thank you so much for the clarification. I have updated the code accordingly.

PR #391, opened 2023-04-16
[COLLECTIONS-839] migrate Map tests to use junit5 assert

samabcde:
As part of the migration to Junit 5, target to replace all assertion api to Junit 5, for all Map related test class.
So there is no need for `BulkTest`  to extend `TestCase`.

### DIFF
--- a/src/test/java/org/apache/commons/collections4/AbstractArrayListTest.java
+++ b/src/test/java/org/apache/commons/collections4/AbstractArrayListTest.java
@@ -16,7 +16,9 @@
  */
 package org.apache.commons.collections4;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.util.ArrayList;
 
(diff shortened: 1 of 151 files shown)


garydgregory (changes requested):
Hi @samabcde 
Let's keep the Assertions API as static imports, please.

samabcde:
@garydgregory 

Thx for your comment. The reason I don't use static import is when I use static import, for instance `assertEquals`. 
The method will actually refer to `junit.framework.TestCase.assertEquals` instead of `org.junit.jupiter.api.Assertions.assertEquals`.
It only refer to `org.junit.jupiter.api.Assertions.assertEquals` when the argument don't match `junit.framework.TestCase.assertEquals` like `assertEquals(2, 2, "foo")`.

Some possible solution can be:
1. Use `Assertions.assertXXX` only for those method with `message` argument position changed.
  e.g. `Assertions.assertEquals(expected, actual, message). and keep others using `junit.framework.TestCase`, as they can be changed easily.
2. Complete all test classes API migration in one PR, hence `BulkTest` does not extend `TestCase` and we can use static import for all assert API.

Which one do you prefer? Or any other idea?

kinow (commented):
>Complete all test classes API migration in one PR, hence BulkTest does not extend TestCase and we can use static import for all assert API.

I think it'd be simpler to migrate that part first, or include that in this change? That way we would be sure that tests are passing with JUnit5 API, without running the risk of some JUnit4 API leaking into the tests and failing later when we change the rest of the code.

samabcde:
> > > Complete all test classes API migration in one PR, hence BulkTest does not extend TestCase and we can use static import for all assert API.
> > 
> > 
> > I think it'd be simpler to migrate that part first, or include that in this change? That way we would be sure that tests are passing with JUnit5 API, without running the risk of some JUnit4 API leaking into the tests and failing later when we change the rest of the code.
> 
> Agreed.

@garydgregory , @kinow updated as suggested, please kindly help to review.

PR #396, opened 2023-06-10
make EnhancedDoubleHasher.mod() public

Claudenw:
Fixed COLLECTIONS-840

### DIFF
--- a/src/main/java/org/apache/commons/collections4/bloomfilter/EnhancedDoubleHasher.java
+++ b/src/main/java/org/apache/commons/collections4/bloomfilter/EnhancedDoubleHasher.java
@@ -132,10 +132,10 @@ public class EnhancedDoubleHasher implements Hasher {
     /**
      * Performs a modulus calculation on an unsigned long and an integer divisor.
      * @param dividend a unsigned long value to calculate the modulus of.
-     * @param divisor the divisor for the modulus calculation.
+     * @param divisor the divisor for the modulus calculation, must be positive.
      * @return the remainder or modulus value.
      */
-    static int mod(final long dividend, final int divisor) {
+    public static int mod(final long dividend, final int divisor) {
         // See Hacker's Delight (2nd ed), section 9.3.
         // Assume divisor is positive.
         // Divide half the unsigned number and then double the quotient result.


aherbert:
-1 to simply exposing the internal method. Let's put it somewhere more logical in the package hierarchy, e.g. BitMap.

garydgregory (changes requested):
Hi @Claudenw 
Would you please add a unit test in `BitMapTest` for the new `mod()` public API? Make sure to establish expected behavior for normal and edge cases, especially for what should happen if overflow or underflow takes place.
TY!

src/test/java/org/apache/commons/collections4/bloomfilter/EnhancedDoubleHasherTest.java
```diff
@@ -99,7 +99,7 @@ void testModEdgeCases() {
         for (final long dividend : new long[] {-1, -2, -3, -6378683, -23567468136887892L, Long.MIN_VALUE, 345, 678686,
             67868768686878924L, Long.MAX_VALUE}) {
             for (final int divisor : new int[] {1, 2, 3, 5, 13, Integer.MAX_VALUE}) {
-                assertEquals((int) Long.remainderUnsigned(dividend, divisor), EnhancedDoubleHasher.mod(dividend, divisor),
+                assertEquals((int) Long.remainderUnsigned(dividend, divisor), BitMap.mod(dividend, divisor),
```
aherbert, line 102: @garydgregory I think your issue would be solved if @Claudenw moves this test to the BitMapTest. It is no longer applicable to the EnhancedDoubleHasher.

aherbert: When doing this can you add an extra case for `dividend=0`. This is not currently covered.

I don't think we need edge cases to show behaviour when the divisor is negative. This method is written for performance and omits checks for bad arguments. However the javadoc can be updated to make it a user beware method. It should add strong wording that the result is undefined if the divisor is negative, and there will be an java.lang.ArithmeticException if the divisor is zero.

PR #398, opened 2023-06-14
Changes required to open testing to new implementations.

Claudenw:
Makes some previously package private classes and methods public so that they may be used in developing tests for new Bloom filter implementations.

In addition, a correction for an error where specific cases can lead to a negative value for an estimated intersection.  Test case is provided.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/bloomfilter/BloomFilter.java
+++ b/src/main/java/org/apache/commons/collections4/bloomfilter/BloomFilter.java
@@ -305,6 +305,7 @@ public interface BloomFilter extends IndexProducer, BitMapProducer {
             // maximum estimate value using integer values is: 46144189292 thus
             // eThis + eOther can not overflow the long value.
             estimate = Math.round(eThis + eOther - eUnion);
+            estimate = estimate < 0 ? 0 : estimate;
         }
         return estimate>Integer.MAX_VALUE?Integer.MAX_VALUE:(int) estimate;
     }
--- a/src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBloomFilterTest.java
+++ b/src/test/java/org/apache/commons/collections4/bloomfilter/AbstractBloomFilterTest.java
@@ -216,9 +216,9 @@ public abstract class AbstractBloomFilterTest<T extends BloomFilter> {
 
     @Test
     public final void testNegativeIntersection() {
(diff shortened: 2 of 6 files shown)


garydgregory:
Hi @Claudenw 
You can catch these build failures early by running `mvn` from the command line. This will run the default Maven goal which the GHA build also uses.

aherbert:
This doesn't justify why the classes need to be `public`. Are you doing development in a different package?

Claudenw:
> This doesn't justify why the classes need to be `public`. Are you doing development in a different package?

I am.  I plan on submitting the work back and would like to discuss it.

The code is a stable Bloom filter which has some different properties from a normal Bloom filter.  It is somewhat like a counting Bloom filter except that:

- It counts down from a fairly low number.  A byte or smaller.
- It automatically decrements on an insert.

It is used in streaming data to detect duplicates in what is effectively a sliding window.

I was thinking of submitting it in a the package org.apache.commons.collections4.bloomfilter.stable as there are several helper classes to go with the Bloom filter class itself.

aherbert:
IIRC we consolidated the hasher sub-package into the main BloomFilter package as all functionality is coherent. Unless the stable Bloom filter is implementing a substantially different interface then I do not see the need to put it in its own package. I would do the development in the same package and when ready for merge you can open a PR and discuss the best location (main package or sub-package) on the dev mailing list, or Jira.

So this PR can remove all the exposure to public and reduce it to the fix for the `estimateIntersection`  method.

garydgregory:
I like the idea of providing bloom filters all in one package to reduce the new public and protected API footprint. We can create a new sub-package and/or make elements public later if needed.

Claudenw:
I have one more implementation that I am working on for a Kafka fix.  In this case I am creating an implementation of layered Bloom filters, where the layers are added based on a predicate.  For Kafka a new layer will be added each minute and after 60 layers the first layer will be removed.  This structure causes problems with estimation as it is much more accurate that a single filter alone.  The estimate tests are all `final` so, while the new implementation passes all the other tests, it does not pass the estimateX tests.  For this reason I think that those tests should not be final.

Claudenw:
In thinking about the other cases, I have no problem putting them into the same package, however, when developing new classes of Bloom filters, it would be handy, as an external developer, to be able to utilize the standard tests without having to duplicate them.  To state the obvious, when new tests are added to the base and duplicates will need the addition, and so the developer will have to know that the addition was made.  It just makes sense that the testing be open so that new implementations can be developed.  As and example, I recently added a test to show that some filters can generate negative values for union.

garydgregory (approved):
LGTM

garydgregory:
> Member

Also, note that these are test classes that are being made public, there increasing the API footprint is NOT an issue.

garydgregory:
@aherbert I am OK to merge this, are you?

aherbert:
Sure. If it helps Claude develop his prototype before merge then this moves the project forward. I just questioned the making public of test methods and classes without a reason. Prototype development is reason enough.

PR #400, opened 2023-06-21
[COLLECTIONS-841] Open up bloom filter tests - test changes to support bit decay based Bloom filters

Claudenw:
Changes to test setup to account for Bloom filters that use bit decay (e.g. Stable Bloom filter).   These changes do not change the tests, just the way the filters are configured so that decay based filters will have the proper results.


garydgregory:
https://issues.apache.org/jira/browse/COLLECTIONS-841

PR #406, opened 2023-07-16
COLLECTIONS-844 - allow counting Bloom filters with cell size other than Integer.SIZE

Claudenw:
Counting bloom filter tests are dependant upon the maximum value that can be stored in a cell in a Bloom filter.

This pull adds a `getMaxValue()` method to the CountingBloomFilter interface.  This is to indicate the maximum value that can be stored in a single cell of the counting filter.  A cell is the counter that replaces the bit in the counting filter.

Also adds the missing standard function `getMaxInsert()` which returns the maximum times an item could have been merged into the counting filter.

Changes to tests

Fix for COLLECTIONS-844


aherbert:
Some extra checkstyle files were added. I hope to have time to review in the next few days.

aherbert:
The functional changes are fine, i.e. expose some measure of the capacity of the counts. But without an exact use case I am not certain of how it would be used.

I think this could do with some naming improvement. You cannot `get` values from a counting Bloom filter. So `getMaxValue` seems out of context. Plus no other docs in the class refer to a 'cell'. The filter always uses the term 'bit index'.

Should this property instead reflect that it is the maximum count of the same bit index that can be added to the filter before the state is invalidated. So change to `getMaxCount` and document that it is the maximum count that can be held for any bit index. If this is exceeded then the filter is invalidated (see `#isValid`).

Q. Is there a reference computation for an estimate of the number of items that can be added before saturating a bit index with a certain probability. It may be useful to point a reader in the direction of some calculations. Thus they can use the Shape and this property to get an estimate of capacity.

Claudenw:
@aherbert 

Cell is a standard term in the literature.  I am preparing two more pull requests.  

1. A stable Bloom filter (weird beast) that acts sorta like a counting Bloom filter with random deletions.
2. Another implementation of counting Bloom filter that uses the core classes from stable Bloom filter to implement counting Bloom filters that can use from 1 to 32 bits (1 is a bit ridiculous but it is supported).

Both the new counting implementation and the stable Bloom filter are dependant upon a CellManager class and some supporting classes.  I could present the pull requests in either order, the first one being largish (like the layerd change) with the second being a couple of classes for the specific implementation. 

I can update the current counting Bloom filter with the term "cell", but I would like to keep the terminology.  I thought I had change the method name to `getMaxCount`, I'll figure out where that change went.

Would you be amenable to updating the documentation to refer to cell when we have "bits" that are values more than bits?

aherbert:
There are two terms here used through the API: _bit index_ which refers to a position within the shape; and the _count_ of the number of observations of that index. In that context you are stating a cell is a position that can be filled with counts. So it is a countable bit index. Would the term _cell_ ever be used for non-counting Bloom filters?

Where does that leave the naming of BitCountProducer? CellCountProducer; CellProducer? This is why I questioned it. Adding new terminology should make sense to those reading the API for the first time. Perhaps it can be done by adding a note to the class javadoc for CountingBloomFilter that bit indices that can be counted are also known as cells (if this is correct).

Claudenw:
Cells are used in stable Bloom filters which are technically not a counting bloom filter they are in a class called "Element decay" Bloom filters.  They do use cells, not bits, but only report and accept bits as output and input.  There are several Bloom filters in this class.

This change is to make the development of the stable and a variable size CountingBloomFilter possible.  In that change I was going to propose changing `BitCountProducer` and `BitCountConsumer` to `CellProdudcer` and `CellConsumer` respectively and change the method call from `forEachCount` to `forEachCell`.  However, I see that it may make sense to do that in this change to minimize the changes for the stable and variable counting filters.

Are you ok with my proceeding to;

1. Update the documentation to talk about "cells"
2. Change BitCountProducer/BitCountConsumer to CellProducer/CellConsumer as noted above?
3. Change `forEachCount` to `forEachCell`

aherbert:
Go ahead. If if more closely matches the literature then this is what we should do.

src/main/java/org/apache/commons/collections4/bloomfilter/CountingBloomFilter.java
```diff
@@ -164,17 +221,17 @@ default boolean merge(final BitMapProducer bitMapProducer) {
     /**
      * Removes the specified Bloom filter from this Bloom filter.
      *
-     * <p>Specifically: all counts for the indexes identified by the {@code other} filter will be decremented by 1.</p>
+     * <p>Specifically: all cells for the indexes identified by the {@code other} filter will be decremented by 1.</p>
      *
-     * <p>Note: If the other filter is a counting Bloom filter the index counts are ignored and it is treated as an
+     * <p>Note: If the other filter is a counting Bloom filter the othre filter's cells are ignored and it is treated as an
```
aherbert, line 226: `othre`

src/main/java/org/apache/commons/collections4/bloomfilter/CountingBloomFilter.java
```diff
@@ -77,22 +77,79 @@ public interface CountingBloomFilter extends BloomFilter, BitCountProducer {
...
+    }
+
+    /**
+     * Determines the maximum number of times the IndexProducer could have been inserted
+     * into this counting filter.
+     * @param idxProducer the producer to drive the count check.
+     * @return the maximum number of times the IndexProducer could have been inserted.
+     */
+    default int getMaxInsert(IndexProducer idxProducer) {
+        return getMaxInsert( BitMapProducer.fromIndexProducer(idxProducer, getShape().getNumberOfBits()));
```
aherbert, line 103: whitespace `( BitMap`

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractCountingBloomFilterTest.java
```diff
@@ -289,6 +291,50 @@ public void testExcludesDuplicates() {
...
+
+    private void verifyMaxInsert( CountingBloomFilter bf, int from1, int from11) {
+        BloomFilter bfFrom0 = new DefaultBloomFilterTest.SparseDefaultBloomFilter(getTestShape());
+        bfFrom0.merge(new IncrementingHasher(0, 1));
+        BloomFilter bfFrom1 = new DefaultBloomFilterTest.SparseDefaultBloomFilter(getTestShape());
+        bfFrom1.merge(TestingHashers.FROM1);
+        BloomFilter bfFrom11 = new DefaultBloomFilterTest.SparseDefaultBloomFilter(getTestShape());
+        bfFrom11.merge(TestingHashers.FROM11);
+
+        assertEquals( 0, bf.getMaxInsert(new IncrementingHasher(0, 1)));
```
aherbert, line 305: No whitespace after `assertEquals(`

aherbert (changes requested):
Because you squashed and force pushed to the branch the comments from the last review I made has been lost. FYI: don't do this until the PR is ready to merge.

Note: I started by adding a single comment as I clicked the wrong button in GH. I later noted that you explain what cells are in the package-info. This should probably be repeated in the appropriate interfaces/classes as not many people will read the package info. Simple changing `counts` to `cells` makes some of the grammar less intuitive. This could use some work.

I cannot remember everything I stated last time. Overall the PR looks OK. I didn't do a deep dive because there is an issue with the `getMaxIndex` I raised last time that you have not addressed. It does not count repeats of the same index in the query item.
```java
@Test
void test() {
    ArrayCountingBloomFilter f = new ArrayCountingBloomFilter(Shape.fromKM(1, 10));
    IndexProducer ip = p -> { 
        p.test(3);
        p.test(3);
        return true;
    };
    f.merge(ip);
    System.out.println(f.getMaxInsert(ip));
}
```
Prints:
```
2
```
This is non-sense. It could be fixed. If you don't care that the returned value is an overestimate then the method should be documented with warnings. The value from `getMaxInsert` is actually returning the number of times a cell touched by your item has been incremented, and not a factor of the number of times your item has been inserted. As such it is an overestimate of any item that touches the same cell multiple times.

Claudenw:
I fixed the issues with getMaxInsert by requiring the CellProducer return distinct ordered indices.  It made much of the code simpler, but added more test cases.  I will go back and review all the places where Cell is referenced in java doc and ensure that there is some sort of description.

src/main/java/org/apache/commons/collections4/bloomfilter/CountingBloomFilter.java
```diff
@@ -19,7 +19,7 @@
 import java.util.Objects;
 
 /**
- * The interface that describes a Bloom filter that associates a count with each
+ * The interface that describes a Bloom filter that associates a cell with each
```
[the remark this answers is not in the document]
Claudenw: Definition of cell and usage of cell vs count was reviewed and corrections made

src/main/java/org/apache/commons/collections4/bloomfilter/CountingBloomFilter.java
```diff
@@ -121,7 +186,7 @@ default boolean merge(final Hasher hasher) {
     /**
      * Merges the specified index producer into this Bloom filter.
      *
-     * <p>Specifically: all counts for the indexes identified by the {@code indexProducer} will be incremented by 1.</p>
+     * <p>Specifically: all cells for the indexes identified by the {@code indexProducer} will be incremented by 1.</p>
```
aherbert, line 189: I don't think your conversion to a CellProducer upholds this contract. The cells will not be incremented by 1; they will be incremented by the count of the indices because merge then calls add.

IIUC a merge is from BloomFilter and should flatten all arguments to an effective bitmap, removing duplicates. The add operation should not remove duplicate indices because they are treated as cells with a count above 1.

We should have some tests to check these definitions are upheld.

```Java
ArrayCountingBloomFilter f1 = new ArrayCountingBloomFilter(Shape.fromKM(1, 10));
ArrayCountingBloomFilter f2 = f1.copy();
ArrayCountingBloomFilter f3 = f1.copy();
IndexProducer ip = p -> {
    p.test(3);
    p.test(3);
    return true;
};
// The merge SHOULD increment cells by 1
f1.merge(ip);
// The add should increment cells by 2
f2.add(CellProducer.from(ip));
// This merge will increment by 1 as the round-trip makes the indices unique
f3.merge(IndexProducer.fromIndexArray(ip.asIndexArray()));

f1.forEachCell((k, v) -> {
    System.out.printf("%d = %d%n", k, v);
    return true;
});
f2.forEachCell((k, v) -> {
    System.out.printf("%d = %d%n", k, v);
    return true;
});
f3.forEachCell((k, v) -> {
    System.out.printf("%d = %d%n", k, v);
    return true;
});
```
Prints:
```
3 = 2
3 = 2    // ERROR !
3 = 1
```

Claudenw: Agree this is incorrect.   I am adding this as a test to the AbstractCountingFilterTest.

I think that we should make the following changes.

for counting filters

- merge always increments the cell value by at most 1.
- remove always decrements the cell value by at most 1.
- add and subtract as used to increment/decrement by more than 1.

for index producers

- as array always returns the indices in the number and order they are produced.
- move the "unique" method from hashers to index producers. Hashers create index producers.  Let the use of the producer determine if it needs to be unique.

aherbert: Agree with all points except:

"as array always returns the indices in the number and order they are produced."

This constraint may impact performance. Is it required?

A hasher asIndexArray may not wish to do this as the indices are random. The EnhancedDoubleHasher current returns the indices unordered.

Claudenw: Is it "required" no, but when working in complex arrangements it mean that the simply converting to an array does not change the state of the data.  In most cases this is not an issue.   The only case where it is an issue is when a Hasher backed IndexProducer.asArray() is called.  In all other cases (I think) the array either already exists or exists in a different form (BitMaps, CellProducers).   BitMaps and CellProducers always return distinct values in order.

My plan was to lift the code from BitMapProducer that produces bitMaps and use it to produce an array of indices from a Hasher.

Perhaps my original comment is not clear.  Basically, however the underlying storage produces the indices is the order and number that will be produced in the array.

aherbert: If you wish to return a distinct array then the default method in the interface using a bit set will be fine.

If you wish to return a count of duplicate indices then the current implementation will work if you just sort the array it currently creates. This may be faster than using a complicated intermediate data structure. The hasher is likely to have a small array and the speed of the sort will be fast.

This may be an opportunity to use some more characteristics flags where an index producer can declare if the indices are sorted and or duplicates. Then the interface can provide default methods to remove duplicates or sort them.

Claudenw: If we have IndexProducer.asArray output the indices in the order that they would be returned by IndexProducer.forEachIndex() then we are not changing the data based on the representation.

This also means that sorting can be done by generating an array, using Array.sort() and then using the result to create an IndexProducer.fromIndexArray() to create the sorted index.

The characteristics flag may be a good idea. We already have them in the test code.

How about we move this forward as is and merge it.  I think the change is large enough (I was trying to keep is small).

I still need to get the LayeredBloom filter added and it will need to be updated once we have this in place. 
Finally, I have the StableBloom filter, which may clarify some of the questions we have a bout characteristics and whether we have chosen the proper restrictions on IndexProducer and CellProducers.

Claudenw: The last bit of my previous message was not meant to indicate that this should not be merged, just that there are 2 more changes coming as different pull requests.

aherbert: I misread your previous statement "as array always returns the indices in the number and order they are produced." I took that to mean in order, i.e. sorted. Re-reading it I see that you are stating that the order matches the output produced by forEachIndex.

This is fine. But it now means the default implementation of asIndexArray (which creates sorted, unique indices) is now in conflict with the javadoc.

src/main/java/org/apache/commons/collections4/bloomfilter/IndexProducer.java
```diff
@@ -117,11 +118,64 @@ public boolean test(long word) {
...
-        final BitSet result = new BitSet();
+        class Indices {
+            private int[] data = new int[32];
+            private int size;
+
+            boolean add(final int index) {
+                if (size == data.length) {
+                    // This will throw an out-of-memory error if there are too many bits.
+                    // Since bits are addressed using 32-bit signed integer indices
+                    // the maximum length should be ~2^31 / 2^6 = ~2^25.
```
aherbert, line 129: This is wrong. You are not using the int[] as a bitmap. You are using it to store indices. So you can get out of memory. However at some point you will not be able to return the array if it needs more than the max capacity of an array. A larger limit is:
```java
data = Arrays.copyOf(data, (int) Math.min(MAX_ARRAY_SIZE, size * 2L);
```
with MAX_ARRAY_SIZE something big. Previous versions of the JDK java.util.ArrayList use Integer.MAX_VALUE - 8.

src/main/java/org/apache/commons/collections4/bloomfilter/CountingBloomFilter.java
aherbert: PS. I just saw that you changed the default implementation of asIndexArray to respect the order of forEachIndex. I noted a problem with the internal comment in that method about the capacity.

Claudenw: Should all be fixed and ready to go now.

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractCellProducerTest.java
```diff
@@ -0,0 +1,154 @@
...
+    @Test
+    public void testForEachCellEarlyExit() {
+        final int[] passes = new int[1];
+        assertTrue(createEmptyProducer().forEachCell((i, j) -> {
+            passes[0]++;
+            return false;
+        }));
+        assertEquals(0, passes[0]);
+
+        assertFalse(createProducer().forEachCell((i, j) -> {
```
aherbert, line 148: This test heavily assumes that the `createProducer` will have more than 1 cell populated so you can early exit.

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractCellProducerTest.java
```diff
@@ -0,0 +1,154 @@
...
+            return true;
+        });
+        assertEquals(bs1, bs2);
+    }
+
+    @Test
+    public void testForEachCellValues() {
+        int[] expectedIdx = getExpectedIndices();
+        int[] expectedValue = getExpectedValues();
+        assertEquals( expectedIdx.length, expectedValue.length, "expected index length and value length do not match");
```
aherbert, line 111: whitespace

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractCellProducerTest.java
```diff
@@ -0,0 +1,154 @@
...
+        int[] idx = {0};
+        createProducer().forEachCell((i, j) -> {
+            assertEquals(expectedIdx[idx[0]], i, "bad index at "+idx[0]);
```
aherbert, line 114: `" + idx`
Same on next line

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractCellProducerTest.java
```diff
@@ -0,0 +1,154 @@
...
+     * A testing CellConsumer that always returns true.
+     */
+    private static final CellConsumer TRUE_CONSUMER = (i, j) -> true;
+    /**
+     * A testing CellConsumer that always returns false.
+     */
+    private static final CellConsumer FALSE_CONSUMER = (i, j) -> false;
+
+    /**
+     * Creates an array of expected values that alignes with the expected indices entries.
```
aherbert, line 43: `aligns`

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractCellProducerTest.java
aherbert, line 1: This file has lost the git history from `AbstractBitCountProducerTest`. It is showing as a new file and the old test as deleted.

src/main/java/org/apache/commons/collections4/bloomfilter/CellProducer.java
aherbert, line 1: This file has lost the git history from `BitCountProducer`. It is showing as a new file and the old interface as deleted.

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractCountingBloomFilterTest.java
```diff
@@ -289,6 +296,114 @@ public void testExcludesDuplicates() {
         bf1.merge(hasher);
         bf1.remove(hasher);
         assertEquals(0, bf1.cardinality());
-        assertTrue(bf1.forEachCount((x, y) -> false), "Hasher in removes results in value not equal to 0");
+        assertTrue(bf1.forEachCell((x, y) -> false), "Hasher in removes results in value not equal to 0");
+    }
+
+    private void verifyMaxInsert( CountingBloomFilter bf, int from1, int from11) {
```
aherbert, line 302: `(Counting`

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractCountingBloomFilterTest.java
```diff
@@ -289,6 +296,114 @@ public void testExcludesDuplicates() {
...
+        assertEquals(1, bf.getMaxInsert(new IncrementingHasher(5, 1)));
+        bf.remove(new IncrementingHasher(5, 1));
+        verifyMaxInsert(bf, 0, 0);
+        assertEquals(0, bf.getMaxInsert(new IncrementingHasher(5, 1)));
+    }
+
+    private void assertCell3(CountingBloomFilter bf, int value) {
+        bf.forEachCell((k, v) -> {
+            if (k == 3) {
+                assertEquals(value, v, "Mismatch at position (3) "+k);
```
aherbert, line 349: Drop the `+k`, or drop the `(3)` as they are the same.

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractCountingBloomFilterTest.java
```diff
@@ -289,6 +296,114 @@ public void testExcludesDuplicates() {
...
+            } else {
+                assertEquals(0, v, "Mismatch at position "+k);
```
aherbert, line 351: `" + k`

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractCountingBloomFilterTest.java
```diff
@@ -289,6 +296,114 @@ public void testExcludesDuplicates() {
...
+        };
+        // The merge should increment cell 3 by 1
+        f1.merge(ip);
+        assertCell3(f1, 1);
+
+        // The add should increment cells 3 by 2
+        f2.add(CellProducer.from(ip));
+        assertCell3(f2, 2);
+
+        // This merge will increment by 1 as the round-trip makes the indices unique
```
aherbert, line 376: This test is redundant. The round trip does not make the indices unique. So the test is effectively:
```java
f3.merge(ip)
```
This is same as the test using f1.

src/main/java/org/apache/commons/collections4/bloomfilter/IndexProducer.java
```diff
@@ -117,11 +120,60 @@ public boolean test(long word) {
      * @return An int array of the data.
      */
     default int[] asIndexArray() {
-        final BitSet result = new BitSet();
+        class Indices {
```
aherbert, line 123: Note: This method now avoids the BitSet. As such I think we can remove the javadoc warning that the default implementation is slow. The note about returning in unique values in order is wrong. The default implementation now returns an array using the same order and cardinality as the `forEachIndex` method.

src/test/java/org/apache/commons/collections4/bloomfilter/CellProducerFromDefaultIndexProducerTest.java
```diff
@@ -0,0 +1,45 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.collections4.bloomfilter;
+
+public class CellProducerFromDefaultIndexProducerTest extends AbstractCellProducerTest {
+
+    int[] data = { 0, 63, 1, 64, 128, 1, 127 };
```
aherbert, line 21: whitespace

src/test/java/org/apache/commons/collections4/bloomfilter/CellProducerFromArrayCountingBloomFilterTest.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Override
+    protected int[] getExpectedIndices() {
+        return new int[] {0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21};
+    }
+
+    @Override
+    protected int[] getExpectedValues() {
+        return new int[] { 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 1, 1, 1, 1, 1};
```
aherbert, line 43: whitespace after {

src/test/java/org/apache/commons/collections4/bloomfilter/DefaultCellProducerTest.java
```diff
@@ -16,21 +16,27 @@
...
         return values;
     }
 
     @Override
-    protected BitCountProducer createProducer() {
+    protected CellProducer createProducer() {
         return consumer -> {
-            for (final int i : values) {
-                if (!consumer.test(i, 1)) {
+            for (int i=0; i<indices.length; i++) {
```
aherbert, line 38: `for (int i = 0; i < indices.length; i++) {`

src/test/java/org/apache/commons/collections4/bloomfilter/DefaultIndexProducerTest.java
```diff
@@ -119,4 +118,42 @@ public void testFromIndexArray() {
             assertArrayEquals(expected, ip.asIndexArray());
         }
     }
+
+    @Test
+    public void testMoreThan32Entries() {
+        int[] values = new int[33];
+        for (int i=0; i<33; i++) {
```
aherbert, line 125: int[] values = IntStream.range(0, 33).toArray();

src/test/java/org/apache/commons/collections4/bloomfilter/DefaultIndexProducerTest.java
```diff
@@ -119,4 +118,42 @@ public void testFromIndexArray() {
...
+                }
+            }
+            return true;
+        };
+        int[] other = producer.asIndexArray();
+        assertArrayEquals(values, other);
+    }
+
+    @Test
+    public void test32Entries() {
```
aherbert, line 142: You can combine this with the test above as:
```java
@ParameterizedTest
@ValueSource(ints = {32, 33})
void testEntries(int size) {
    int[] values = IntStream.range(0, size).toArray();
```

aherbert (changes requested):
Can you try and update the PR so that the git history is kept for two of the renamed files. You should be able to restore the files from an old commit, use `git mv` to rename them and then replace the contents with the new code files you have here.

Please remove changes to IndexFilter. They may be useful. They are not related to this change.

src/main/java/org/apache/commons/collections4/bloomfilter/CellProducer.java
```diff
@@ -0,0 +1,170 @@
...
+                @Override
+                public int compareTo(CounterCell other) {
+                    return Integer.compare( idx,  other.idx);
+                }
+            }
+        };
+    }
+
+    /**
+     * Represents an operation that accepts an {@code <index, cell>} pair.
```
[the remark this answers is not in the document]
Claudenw: fixed

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractCellProducerTest.java
Claudenw: Every action I tried left github with delete and new.  Local git showed renamed during commit.

src/main/java/org/apache/commons/collections4/bloomfilter/IndexProducer.java
```diff
@@ -32,6 +33,8 @@
 @FunctionalInterface
 public interface IndexProducer {
 
+    int MAX_ARRAY_SIZE = Integer.MAX_VALUE - 8;
```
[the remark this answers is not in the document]
Claudenw: done

src/main/java/org/apache/commons/collections4/bloomfilter/IndexProducer.java
```diff
@@ -117,11 +120,60 @@ public boolean test(long word) {
...
+        }
+        Indices indices = new Indices();
+        forEachIndex(indices::add);
+        return indices.toArray();
+    }
+
+    /**
+     * Creates an IndexProducer of unique indices for this index.
+     *
+     * <p>This is like the `indices(Shape)` method except that it adds the guarantee that no
```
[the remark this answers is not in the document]
Claudenw: fixed and adde more info

src/main/java/org/apache/commons/collections4/bloomfilter/IndexProducer.java
```diff
@@ -117,11 +120,60 @@ public boolean test(long word) {
...
+                return size == data.length ? data : Arrays.copyOf(data, size);
+            }
+        }
+        Indices indices = new Indices();
+        forEachIndex(indices::add);
+        return indices.toArray();
+    }
+
+    /**
+     * Creates an IndexProducer of unique indices for this index.
```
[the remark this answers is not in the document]
Claudenw: changed text

src/main/java/org/apache/commons/collections4/bloomfilter/CountingBloomFilter.java
```diff
@@ -103,40 +172,41 @@ default boolean merge(final BloomFilter other) {
...
-     * @see #add(BitCountProducer)
+     * @see #add(CellProducer)
      */
     @Override
     default boolean merge(final IndexProducer indexProducer) {
         Objects.requireNonNull(indexProducer, "indexProducer");
         try {
-            return add(BitCountProducer.from(indexProducer));
+            return add(CellProducer.from(indexProducer.uniqueIndices()));
         } catch (final IndexOutOfBoundsException e) {
```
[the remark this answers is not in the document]
Claudenw: We do this because BloomFilter throws and IndexOutOfBoundsException if any index is <0 or >=bitCount.

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractCellProducerTest.java
Claudenw: The test requires that the producer have at least 1 entry and we force an exit on the first one.

src/main/java/org/apache/commons/collections4/bloomfilter/CellProducer.java
Claudenw: Seems to be an issue with github display local git showed that it was renamed during commit.

src/main/java/org/apache/commons/collections4/bloomfilter/IndexProducer.java
Claudenw: updated

src/test/java/org/apache/commons/collections4/bloomfilter/DefaultIndexProducerTest.java
Claudenw: fixed

aherbert:
I think you have fixed it. The GH UI sometimes does not show the renamed file as a move. I have cloned the latest PR code and verified locally that the history is there:
```
git log --follow  src/main/java/org/apache/commons/collections4/bloomfilter/CellProducer.java
git log --follow  src/test/java/org/apache/commons/collections4/bloomfilter/AbstractCellProducerTest.java
```
Log revisions date back to 2022.

I should have done this before I asked you to restore the history as it may have been there already. Anyway, it is OK now.

I think we should revert the IndexFilter change and do that separately. This PR is related to the counting Bloom filters and their ability to indicate the max capacity of a cell and estimate the number of entries of an item.

src/test/java/org/apache/commons/collections4/bloomfilter/AbstractCellProducerTest.java
aherbert: My point was that the test is not checking the producer has more than 1 entry: it is assuming this. Thus it cannot ensure that the exit was early. If the producer only has 1 entry then there is no way to distinguish an early exit. But at least you can mark the test as skipped. To do this would require the test for the empty producer is separated.

If we separate the test of the non-empty producer then a skipped test will be identified in the JUnit summary and should prompt an investigation as to why it was skipped (i.e. the producer only produces 1 cell).

Claudenw:
I think that IndexUtils method should be moved to ArrayUtils in the root package.

aherbert:
> I think that IndexUtils method should be moved to ArrayUtils in the root package.

Note that class is package private, and copied and adapted from Commons Lang.

To make the method public requires a bit more effort. For example duplicating for all 8 primitive array types and also a generic type. It may also require adding a method to expand the array for adding more than 1 item. (The current method only requires expanding the array to add 1 item.) This generic functionality may be better suited to ArrayUtils in commons-lang.

Claudenw:
OK.  If there are any issues remaining please let me know.

src/main/java/org/apache/commons/collections4/bloomfilter/IndexUtils.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+package org.apache.commons.collections4.bloomfilter;
+
+import java.util.Arrays;
+
+/**
+ * Provides functions to assist in IndexProducer creation and manipulation.
+ * @see IndexProducer
+ */
+public class IndexUtils {
```
aherbert, line 25: Should not be public; should be final.

src/main/java/org/apache/commons/collections4/bloomfilter/IndexUtils.java
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * The maximum array size for the methods in this class.
+     */
+    public static final int MAX_ARRAY_SIZE = Integer.MAX_VALUE - 8;
```
aherbert, line 30: Should not be public

src/main/java/org/apache/commons/collections4/bloomfilter/IndexUtils.java
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * Ensure the array can add an element at the specified index.
+     * @param array the array to check.
+     * @param index the index to add at.
+     * @return the array or a newly allocated copy of the array.
+     */
+    static int[] ensureCapacityForAdd(int[] array, int index) {
+        if (index >= array.length) {
+            return Arrays.copyOf(array, (int) Math.min(IndexUtils.MAX_ARRAY_SIZE, Math.max( array.length * 2L, index+1)));
```
aherbert, line 43: whitespace: `Math.max(array.length * 2L, index + 1)`

src/main/java/org/apache/commons/collections4/bloomfilter/CellProducer.java
```diff
@@ -115,14 +110,14 @@ private void populate() {
             @Override
             public int[] asIndexArray() {
                 populate();
-                return counterCells.keySet().stream().mapToInt( c -> c.idx ).toArray();
+                return counterCells.keySet().stream().mapToInt(c -> c.idx ).toArray();
```
aherbert, line 113: Still trailing whitespace: `(c -> c.idx)`

src/main/java/org/apache/commons/collections4/bloomfilter/CellProducer.java
```diff
@@ -143,14 +138,14 @@ final class CounterCell implements Comparable<CounterCell> {
 
                 @Override
                 public int compareTo(CounterCell other) {
-                    return Integer.compare( idx,  other.idx);
+                    return Integer.compare(idx,  other.idx);
```
aherbert, line 141: No need for two spaces after the comma

src/main/java/org/apache/commons/collections4/bloomfilter/IndexProducer.java
```diff
@@ -125,9 +121,7 @@ class Indices {
             private int size;
 
             boolean add(final int index) {
-                if (size == data.length) {
-                    data = Arrays.copyOf(data, (int) Math.min(MAX_ARRAY_SIZE, size * 2L));
-                }
+                data = IndexUtils.ensureCapacityForAdd(data, index);
```
aherbert, line 124: This is wrong. It should use `size` when checking the capacity:
```java
data = IndexUtils.ensureCapacityForAdd(data, size);
```
I put this into `IndexProducerTest`:
```java
@ParameterizedTest
@ValueSource(ints = {32, 33})
void testAsIndexArray(int n) {
    IndexProducer ip = i -> {
        for (int j = 0; j < n; j++) {
            // Always test index zero
            i.test(0);
        }
        return true;
    };
    Assertions.assertArrayEquals(new int[n], ip.asIndexArray());
}
```
The size 33 fails as the initial array of capacity 32 is not enlarged.

src/main/java/org/apache/commons/collections4/bloomfilter/IndexFilter.java
```diff
@@ -72,7 +72,10 @@ public boolean test(final int number) {
         if (number >= size) {
             throw new IndexOutOfBoundsException(String.format("number too large %d >= %d", number, size));
         }
-        return !tracker.test(number) || consumer.test(number);
+        if (tracker.test(number)) {
```
aherbert, line 75: You reverted the IndexFilter file but left this change. I think the statements are identical. Since the plan is to replace this with a separate change to the IndexFilter then can you revert this too. The IndexFilter should be unchanged by this PR.

aherbert (changes requested):
Sorry for the delay in reviewing. There is an issue with the use of the IndexUtils to enlarge the array. The class should also be a package-private final utility class.

src/main/java/org/apache/commons/collections4/bloomfilter/IndexFilter.java
Claudenw: This change fixes code coverage check.  In the original case there are 4 branches, one of which can not be tested.  In this case there are only 3 and the code coverage report shows 100% coverage for the Bloomfilter code.

src/main/java/org/apache/commons/collections4/bloomfilter/IndexUtils.java
Claudenw: fixed

src/main/java/org/apache/commons/collections4/bloomfilter/IndexUtils.java
Claudenw: fixed

PR #407, opened 2023-07-25
Make AbstractPatriciaTrie public

vad0:
I extend this class [here](https://github.com/vad0/gfjson/blob/master/src/main/java/org/apache/commons/collections4/trie/AsciiTrie.java) to make an implementation which works with an `AsciiSequenceView` instead of `String`. Package visibility forces me to create a package org.apache.commons.collections4.trie in my project. This hack works, but it prevents me from migrating to the java modules system, which prohibits using the same package names in the different modules. So I think the best solution here wil be to make AbstractPatriciaTrie public.


Claudenw:
I encountered a similar problem with AbstractPatriciaTrie but ended up not using the Trie.

garydgregory:
Seems reasonable to make public unless I am missing something. @aherbert ?

vad0:
Copied some javadocs from the parent class

PR #476, opened 2024-04-12
[COLLECTIONS-852] Add layerd bloom filter clean method

Claudenw:
Provides a method to call the LayerManager cleanup consumer without adding additional layers to the LayeredBloomFilter.

fix for [COLLECTIONS-852](https://issues.apache.org/jira/browse/COLLECTIONS-852)

### DIFF
--- a/src/main/java/org/apache/commons/collections4/bloomfilter/LayeredBloomFilter.java
+++ b/src/main/java/org/apache/commons/collections4/bloomfilter/LayeredBloomFilter.java
@@ -372,9 +372,18 @@ public class LayeredBloomFilter implements BloomFilter, BloomFilterProducer {
      * Forces and advance to the next layer. Executes the same logic as when
      * LayerManager.extendCheck returns {@code true}
      *
-     * @see LayerManager
+     * @see LayerManager#next()
      */
     public void next() {
         layerManager.next();
+    }
+
+    /**
+     * Forces the execution of {@code LayerManager.clean()}.
+     * 
+     * @see LayerManager#clear()
+     */
+    public void clean() {
+        layerManager.clean();
     }
 }
--- a/src/main/java/org/apache/commons/collections4/bloomfilter/LayeredBloomFilter.java
+++ b/src/main/java/org/apache/commons/collections4/bloomfilter/LayeredBloomFilter.java
@@ -369,21 +369,24 @@ public class LayeredBloomFilter implements BloomFilter, BloomFilterProducer {
     }
 
     /**
-     * Forces and advance to the next layer. Executes the same logic as when
-     * LayerManager.extendCheck returns {@code true}
+     * Forces and advance to the next layer. This method will clean-up the current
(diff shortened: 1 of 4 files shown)


src/test/java/org/apache/commons/collections4/bloomfilter/LayeredBloomFilterTest.java
```diff
@@ -311,4 +313,36 @@ public final void testNext() {
         assertFalse(filter.get(1).contains(TestingHashers.FROM11));
         assertTrue(filter.get(1).contains(new IncrementingHasher(11, 2)));
     }
+
+    @Test
+    public void testClean() {
```
aherbert, line 318: I think this test is made redundant by testing the method on the LayeredBloomFilter. In the interest of removing code duplication I would drop this and just leave the test calling the public API on the filter.

aherbert (changes requested):
The changes are fine but the documentation should describe the functionality given that the methods in LayerManager are package-private (so cannot be referenced in javadoc).

Reading through this I believe the use of **clean** and **clear** could create coding errors. WDYT?

src/main/java/org/apache/commons/collections4/bloomfilter/LayeredBloomFilter.java
```diff
@@ -372,9 +372,18 @@ public boolean merge(IndexProducer indexProducer) {
...
      *
-     * @see LayerManager
+     * @see LayerManager#next()
      */
     public void next() {
         layerManager.next();
     }
+
+    /**
+     * Forces the execution of {@code LayerManager.clean()}.
```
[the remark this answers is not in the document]
Claudenw: Fixed this as well as a similar issue with the `next()` documentation as well.

src/main/java/org/apache/commons/collections4/bloomfilter/LayeredBloomFilter.java
```diff
@@ -372,9 +372,18 @@ public boolean merge(IndexProducer indexProducer) {
...
+     *
+     * @see LayerManager#clear()
```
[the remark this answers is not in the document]
Claudenw: cleanup was used to refer to this is several places so I change to `cleanup` as the method name.

src/test/java/org/apache/commons/collections4/bloomfilter/LayeredBloomFilterTest.java
Claudenw: I am a but confused.  this is the LayeredBloomFilterTest.  I assume you mean the  LayeredManagerTest should have the `testClean` removed.

aherbert: Yes, drop the LayerManagerTest (as a duplicate). Sorry for the mistake.

aherbert (approved):

PR #481, opened 2024-04-25
COLLECTIONS-853: Change LayerManager to use List and added generics to LayerdedBloomFilter

Claudenw:
Changes to the is code:

1. Added generics to LayeredBloomFilter and LayerManager so that the implementation of the enclosed Bloom filter can be specified.
2. Modified WrappedBloomfFilter to require copy() implementation.
3.  Changed the LayerManager internal LinkedList declaration to List.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/bloomfilter/LayerManager.java
+++ b/src/main/java/org/apache/commons/collections4/bloomfilter/LayerManager.java
@@ -16,7 +16,10 @@
  */
 package org.apache.commons.collections4.bloomfilter;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.Objects;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
--- a/src/main/java/org/apache/commons/collections4/bloomfilter/LayerManager.java
+++ b/src/main/java/org/apache/commons/collections4/bloomfilter/LayerManager.java
@@ -16,7 +16,7 @@
  */
 package org.apache.commons.collections4.bloomfilter;
 
-import java.util.ArrayList;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Objects;
@@ -59,7 +59,7 @@ public class LayerManager<T extends BloomFilter> implements BloomFilterProducer 
     public static class Builder<T extends BloomFilter> {
         private Predicate<LayerManager<T>> extendCheck;
         private Supplier<T> supplier;
-        private Consumer<List<? extends BloomFilter>> cleanup;
+        private Consumer<List<T>> cleanup;
 
         private Builder() {
             extendCheck = ExtendCheck.neverAdvance();
@@ -85,7 +85,7 @@ public class LayerManager<T extends BloomFilter> implements BloomFilterProducer 
          *                dated or stale filters.
          * @return this
          */
-        public Builder<T> setCleanup(Consumer<List<? extends BloomFilter>> cleanup) {
+        public Builder<T> setCleanup(Consumer<List<T>> cleanup) {
             this.cleanup = cleanup;
             return this;
         }
@@ -125,7 +125,7 @@ public class LayerManager<T extends BloomFilter> implements BloomFilterProducer 
          * A Cleanup that never removes anything.
          * @return A Consumer suitable for the LayerManager {@code cleanup} parameter.
          */
-        public static Consumer<List<? extends BloomFilter>> noCleanup() {
+        public static  <T extends BloomFilter>  Consumer<List<T>> noCleanup() {
             return x -> {};
         }
 
@@ -138,7 +138,7 @@ public class LayerManager<T extends BloomFilter> implements BloomFilterProducer 
          * @return A Consumer suitable for the LayerManager {@code cleanup} parameter.
          * @throws IllegalArgumentException if {@code maxSize <= 0}.
          */
-        public static <T extends BloomFilter> Consumer<List<? extends BloomFilter>> onMaxSize(int maxSize) {
+        public static <T extends BloomFilter> Consumer<List<T>> onMaxSize(int maxSize) {
             if (maxSize <= 0) {
                 throw new IllegalArgumentException("'maxSize' must be greater than 0");
             }
@@ -155,12 +155,22 @@ public class LayerManager<T extends BloomFilter> implements BloomFilterProducer 
          *
          * @return A Consumer suitable for the LayerManager {@code cleanup} parameter.
          */
-        public static <T extends BloomFilter> Consumer<List<? extends BloomFilter>> removeEmptyTarget() {
+        public static <T extends BloomFilter> Consumer<List<T>> removeEmptyTarget() {
             return x -> {
                 if (!x.isEmpty() && x.get(x.size()-1).isEmpty()) {
                     x.remove(x.size()-1);
                 }
             };
+        }
+
+        /**
+         * Removes any layer identified by the predicate.
+         *
+         * @param test Predicate.
+         * @return A Consumer suitable for the LayerManager {@code cleanup} parameter.
+         */
+        public static <T extends BloomFilter> Consumer<List<T>> removeIf(Predicate<? super T> test) {
+            return x -> x.removeIf(test);
         }
 
         private Cleanup() {
@@ -246,8 +256,8 @@ public class LayerManager<T extends BloomFilter> implements BloomFilterProducer 
     public static <T extends BloomFilter> Builder<T> builder() {
         return new Builder<>();
     }
-    private final List<T> filters = new ArrayList<>();
-    private final Consumer<List<? extends BloomFilter>> filterCleanup;
+    private final LinkedList<T> filters = new LinkedList<>();
+    private final Consumer<List<T>> filterCleanup;
 
     private final Predicate<LayerManager<T>> extendCheck;
 
@@ -265,7 +275,7 @@ public class LayerManager<T extends BloomFilter> implements BloomFilterProducer 
      * @param initialize     true if the filter list should be initialized.
      */
     private LayerManager(Supplier<T> filterSupplier, Predicate<LayerManager<T>> extendCheck,
-            Consumer<List<? extends BloomFilter>> filterCleanup, boolean initialize) {
+            Consumer<List<T>> filterCleanup, boolean initialize) {
         this.filterSupplier = filterSupplier;
         this.extendCheck = extendCheck;
         this.filterCleanup = filterCleanup;
@@ -306,7 +316,7 @@ public class LayerManager<T extends BloomFilter> implements BloomFilterProducer 
     public LayerManager<T> copy() {
         LayerManager<T> newMgr = new LayerManager<>(filterSupplier, extendCheck, filterCleanup, false);
         for (T bf : filters) {
-            newMgr.filters.add((T) bf.copy());
+            newMgr.filters.add(bf.copy());
         }
         return newMgr;
     }
@@ -362,7 +372,7 @@ public class LayerManager<T extends BloomFilter> implements BloomFilterProducer 
      * @see #getTarget()
      */
     public final T first() {
-        return filters.get(0);
+        return filters.getFirst();
     }
 
     /**
@@ -372,7 +382,7 @@ public class LayerManager<T extends BloomFilter> implements BloomFilterProducer 
      * @see #getTarget()
      */
     public final T last() {
-        return filters.get(filters.size()-1);
+        return filters.getLast();
     }
 
     /**
--- a/src/main/java/org/apache/commons/collections4/bloomfilter/LayerManager.java
+++ b/src/main/java/org/apache/commons/collections4/bloomfilter/LayerManager.java
@@ -16,9 +16,7 @@
  */
 package org.apache.commons.collections4.bloomfilter;
 
-import java.util.LinkedList;
-import java.util.NoSuchElementException;
-import java.util.Objects;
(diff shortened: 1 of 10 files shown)


Claudenw:
Question: Code coverage complains about this block from the advanceOnSaturate predicate.  

```
            return manager -> {
                BloomFilter bf = manager.last();
                return bf != null && maxN <= bf.getShape().estimateN(bf.cardinality());
            };
```

The manager should never have an empty list, so `manager.last()` should always work.  This is defensive programming in case there is a bad cleanup or similar call that makes the list empty.  Should I try to contrive a test that covers this line, or should I not perform the null test and just let and NPE happen if the list is empty?

aherbert:
At present the contract of `LayerManager.last()` states that it can be null, so the null check is required. However this does not seem to be the intention of LayerManager which should always have 1 layer.

If you think that the layers should never be empty then we should enforce this in documentation and can remove empty checks in `first()` and `last()`. Looking at the code, addFilter is called whenever clean-up/clear is performed when the layers are empty. So given this is not meant to be thread-safe (e.g. one thread calls clear while another is calling last) then the intention is to never be empty. Any time the layers are empty is a bug. In this case it may be highlighted by use of a null pointer. So I would be inclined to drop the null checks throughout.

garydgregory:
COMMONS-853?

src/main/java/org/apache/commons/collections4/bloomfilter/WrappedBloomFilter.java
```diff
@@ -25,7 +25,7 @@
  * @since 4.5
  */
 public abstract class WrappedBloomFilter implements BloomFilter {
-    final BloomFilter wrapped;
+    protected final BloomFilter wrapped;
```
garydgregory, line 28: I personally do not like protected instance variables, it just handcuffs us IMO. I would prefer access through a method and keep the ivar private or package-private.

src/test/java/org/apache/commons/collections4/bloomfilter/BitMapProducerFromWrappedBloomFilterTest.java
```diff
@@ -23,13 +23,25 @@ public class BitMapProducerFromWrappedBloomFilterTest extends AbstractBitMapProd
     @Override
     protected BitMapProducer createEmptyProducer() {
         return new WrappedBloomFilter(new DefaultBloomFilterTest.SparseDefaultBloomFilter(shape)) {
+            @Override
```
garydgregory, line 26: I'd like to review the API design of the `BloomFilter#merge()` method. It seems that (after a quick look) it always returns `true`. What's a user to do with a `false` result? Would throwing an exception be better?
If that's the case I think it _might_ be nicer to return `this` from a merge which would allow to use the merge as a fluent API call. For example, this copy method would become 
`return new DefaultBloomFilterTest.SparseDefaultBloomFilter(shape).merge(wrapped);`

aherbert: The `boolean` return was added to support the `CountingBloomFilter`. That filter can be invalidated by a merge operation. If we remove the `boolean` return, then a separate call is required for each merge to check that the filter is still valid.

In 99.9999% of cases this will be true as the current implementation can count up to `Integer.MAX_VALUE`. A more space efficient counting Bloom filter would use a smaller primitive type to store the counts such as a `byte`. This would only be able to count up to 127 merge operations at the same cell index while retaining the sign bit to check for overflow and invalidation of the filter.

This does require a bit of consideration. The `CountingBloomFilter` can **add** and **remove** Bloom filters. However it should also have use as a regular Bloom filter and be able to check for presence of an object with a given false-positive rate. So a future improvement (not yet implemented) is to consider the `Shape` when constructing the filter and use the most efficient storage.

For example, if a filter is created that will accept up to 30000 items before saturation (and there is a decline in the false positive rate beyond the target value), then the counts could be maintained using a `short`.

Note that I am not opposed to an API change here. The `CountingBloomFilter` is a specialised case and usage of it requires some knowledge of its limitations. So any API change to make the `BloomFilter` more fluent should consider how to document that the merge can fail in some cases.

garydgregory: Ty for the explanation. We can leave it as is then. If there is any way to improve the Javadoc to help non-bloom filter experts, then that would be nice.

Claudenw:
> COMMONS-853?

I don't find this.  what is it?

aherbert:
> > COMMONS-853?
> 
> I don't find this. what is it?

You originally named the PR COMMONS-853. I have updated to the title to COLLECTIONS-853.

garydgregory (approved):
LGTM. @aherbert ?

Claudenw:
@aherbert any requested changes for this pull?

aherbert (commented):
The choice of a LinkedList or ArrayList as the backing implementation for the layers requires some consideration. If a user will perform many removals of the head of the list then the ArrayList will be slow due to copy of the entire remaining list for each removal. It will also be slow for a clean-up filter that removes a large proportion of the list on a single pass. The array is faster for layer retrieval by index.

One possible solution would be to offer the choice via the builder and then create the list in the LayerManager constructor.

Note: The bulk removal speed over an ArrayList can be somewhat mitigated if we change the clean-up API to support clean-up operations that can be implemented using a `Predicate<BloomFilter>`. For the predicates we can use `List.removeIf(Predicate)` from the Collections API which is efficient single pass implementation for an ArrayList:

```java
/**
 * Removes any layer identified by the predicate.
 *
 * @param test Predicate.
 * @return A Consumer suitable for the LayerManager {@code cleanup} parameter.
 */
public static <T extends BloomFilter> Consumer<List<T>> removeIf(Predicate<? super T> test) {
    return x -> x.removeIf(test);
}
```

src/main/java/org/apache/commons/collections4/bloomfilter/LayerManager.java
```diff
@@ -50,15 +51,15 @@
...
      * Builder to create Layer Manager
      */
-    public static class Builder {
-        private Predicate<LayerManager> extendCheck;
-        private Supplier<BloomFilter> supplier;
-        private Consumer<LinkedList<BloomFilter>> cleanup;
+    public static class Builder<T extends BloomFilter> {
+        private Predicate<LayerManager<T>> extendCheck;
+        private Supplier<T> supplier;
+        private Consumer<List<? extends BloomFilter>> cleanup;
```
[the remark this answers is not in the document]
Claudenw:
```
    public static <T extends BloomFilter> LayeredBloomFilter<T> fixed(final Shape shape, int maxDepth, Supplier<T> supplier) {
        LayerManager.Builder<T> builder = LayerManager.builder();
        LayerManager<T> manager =  builder.setExtendCheck(LayerManager.ExtendCheck.advanceOnPopulated())
                .setCleanup(LayerManager.Cleanup.onMaxSize(maxDepth)).setSupplier(supplier).build();
        return new LayeredBloomFilter<T>(shape, manager);
    }
```
@aherbert , this is the only way I can see to implement the static method.  Do you have a solution that does not require the ` LayerManager.Builder<T> builder = LayerManager.builder();` line?

Claudenw:
@garydgregory @aherbert Any reason not to merge this?

garydgregory:
I just want to make sure @aherbert approves before I merge.

src/main/java/org/apache/commons/collections4/bloomfilter/LayeredBloomFilter.java
```diff
@@ -89,19 +90,35 @@ public boolean test(BloomFilter x) {
...
+    /**
+     * Creates a fixed size layered bloom filter that adds new filters to the list,
+     * but never merges them. List will never exceed maxDepth. As additional filters
+     * are added earlier filters are removed.  Uses SimpleBloomFilters.
+     *
+     * @param shape    The shape for the enclosed Bloom filters.
+     * @param maxDepth The maximum depth of layers.
+     * @return An empty layered Bloom filter of the specified shape and depth.
+     */
+    public static  LayeredBloomFilter<BloomFilter> fixed(final Shape shape, int maxDepth) {
```
aherbert, line 103: Q. Why are we adding this factory method? It seems strange to have one factory method creating layers to a fixed depth with no merges when there are countless other combinations that can be used to create the layers. Is this the most common usage of a layered Bloom filter, e.g. described in Zhiwang et al?

aherbert (commented):
This looks good. I have a raised a question about the factory method. Why not let the user do this as it is only one of many ways to create the filter? If it is by far the most common usage then this makes sense, but if it is one of many then it seems odd to support only one.

src/main/java/org/apache/commons/collections4/bloomfilter/LayeredBloomFilter.java
Claudenw: Actually, this method was in the original set, I extended it to allow different implementations of Bloom filters on the layers (see next method).  We could probably drop this one with no issues.

garydgregory: Do we foresee the need to add more constructors in the future? I wonder if we should consider having a private constructor and only factory methods, or a builder.

Claudenw: @aherbert I removed the state factory method(s) from LayeredBloomFilter.  I had to move them to the testing classes, but they are no longer part of the public API.  Can you please review and approve?

aherbert (changes requested):
Changes to generics look good.

I have suggested a different method to bulk remove old filter layers since we can no longer use the LinkedList removeFirst method. Use of a subList will allow the most efficient bulk removal. But IIUC the use-case should be for single layer removal. There seems to be no way to add multiple layers without having to do cleanup for each added layer, thus a bulk-removal based on size should not be required. If you agree with that assessment then there is only a formatting change to add. The onMaxSize lambda can be commented:
```java
return ll -> {
    // Expected use-case is for a single removal to make space for a new layer
    while (ll.size() > maxSize) {
        ll.remove(0);
    }
};
```

src/main/java/org/apache/commons/collections4/bloomfilter/LayerManager.java
```diff
@@ -137,13 +138,13 @@ public static Consumer<LinkedList<BloomFilter>> noCleanup() {
...
          */
-        public static Consumer<LinkedList<BloomFilter>> onMaxSize(int maxSize) {
+        public static <T extends BloomFilter> Consumer<List<T>> onMaxSize(int maxSize) {
             if (maxSize <= 0) {
                 throw new IllegalArgumentException("'maxSize' must be greater than 0");
             }
             return ll -> {
                 while (ll.size() > maxSize) {
-                    ll.removeFirst();
+                    ll.remove(0);
```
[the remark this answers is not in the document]
Claudenw: :man_facepalming: @aherbert I finally understand what you were saying.

I changed the `LayerManager` to use a `LinkedList` internally, but changed the Cleanup to be defined as `Consumer<Deque<T>>`

I think this solves most of the problems.  If someone wants to do something weird in the Cleanup they can use the `Collection` or `Iterable` interfaces to navigate.

Hopefully the last push solves the issues and we can get this merged.

src/main/java/org/apache/commons/collections4/bloomfilter/LayerManager.java
```diff
@@ -179,15 +190,15 @@ public static final class ExtendCheck {
...
                 throw new IllegalArgumentException("'breakAt' must be greater than 0");
             }
-            return new Predicate<LayerManager>() {
+            return new Predicate<LayerManager<T>>() {
                 int count;
 
                 @Override
-                public boolean test(LayerManager filter) {
+                public boolean test(LayerManager<T> filter) {
                     return ++count % breakAt == 0;
```
aherbert, line 202: Just noticed this use of modulus. It will be incorrect when the counter overflows. In the following example using mod 3 an extra call is required to reach modulus zero; with 7 it takes 3 calls too few:
```
jshell> int i = Integer.MAX_VALUE
i ==> 2147483647

jshell> i % 3
$2 ==> 1       // Should be 2 calls from here

jshell> (i+1) % 3
$3 ==> -2

jshell> (i+2) % 3
$4 ==> -1

jshell> (i+3) % 3
$5 ==> 0


jshell> i % 7
$10 ==> 1       // Should be 6 calls from here

jshell> (i+1) % 7
$11 ==> -2

jshell> (i+2) % 7
$12 ==> -1

jshell> (i+3) % 7
$13 ==> 0
```

It can be fixed using:
```java
public boolean test(LayerManager<T> filter) {
    if (++count == breakAt) {
        count = 0;
        return true;
    }
    return false;
}
```

The use of a branch statement _may_ be offset by avoiding a modulus division.

Q. Is it realistic to call the extend check 2 billion times? Will anyone notice if the extend check is incorrect after 2 billion calls?

Still, I would prefer to not include a known bug in the release however trivial the consequences of it may be.

aherbert (approved):
Changes to use the Deque are fine. Feel free to merge.

I added a note about the common pattern of using `count % size == 0` to do something at set intervals. It has a bug when using signed integers that is not relevant when unsigned integers are available. Have a think if this is something worth fixing.

If all other code around the use case ensures that 2 billion calls is not possible then this bug never manifests. For example if reporting progress at set intervals when the total progress size is a positive integer. In this case the filter is used by downstream code and we cannot be _absolutely sure_ that usage will not reach overflow.

src/main/java/org/apache/commons/collections4/bloomfilter/LayerManager.java
Claudenw: Should be fixed now.

PR #485, opened 2024-05-02
COLLECTIONS-842: deprecate Lists incompatible with Java 21, add new variant of AbstractLinkedList

reschke:
This is a minimal PR. We may need to improve the deprecation notes.

Coverage tests are failing; we'd need to add test coverage for the new variant, which IMHO requires having an implementation using it.

I've verified that this change would work for Jackrabbit, see https://issues.apache.org/jira/browse/JCR-5049 for details.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/list/CursorableLinkedList.java
+++ b/src/main/java/org/apache/commons/collections4/list/CursorableLinkedList.java
@@ -57,7 +57,7 @@ import java.util.ListIterator;
  *
  * @see java.util.LinkedList
  * @since 1.0
- * @deprecated
+ * @deprecated parent {@link AbstractLinkedList} is source incompatible with List methods added in Java 21
  */
 @Deprecated
 public class CursorableLinkedList<E> extends AbstractLinkedList<E> implements Serializable {
--- a/src/main/java/org/apache/commons/collections4/list/CursorableLinkedList.java
+++ b/src/main/java/org/apache/commons/collections4/list/CursorableLinkedList.java
@@ -57,6 +57,7 @@ import java.util.ListIterator;
  *
  * @see java.util.LinkedList
  * @since 1.0
+ * @deprecated
  */
 public class CursorableLinkedList<E> extends AbstractLinkedList<E> implements Serializable {
 
--- a/src/main/java/org/apache/commons/collections4/list/CursorableLinkedList.java
+++ b/src/main/java/org/apache/commons/collections4/list/CursorableLinkedList.java
@@ -59,6 +59,7 @@ import java.util.ListIterator;
  * @since 1.0
  * @deprecated
  */
+@Deprecated
 public class CursorableLinkedList<E> extends AbstractLinkedList<E> implements Serializable {
 
     /**
--- a/src/main/java/org/apache/commons/collections4/list/NodeCachingLinkedList.java
+++ b/src/main/java/org/apache/commons/collections4/list/NodeCachingLinkedList.java
@@ -40,7 +40,9 @@ import java.util.Collection;
  * </p>
  *
  * @since 3.0
+ * @deprecated
  */
+@Deprecated
 public class NodeCachingLinkedList<E> extends AbstractLinkedList<E> implements Serializable {
 
     /** Serialization version */
--- a/src/main/java/org/apache/commons/collections4/list/NodeCachingLinkedList.java
+++ b/src/main/java/org/apache/commons/collections4/list/NodeCachingLinkedList.java
@@ -40,7 +40,7 @@ import java.util.Collection;
  * </p>
  *
  * @since 3.0
- * @deprecated
+ * @deprecated parent {@link AbstractLinkedList} is source incompatible with List methods added in Java 21
  */
 @Deprecated
 public class NodeCachingLinkedList<E> extends AbstractLinkedList<E> implements Serializable {
--- a/src/test/java/org/apache/commons/collections4/list/AbstractLinkedListForJava21Test.java
+++ b/src/test/java/org/apache/commons/collections4/list/AbstractLinkedListForJava21Test.java
@@ -1,0 +1,216 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.collections4.list;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.util.Arrays;
+import java.util.List;
+
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test case for {@link AbstractLinkedListForJava21}.
+ */
+public abstract class AbstractLinkedListForJava21Test<E> extends AbstractListTest<E> {
+
+    public AbstractLinkedListForJava21Test() {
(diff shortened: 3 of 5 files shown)


aherbert:
I duplicated AbstractLinkedListTest with some modifications:
```java
class AbstractLinkedListForJava21Test<E> extends AbstractListTest<E> {

    AbstractLinkedListForJava21Test() {
        super(AbstractLinkedListForJava21Test.class.getSimpleName());
    }

    // ...

    @Override
    public String getCompatibilityVersion() {
        return null;
    }

    @Override
    protected boolean skipSerializedCanonicalTests() {
        return true;
    }

    @Override
    public AbstractLinkedListForJava21<E> getCollection() {
        return (AbstractLinkedListForJava21<E>) super.getCollection();
    }

    @Override
    public List<E> makeObject() {
        return new DefaultAbstractLinkedListForJava21<>();
    }

    private class DefaultAbstractLinkedListForJava21<E> extends AbstractLinkedListForJava21<E> {
        DefaultAbstractLinkedListForJava21() {
            init();
        }

        private void readObject(final ObjectInputStream in) throws IOException, ClassNotFoundException {
            in.defaultReadObject();
            doReadObject(in);
        }

        private void writeObject(final ObjectOutputStream out) throws IOException {
            out.defaultWriteObject();
            doWriteObject(out);
        }
    }
}
```

Coverage of the new linked list class is 94%. There is no coverage of sublist iterators or creating from a Collection. These may have to be added from other tests of the old AbstractLinkedList, or elsewhere.

I have skipped the canonical serialization test. For this release there is no backward compatibility to maintain.

It is a start towards hitting the coverage thresholds.

reschke:
@aherbert - excellent; do you want to push to my repo, or should we move over to the commons-collections git repo?

aherbert:
Just add this mod to your repo and it will be collected here for the PR. Another maintainer may wish to pass an eye over it.

src/test/java/org/apache/commons/collections4/list/AbstractLinkedListForJava21Test.java
```diff
@@ -0,0 +1,216 @@
...
+    public AbstractLinkedListForJava21<E> getCollection() {
+        return (AbstractLinkedListForJava21<E>) super.getCollection();
+    }
+
+    @Override
+    public List<E> makeObject() {
+        return new DefaultAbstractLinkedListForJava21<>();
+    }
+
+    private class DefaultAbstractLinkedListForJava21<E> extends AbstractLinkedListForJava21<E> {
```
aherbert, line 201: This class should be static. Sorry for my cut and paste error. Otherwise the serialization test triggers serialization of the enclosing class AbstractLinkedListForJava21Test (which is not serializable).

src/test/java/org/apache/commons/collections4/list/AbstractLinkedListForJava21Test.java
```diff
@@ -0,0 +1,216 @@
...
+import java.io.ObjectOutputStream;
+import java.util.Arrays;
+import java.util.List;
+
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test case for {@link AbstractLinkedListForJava21}.
+ */
+public abstract class AbstractLinkedListForJava21Test<E> extends AbstractListTest<E> {
```
aherbert, line 33: This test class should not be abstract. JUnit will not run any tests.

aherbert (commented):
@reschke This test class needs to be concrete. JUnit will not run tests on an abstract class.

Can you update the `DefaultAbstractLinkedListForJava21` to be a static inner class. This should fix the serialization error when the tests actually run.

src/test/java/org/apache/commons/collections4/list/AbstractLinkedListForJava21Test.java
reschke: Should I rename as well?

reschke:
Ths helped, but still fails with:
~~~
[WARNING] Rule violated for bundle commons-collections4: instructions covered ratio is 0.84, but expected minimum is 0.85
~~~

aherbert:
Given that a lot of dynamically generated Junit4 tests were lost in the conversion to JUnit5, we may have to write some more tests just to get this into a working build. My initial investigation into fixing the lost tests discovered only that it is a larger task which is out-of-scope here.

Try pasting this into the current test. These are copied from `CursorableLinkedListTest`:
```java
    @Test
    @SuppressWarnings("unchecked")
    public void testSubList() {
        List<E> list = makeObject();
        list.add((E) "A");
        list.add((E) "B");
        list.add((E) "C");
        list.add((E) "D");
        list.add((E) "E");

        assertEquals("[A, B, C, D, E]", list.toString());
        assertEquals("[A, B, C, D, E]", list.subList(0, 5).toString());
        assertEquals("[B, C, D, E]", list.subList(1, 5).toString());
        assertEquals("[C, D, E]", list.subList(2, 5).toString());
        assertEquals("[D, E]", list.subList(3, 5).toString());
        assertEquals("[E]", list.subList(4, 5).toString());
        assertEquals("[]", list.subList(5, 5).toString());
    }

    @Test
    @SuppressWarnings("unchecked")
    public void testSubListAddBegin() {
        List<E> list = makeObject();
        list.add((E) "A");
        list.add((E) "B");
        list.add((E) "C");
        list.add((E) "D");
        list.add((E) "E");

        final List<E> sublist = list.subList(0, 0);
        sublist.add((E) "a");
        assertEquals("[a, A, B, C, D, E]", list.toString());
        assertEquals("[a]", sublist.toString());
        sublist.add((E) "b");
        assertEquals("[a, b, A, B, C, D, E]", list.toString());
        assertEquals("[a, b]", sublist.toString());
    }

    @Test
    @SuppressWarnings("unchecked")
    public void testSubListAddEnd() {
        List<E> list = makeObject();
        list.add((E) "A");
        list.add((E) "B");
        list.add((E) "C");
        list.add((E) "D");
        list.add((E) "E");

        final List<E> sublist = list.subList(5, 5);
        sublist.add((E) "F");
        assertEquals("[A, B, C, D, E, F]", list.toString());
        assertEquals("[F]", sublist.toString());
        sublist.add((E) "G");
        assertEquals("[A, B, C, D, E, F, G]", list.toString());
        assertEquals("[F, G]", sublist.toString());
    }

    @Test
    @SuppressWarnings("unchecked")
    public void testSubListAddMiddle() {
        List<E> list = makeObject();
        list.add((E) "A");
        list.add((E) "B");
        list.add((E) "C");
        list.add((E) "D");
        list.add((E) "E");

        final List<E> sublist = list.subList(1, 3);
        sublist.add((E) "a");
        assertEquals("[A, B, C, a, D, E]", list.toString());
        assertEquals("[B, C, a]", sublist.toString());
        sublist.add((E) "b");
        assertEquals("[A, B, C, a, b, D, E]", list.toString());
        assertEquals("[B, C, a, b]", sublist.toString());
    }

    @Test
    @SuppressWarnings("unchecked")
    public void testSubListRemove() {
        List<E> list = makeObject();
        list.add((E) "A");
        list.add((E) "B");
        list.add((E) "C");
        list.add((E) "D");
        list.add((E) "E");

        final List<E> sublist = list.subList(1, 4);
        assertEquals("[B, C, D]", sublist.toString());
        assertEquals("[A, B, C, D, E]", list.toString());
        sublist.remove("C");
        assertEquals("[B, D]", sublist.toString());
        assertEquals("[A, B, D, E]", list.toString());
        sublist.remove(1);
        assertEquals("[B]", sublist.toString());
        assertEquals("[A, B, E]", list.toString());
        sublist.clear();
        assertEquals("[]", sublist.toString());
        assertEquals("[A, E]", list.toString());
    }
```

When I added this locally the coverage is above 85% which may be enough to get green lights on the build.

reschke:
@aherbert - thanks for helping out!

src/main/java/org/apache/commons/collections4/list/CursorableLinkedList.java
```diff
@@ -57,7 +57,9 @@
  *
  * @see java.util.LinkedList
  * @since 1.0
+ * @deprecated
```
aherbert, line 60: All the other deprecated tags in the library have a comment on what to use instead. We currently do not have an alternative but we should add something useful here to note why it is deprecated:
```java
@deprecated parent {@link AbstractLinkedList} is source incompatible with List methods added in Java 21
```
If a request comes to add a replacement then it can be done under a new enhancement.

src/main/java/org/apache/commons/collections4/list/NodeCachingLinkedList.java
```diff
@@ -40,7 +40,9 @@
  * </p>
  *
  * @since 3.0
+ * @deprecated
```
aherbert, line 43: Same comment here

aherbert (changes requested):
The build is now fixed which is enough to allow a merge. Can you add a simple note to the deprecated tags.

PR #492, opened 2024-05-17
COLLECTIONS-854 : renaming bloomfilter components and methods as per ticket

Claudenw:
Fixes the issues listed in COLLECTIONS-854 and as agreed in the dev discussions.

Only names and documentation changes are applied here with the following exceptions as noted in the 854 ticket.

- Changed `forEachRemaining()` in `CountingLongPredicate` and `CountingPredicate` classes to `processRemaining()` to keep in spirit with Extractor usage.
- Added missing `FunctionalInterface` annotation to `BloomFilterExtractor`.


src/main/java/org/apache/commons/collections4/bloomfilter/CountingBloomFilter.java
```diff
@@ -172,21 +172,21 @@ default int getMaxInsert(final IndexProducer idxProducer) {
...
      * @see #isValid()
-     * @see #add(CellProducer)
+     * @see #add(CellExtractor)
      */
     @Override
-    default boolean merge(final BitMapProducer bitMapProducer) {
-        Objects.requireNonNull(bitMapProducer, "bitMapProducer");
-        return merge(IndexProducer.fromBitMapProducer(bitMapProducer));
+    default boolean merge(final BitMapExtractor bitMapExtractor) {
+        Objects.requireNonNull(bitMapExtractor, "bitMapExtractor");
```
garydgregory, line 188: @Claudenw 
Since `bitMapExtractor` is not dereferenced here, I don't think it makes sense to check for nulls here. Instead, push the check down to implementations. Some implementations may choose to turn the null into a no-op.

src/main/java/org/apache/commons/collections4/bloomfilter/CountingBloomFilter.java
```diff
@@ -229,47 +229,47 @@ default boolean merge(final Hasher hasher) {
...
     @Override
-    default boolean merge(final IndexProducer indexProducer) {
-        Objects.requireNonNull(indexProducer, "indexProducer");
+    default boolean merge(final IndexExtractor indexExtractor) {
+        Objects.requireNonNull(indexExtractor, "indexExtractor");
         try {
-            return add(CellProducer.from(indexProducer.uniqueIndices()));
+            return add(CellExtractor.from(indexExtractor.uniqueIndices()));
         } catch (final IndexOutOfBoundsException e) {
             throw new IllegalArgumentException(
```
garydgregory, line 252: This assumes a lot of the implementations. I think it is OK to just Javadoc the `IndexOutOfBoundsException` in the implementation, or, have the implementations convert whatever error condition into an `IllegalArgumentException `. There Javadoc _here_ can state that it is expected that an implementation throws `IllegalArgumentException` and not other exceptions on bad input.

src/main/java/org/apache/commons/collections4/bloomfilter/CountingBloomFilter.java
```diff
@@ -229,47 +229,47 @@ default boolean merge(final Hasher hasher) {
...
      * @return {@code true} if the removal was successful and the state is valid
      * @see #isValid()
-     * @see #subtract(CellProducer)
+     * @see #subtract(CellExtractor)
      */
-    default boolean remove(final BitMapProducer bitMapProducer) {
-        Objects.requireNonNull(bitMapProducer, "bitMapProducer");
-        return remove(IndexProducer.fromBitMapProducer(bitMapProducer));
+    default boolean remove(final BitMapExtractor bitMapExtractor) {
+        Objects.requireNonNull(bitMapExtractor, "bitMapExtractor");
```
garydgregory, line 271: See above comment.

src/main/java/org/apache/commons/collections4/bloomfilter/CountingBloomFilter.java
```diff
@@ -303,33 +303,33 @@ default boolean remove(final BloomFilter other) {
...
-     * @see #subtract(CellProducer)
+     * @see #subtract(CellExtractor)
      */
-    default boolean remove(final IndexProducer indexProducer) {
-        Objects.requireNonNull(indexProducer, "indexProducer");
+    default boolean remove(final IndexExtractor indexExtractor) {
+        Objects.requireNonNull(indexExtractor, "indexExtractor");
         try {
-            return subtract(CellProducer.from(indexProducer.uniqueIndices()));
+            return subtract(CellExtractor.from(indexExtractor.uniqueIndices()));
```
garydgregory, line 332: See above comment.

src/main/java/org/apache/commons/collections4/bloomfilter/CountingBloomFilter.java
```diff
@@ -338,23 +338,23 @@ default boolean remove(final IndexProducer indexProducer) {
...
      * @see #isValid()
-     * @see #add(CellProducer)
+     * @see #add(CellExtractor)
      */
-    boolean subtract(CellProducer other);
+    boolean subtract(CellExtractor other);
 
     @Override
-    default IndexProducer uniqueIndices() {
+    default IndexExtractor uniqueIndices() {
```
garydgregory, line 357: Shouldn't this method have a Javadoc that covers this implementation as a no-op?

src/main/java/org/apache/commons/collections4/bloomfilter/IndexExtractor.java
```diff
@@ -31,15 +31,15 @@
  * @since 4.5
```
garydgregory, line 31: Note that the since tags should state `4.5.0` but I'll do a pass before the next release, so you don't have to make this PR bigger ;-)

garydgregory (changes requested):
@Claudenw 
Thank you for your PR!
Please see my comments.

src/main/java/org/apache/commons/collections4/bloomfilter/BitMapExtractor.java
```diff
@@ -61,33 +61,33 @@ public boolean forEachBitMap(final LongPredicate predicate) {
...
+    static BitMapExtractor fromIndexExtractor(final IndexExtractor extractor, final int numberOfBits) {
+        Objects.requireNonNull(extractor, "extractor");
         Objects.requireNonNull(numberOfBits, "numberOfBits");
 
-        final long[] result = new long[BitMap.numberOfBitMaps(numberOfBits)];
-        producer.forEachIndex(i -> {
-            BitMap.set(result, i);
+        final long[] result = new long[BitMaps.numberOfBitMaps(numberOfBits)];
+        extractor.processIndices(i -> {
+            BitMaps.set(result, i);
```
garydgregory, line 83: This is awkward (the hardcoded `return true` part). How about adding a more standard `Consumer` API?

garydgregory: Also (unrelated), would it ever make sense for `BitMaps.set()` to return a boolean for the result of its `OR` operation? I'm not sure anyone would use that.

garydgregory:
@Claudenw 
I think the internal `Bits` class in https://github.com/apache/commons-collections/blob/9b060d0050483250f7252e3ec13dda24003e8b09/src/main/java/org/apache/commons/collections4/bloomfilter/BitMapExtractor.java#L98 can `final`.

src/main/java/org/apache/commons/collections4/bloomfilter/BitMapExtractor.java
aherbert: To return a result would require reading the bit before setting it to 1. This is an unwanted overhead in the common case where you set the bit and don't care if it was already set.

Possibly better served in the future by:
```java
    public static boolean getAndSet(final long[] bitMaps, final int bitIndex) {
        // read value, set value as true, return previous value (true/false)
    }
```

src/main/java/org/apache/commons/collections4/bloomfilter/IndexExtractor.java
Claudenw: I don't think this PR can get any bigger ;)

src/main/java/org/apache/commons/collections4/bloomfilter/BitMapExtractor.java
Claudenw: Gary,  I am happy to discuss any changes you would like to see in the codebase.

However this change was simply to effect the change in nomenclature as discussed on the mailing list.  I would like to restrict this to just the nomenclature changes without any functional changes.

garydgregory: OK, sounds good, please make sure to follow up with another PR then :-)

garydgregory (approved):
LGTM

garydgregory:
@aherbert OK to merge?

src/main/java/org/apache/commons/collections4/bloomfilter/package-info.java
```diff
@@ -41,30 +41,60 @@
...
- * Programs that utilize the Bloom filters may use the {@code BitMapProducer} or {@code IndexProducer} to retrieve a
- * representation of the internal structure. Additional methods are available in the {@code BitMap} to assist in
- * manipulation of the representations.</p>
+ * <h4>Nomenclature</h4>
  *
- * <p>The Bloom filter code is an interface that requires implementation of 9 methods:</p>
+ * <ul>
+ *     <li>BitMap - In the bloomfilter package a BitMap is not a structure but a logical construct.  It is conceptualized
+ *     as an ordered collection of {@code long} values each of which is interpreted as a 64-bit bit vector.  The mapping of
+ *     bits into the {@code long} values ss described in the the {@code BitMaps} javadoc.</li>
```
aherbert, line 49: ss -> is

src/main/java/org/apache/commons/collections4/bloomfilter/package-info.java
```diff
@@ -41,30 +41,60 @@
...
- * <p>In addition the architecture is designed so that the implementation of the storage of bits is abstracted.
- * Programs that utilize the Bloom filters may use the {@code BitMapProducer} or {@code IndexProducer} to retrieve a
- * representation of the internal structure. Additional methods are available in the {@code BitMap} to assist in
- * manipulation of the representations.</p>
+ * <h4>Nomenclature</h4>
  *
- * <p>The Bloom filter code is an interface that requires implementation of 9 methods:</p>
+ * <ul>
+ *     <li>BitMap - In the bloomfilter package a BitMap is not a structure but a logical construct.  It is conceptualized
+ *     as an ordered collection of {@code long} values each of which is interpreted as a 64-bit bit vector.  The mapping of
```
aherbert, line 48: as the enabled true/false state of 64 continuous indices

src/main/java/org/apache/commons/collections4/bloomfilter/package-info.java
```diff
@@ -41,30 +41,60 @@
...
+ *     bits into the {@code long} values ss described in the the {@code BitMaps} javadoc.</li>
+ *
+ *     <li>Index - In the bloomfilter package an Index is a logical collection of {@code int}s specifying the enabled
+ *     bits in the BitMap.</li>
+ *
+ *     <li>Cell - Some Bloom filters (e.g. CountingBloomFilter) use counters rather than bits.  In the bloomfilter package
+ *     Cells are pairs of ints representing an index and a value.  They are not {@code Pair} objects.  </li>
```
aherbert, line 55: I do not see why we require mentioning Pair. A Cell could be a Pair<Integer, Integer>. But it could be a long packed as upper and lower 32-bits, or some other structure. Dropping the sentence does no harm IMO.

src/main/java/org/apache/commons/collections4/bloomfilter/package-info.java
```diff
@@ -41,30 +41,60 @@
...
+ *
+ *     <li>Extractor - The Extractors are {@code FunctionalInterfaces} that are conceptually iterators on a {@code BitMap}, an {@code Index}, or a
+ *     collection of {@code Cell}s, with an early termination switch.  Extractors have
+ *     names like {@code BitMapExtractor} or {@code IndexExtractor} and  have a {@code processXs} methods that take a
```
aherbert, line 59: double space between `and  have`

src/main/java/org/apache/commons/collections4/bloomfilter/package-info.java
```diff
@@ -41,30 +41,60 @@
...
+ *     {@code Predicate<X>} argument (e.g. {@code processBitMaps(LongPredicate)} or {@code processIndicies(IntPredicate)}).
+ *     That predicate is expected to process each of the Xs in turn and return {@code true} if the processing should continue
+ *     or {@code false} to stop it. </li>
+ * </ul>
+ *
+ * <p>There is an obvious association between the BitMap and the Index in that if bit 5 is enabled in the BitMap than the index must contain the index 5.</p>
```
aherbert, line 65: the BitMap and a collection of indices   ...   then the indices must contain

src/main/java/org/apache/commons/collections4/bloomfilter/package-info.java
```diff
@@ -41,30 +41,60 @@
...
+ *
+ *
+ * <h4>Implementation Notes</h4>
+ *
+ * <p>The architecture is designed so that the implementation of the storage of bits is abstracted. Rather than specifying a
+ * specific state representation we require that all Bloom filters implement the BitMapExtractor and IndexExtractor interfaces,
+ * Counting based Bloom filters implement {@code CellExtractor} as well.  There are static
```
aherbert, line 72: Counting-based

src/main/java/org/apache/commons/collections4/bloomfilter/package-info.java
```diff
@@ -83,7 +113,8 @@
  *
  * <h3>Shape</h3>
  *
- * <p>The Shape describes the Bloom filter using the number of bits and the number of hash functions</p>
+ * <p>The Shape describes the Bloom filter using the number of bits and the number of hash functions.  It can be specified
+ * by the number of exptected items and desired false positive rate.</p>
```
aherbert, line 117: expected

aherbert (changes requested):
Refactoring changes are good.

I made some suggestions for the extra details added to package info.

src/main/java/org/apache/commons/collections4/bloomfilter/package-info.java
Claudenw: I mentioned Pair because the English phrase for two of something is a pair so it is a pair of ints, but not the Java Pair.  Since the reason for the renaming was to bring the nomenclature into line with Java thinking it makes sense to call out where it differs.

src/main/java/org/apache/commons/collections4/bloomfilter/package-info.java
Claudenw: I changed the sentence to read: `There is an obvious association between the BitMap and the Index in that if bit 5 is enabled in the BitMap than the Index must contain the value 5.`

Claudenw:
Except where noted all changes requested by @aherbert have been made.

src/main/java/org/apache/commons/collections4/bloomfilter/package-info.java
aherbert: I was considering Index to be single, i.e. an Index. Index refers to a single bit.

Your sentence is treating Index as if it is pleural and can contain many bits.

We use the term indices to refer to several index bits. So this sentence is out of context of the rest of the code which uses indices and index.

garydgregory:
@aherbert ?

aherbert:
I made a comment on the new package javadoc that had no reply from @Claudenw. There is some mismatch between discussing Index and index and indices, and then BitMap (the entire set of possible indices).

See https://github.com/apache/commons-collections/pull/492#discussion_r1608148457

garydgregory:
@Claudenw ?

garydgregory:
Ping @Claudenw ;-)

Claudenw:
Sorry been buried.  I'll try to get back to this shortly.

src/main/java/org/apache/commons/collections4/bloomfilter/CountingBloomFilter.java
Claudenw: removed

src/main/java/org/apache/commons/collections4/bloomfilter/CountingBloomFilter.java
Claudenw: removed unnecessary `Objects.requireNonNull` calls

src/main/java/org/apache/commons/collections4/bloomfilter/CountingBloomFilter.java
Claudenw: Updated javadoc to mention the `IllegalArgumentException` requirement.

src/main/java/org/apache/commons/collections4/bloomfilter/package-info.java
Claudenw: fixed

src/main/java/org/apache/commons/collections4/bloomfilter/package-info.java
Claudenw: fixed

src/main/java/org/apache/commons/collections4/bloomfilter/package-info.java
Claudenw: fixed

src/main/java/org/apache/commons/collections4/bloomfilter/package-info.java
Claudenw: fixed

Claudenw:
the Java 20 failure reads:
```
org.apache.commons.collections4.IteratorUtilsTest.testGetIterator(IteratorUtilsTest.java:781)
```

So it appears to be in a different package.  I think this package can be merged. 
Any objections from @garydgregory or @aherbert

src/main/java/org/apache/commons/collections4/bloomfilter/CountingBloomFilter.java
```diff
@@ -235,9 +234,12 @@ default boolean merge(final Hasher hasher) {
...
      * <p>This method will return {@code true} if the filter is valid after the operation.</p>
      *
-     * <p>Note: If indices that are returned multiple times should be incremented multiple times convert the IndexExtractor
-     * to a CellExtractor and add that.</p>
-     *
+     * <p>Notes:</p>
+     * <ul>
+     * <li>If indices that are returned multiple times should be incremented multiple times convert the IndexExtractor
+     * to a CellExtractor and add that.</li>
+     * <li>Implementations should throw {@code IllegalArgumentException} on and no other exception on bad input.</li>
```
aherbert, line 241: Is there an extra `on` here after the `IllegalArgumentException`?

src/main/java/org/apache/commons/collections4/bloomfilter/package-info.java
```diff
@@ -56,20 +56,21 @@
...
  *     collection of {@code Cell}s, with an early termination switch.  Extractors have
- *     names like {@code BitMapExtractor} or {@code IndexExtractor} and  have a {@code processXs} methods that take a
+ *     names like {@code BitMapExtractor} or {@code IndexExtractor} and have a {@code processXs} methods that take a
  *     {@code Predicate<X>} argument (e.g. {@code processBitMaps(LongPredicate)} or {@code processIndicies(IntPredicate)}).
  *     That predicate is expected to process each of the Xs in turn and return {@code true} if the processing should continue
  *     or {@code false} to stop it. </li>
  * </ul>
  *
- * <p>There is an obvious association between the BitMap and the Index in that if bit 5 is enabled in the BitMap than the index must contain the index 5.</p>
+ * <p>There is an obvious association between the BitMap and the Index in that if bit 5 is enabled in the BitMap than the indices must contain the
```
aherbert, line 65: This still does not address what you are referring to with the noun `Index`. We do not have an `Index` object and do not refer to this elsewhere in the code. Can we replace `Index` with `indices`?

aherbert (changes requested):
A few more queries to address

src/main/java/org/apache/commons/collections4/bloomfilter/package-info.java
Claudenw: I think I managed to clarify it.  These are in the Nomenclature section and are referring to the definitions just provided.  It is to provide some background for the naming of classes like `IndexExtractor` and `BitMapExtractor` and what they mean conceptually.

src/main/java/org/apache/commons/collections4/bloomfilter/CountingBloomFilter.java
Claudenw: good catch.

aherbert (approved):

PR #509, opened 2024-06-25
[COLLECTIONS-858] CartesianProductIterator

alexey-pelykh:
The `CartesianProductIterator` is suitable to enhance a deeply-nested for-loops over a set of collections.

Given
```java
var a = new Arrays.asList('a1', 'a2', 'a3');
var b = new Arrays.asList('b1', 'b2', 'b3');
var c = new Arrays.asList('c1', 'c2', 'c3');
var d = new Arrays.asList('d1', 'd2', 'd3');
var e = new Arrays.asList('e1', 'e2', 'e3');
var f = new Arrays.asList('f1', 'f2', 'f3');
```

instead of

```java
for (var aElement : a) {
    for (var bElement : b) {
        for (var cElement : c) {
            for (var dElement : d) {
                for (var eElement : e) {
                    for (var fElement : f) {
                        // ...
                    }
                }
            }
        }
    }
}
```

it would look like:

```
var it = new CartesianProductIterator<>(a, b, c, d, e, f);
while (it.hasNext()) {
    var elements = it.next();
}
```

Extremely useful for nesting levels 10+.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/iterators/CartesianProductIterator.java
+++ b/src/main/java/org/apache/commons/collections4/iterators/CartesianProductIterator.java
@@ -1,0 +1,156 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.collections4.iterators;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
+/**
+ * This iterator creates a Cartesian product of the input iterables,
+ * equivalent to nested for-loops.
+ * <p>
+ * The iterables provided to the constructor are used in reverse order, each
+ * until exhaustion before proceeding to the next element of the prior iterable
+ * and repeating. Consider the following example:
+ *
+ * <pre>{@code
+ * List<Character> iterable1 = Arrays.asList('A', 'B', 'C');
+ * List<Character> iterable2 = Arrays.asList('1', '2', '3');
+ * CartesianProductIterator<Character> it = new CartesianProductIterator<>(
+ *         iterable1,
+ *         iterable2);
+ * while (it.hasNext()) {
+ *     List<Character> tuple = it.next();
+ *     System.out.println(tuple.get(0) + ", " + tuple.get(1));
+ * }
+ * }</pre>
+ *
+ * The output will be:
+ *
+ * <pre>
+ * A, 1
+ * A, 2
+ * A, 3
+ * B, 1
+ * B, 2
+ * B, 3
+ * C, 1
+ * C, 2
+ * C, 3
+ * </pre>
+ * <p>
+ * The {@code remove()} operation is not supported, and will throw an
+ * {@code UnsupportedOperationException}.
+ * <p>
+ * If any of the input iterables is empty, the Cartesian product will be empty.
+ * If any of the input iterables is infinite, the Cartesian product will be
+ * infinite.
+ *
+ * @param <E> the type of the objects being permuted
+ * @since 4.5.0
+ */
+public class CartesianProductIterator<E> implements Iterator<List<E>> {
+
+    /**
+     * The iterables to create the Cartesian product from.
+     */
+    private final List<Iterable<? extends E>> iterables;
+
+    /**
+     * The iterators to generate the Cartesian product tuple from.
+     */
+    private final List<Iterator<? extends E>> iterators;
+
+    /**
+     * The previous generated tuple of elements.
+     */
+    private List<E> previousTuple;
+
+    /**
+     * Constructs a new {@code CartesianProductIterator} instance with given iterables.
+     *
+     * @param iterables the iterables to create the Cartesian product from
+     * @throws NullPointerException if any of the iterables is null
+     */
+    @SafeVarargs
+    public CartesianProductIterator(final Iterable<? extends E>... iterables) {
+        Objects.requireNonNull(iterables, "iterables");
+        this.iterables = new ArrayList<>(iterables.length);
+        this.iterators = new ArrayList<>(iterables.length);
+        for (final Iterable<? extends E> iterable : iterables) {
+            Objects.requireNonNull(iterable, "iterable");
+            this.iterables.add(iterable);
+            final Iterator<? extends E> iterator = iterable.iterator();
+            if (!iterator.hasNext()) {
+                iterators.clear();
+                break;
+            }
+            iterators.add(iterator);
+        }
+    }
+
+    /**
+     * Returns {@code true} if the iteration has more elements.
+     *
+     * @return true if there are more tuples, otherwise false
+     */
+    @Override
+    public boolean hasNext() {
+        return iterators.stream().anyMatch(Iterator::hasNext);
+    }
+
+    /**
+     * Returns the next tuple of the input iterables.
+     *
+     * @return a list of the input iterables' elements
+     * @throws NoSuchElementException if there are no more tuples
+     */
+    @Override
+    public List<E> next() {
+        if (!hasNext()) {
+            throw new NoSuchElementException();
+        }
+
+        if (previousTuple == null) {
+            previousTuple = new ArrayList<>(iterables.size());
+            for (final Iterator<? extends E> iterator : iterators) {
+                previousTuple.add(iterator.next());
+            }
+            return new ArrayList<>(previousTuple);
+        }
+
+        for (int i = iterators.size() - 1; i >= 0; i--) {
+            Iterator<? extends E> iterator = iterators.get(i);
+            if (iterator.hasNext()) {
+                previousTuple.set(i, iterator.next());
+                return new ArrayList<>(previousTuple);
+            }
+            iterator = iterables.get(i).iterator();
+            iterators.set(i, iterator);
+            previousTuple.set(i, iterator.next());
+        }
+        throw new IllegalStateException("reached unreachable code");
+    }
+
+    @Override
+    public void remove() {
+        throw new UnsupportedOperationException("remove");
+    }
+}
--- a/src/test/java/org/apache/commons/collections4/iterators/CartesianProductIteratorTest.java
+++ b/src/test/java/org/apache/commons/collections4/iterators/CartesianProductIteratorTest.java
@@ -1,0 +1,209 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.collections4.iterators;
+
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.NoSuchElementException;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+
+/**
+ * Test class for {@link CartesianProductIterator}.
+ */
+public class CartesianProductIteratorTest extends AbstractIteratorTest<List<Character>> {
+
+    private List<Character> letters;
+    private List<Character> numbers;
+    private List<Character> symbols;
+    private List<Character> emptyList;
+
+    public CartesianProductIteratorTest() {
+        super(CartesianProductIteratorTest.class.getSimpleName());
+    }
+
+    @Override
+    public CartesianProductIterator<Character> makeEmptyIterator() {
+        return new CartesianProductIterator<>();
+    }
+
+    @Override
+    public CartesianProductIterator<Character> makeObject() {
+        return new CartesianProductIterator<>(letters, numbers, symbols);
+    }
+
+    @BeforeEach
+    public void setUp() {
+        letters = Arrays.asList('A', 'B', 'C');
+        numbers = Arrays.asList('1', '2', '3');
+        symbols = Arrays.asList('!', '?');
+        emptyList = Collections.emptyList();
+    }
+
+    @Override
+    public boolean supportsRemove() {
+        return false;
+    }
+
+    @Test
+    public void testRemoveThrows() {
+        final CartesianProductIterator<Character> it = makeObject();
+        assertThrows(UnsupportedOperationException.class, it::remove);
+    }
+
+    @Test
+    public void testEmptyCollection() {
+        final CartesianProductIterator<Character> it = new CartesianProductIterator<>(letters, Collections.emptyList());
+        assertFalse(it.hasNext());
+        assertThrows(NoSuchElementException.class, it::next);
+    }
+
+    /**
+     * test checking that all the tuples are returned
+     */
+    @Test
+    public void testExhaustivity() {
+        final List<Character[]> resultsList = new ArrayList<>();
+        final CartesianProductIterator<Character> it = makeObject();
+        while (it.hasNext()) {
+            final List<Character> tuple = it.next();
+            resultsList.add(tuple.toArray(new Character[0]));
+        }
+        assertThrows(NoSuchElementException.class, it::next);
+        assertEquals(18, resultsList.size());
+        final Iterator<Character[]> itResults = resultsList.iterator();
+        for (final Character a : letters) {
+            for (final Character b : numbers) {
+                for (final Character c : symbols) {
+                    assertArrayEquals(new Character[]{a, b, c}, itResults.next());
+                }
+            }
+        }
+    }
+
+    /**
+     * test checking that no tuples are returned when at least one of the lists is empty
+     */
+    @Test
+    public void testExhaustivityWithEmptyList() {
+        final List<Character[]> resultsList = new ArrayList<>();
+        final CartesianProductIterator<Character> it = new CartesianProductIterator<>(letters, emptyList, symbols);
+        while (it.hasNext()) {
+            final List<Character> tuple = it.next();
+            resultsList.add(tuple.toArray(new Character[0]));
+        }
+        assertThrows(NoSuchElementException.class, it::next);
+        assertEquals(0, resultsList.size());
+    }
+
+    /**
+     * test checking that no tuples are returned when first of the lists is empty
+     */
+    @Test
+    public void testExhaustivityWithEmptyFirstList() {
+        final List<Character[]> resultsList = new ArrayList<>();
+        final CartesianProductIterator<Character> it = new CartesianProductIterator<>(emptyList, numbers, symbols);
+        while (it.hasNext()) {
+            final List<Character> tuple = it.next();
+            resultsList.add(tuple.toArray(new Character[0]));
+        }
+        assertThrows(NoSuchElementException.class, it::next);
+        assertEquals(0, resultsList.size());
+    }
+
+    /**
+     * test checking that no tuples are returned when last of the lists is empty
+     */
+    @Test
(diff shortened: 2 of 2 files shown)


src/test/java/org/apache/commons/collections4/iterators/CartesianProductIteratorTest.java
```diff
@@ -0,0 +1,91 @@
...
+    public void testEmptyCollection() {
+        final CartesianProductIterator<Character> it = new CartesianProductIterator<>(letters, empty);
+        assertFalse(it.hasNext());
+    }
+
+    /**
+     * test checking that all the tuples are returned
+     */
+    @Test
+    public void testCartesianProductExhaustivity() {
```
aherbert, line 79: The test for the total output length should be supplemented with a test that the output is the desired Cartesian product:
```
[A, 1]
[A, 2]
[A, 3]
[B, 1]
[B, 2]
[B, 3]
[C, 1]
[C, 2]
[C, 3]
```

src/test/java/org/apache/commons/collections4/iterators/CartesianProductIteratorTest.java
```diff
@@ -0,0 +1,91 @@
...
+    @Override
+    public boolean supportsRemove() {
+        return false;
+    }
+
+    @Test
+    public void testEmptyCollection() {
+        final CartesianProductIterator<Character> it = new CartesianProductIterator<>(letters, empty);
+        assertFalse(it.hasNext());
+    }
```
aherbert, line 73: This should test `next()` throws an exception.

src/test/java/org/apache/commons/collections4/iterators/CartesianProductIteratorTest.java
```diff
@@ -0,0 +1,91 @@
...
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * Test class for CartesianProductIterator.
+ */
+public class CartesianProductIteratorTest extends AbstractIteratorTest<List<Character>> {
+
+    protected List<Character> letters;
+    protected List<Character> numbers;
+    protected List<Character> empty;
```
aherbert, line 35: Since empty is not reused it does not merit being created for each test fixture. You can use `Collections.emptyList()` instead; the other members can be private.

aherbert (changes requested):

src/test/java/org/apache/commons/collections4/iterators/CartesianProductIteratorTest.java
aherbert: Since this is meant to be like a nested for loop, let's test it like one:
```java
        final CartesianProductIterator<Character> it = makeObject();
        for (final Character a : letters) {
            for (final Character b : numbers) {
                Assertions.assertEquals(Arrays.asList(a, b), it.next());
            }
        }
        assertThrows(NoSuchElementException.class, () -> it.next());
```
You could even add another test with a third iterable to have a triple loop to demonstrate it works for increasing depth.

src/main/java/org/apache/commons/collections4/iterators/CartesianProductIterator.java
```diff
@@ -0,0 +1,124 @@
...
+
+import java.util.*;
+
+/**
+ * This iterator creates a Cartesian product of the input enumerables,
+ * equivalent to nested for-loops.
+ * <p>
+ * The {@code remove()} operation is not supported, and will throw an
+ * {@code UnsupportedOperationException}.
+ *
```
aherbert, line 27: This class could do with some javadoc to state the expected output behaviour, e.g. the iterator creates the Cartesian product of the iterables as a list. Iterables provided to the constructor are used in reverse order, each until exhaustion before proceeding to the next element of the prior iterable and repeating:
```
iterable1 = A, B, C
iterable2 = 1, 2, 3
output:
A, 1
A, 2
A, 3
B, 1
B, 2
...
```

src/main/java/org/apache/commons/collections4/iterators/CartesianProductIterator.java
```diff
@@ -0,0 +1,124 @@
...
+            previousTuple.set(iteratorIndex, iterator.next());
+        }
+        return new ArrayList<>(previousTuple);
+    }
+
+    @Override
+    public void remove() {
+        throw new UnsupportedOperationException("remove() is not supported");
+    }
+
```
aherbert, line 123: Remove empty line

src/test/java/org/apache/commons/collections4/iterators/CartesianProductIteratorTest.java
```diff
@@ -0,0 +1,104 @@
...
+            final List<Character> tuple = it.next();
+            resultsList.add(tuple.toArray(new Character[0]));
+        }
+        assertThrows(NoSuchElementException.class, () -> it.next());
+        assertEquals(9, resultsList.size());
+        for (int i = 0; i < 9; i++) {
+            assertArrayEquals(expectedResultsList.get(i), resultsList.get(i));
+        }
+    }
+
```
aherbert, line 103: Remove empty line

src/main/java/org/apache/commons/collections4/iterators/CartesianProductIterator.java
```diff
@@ -0,0 +1,124 @@
...
+        if (previousTuple == null) {
+            previousTuple = new ArrayList<>(iterables.size());
+            for (final Iterator<? extends E> iterator : iterators) {
+                previousTuple.add(iterator.next());
+            }
+            return new ArrayList<>(previousTuple);
+        }
+
+        for (
```
aherbert, line 102: A single line for loop is in keeping with the style of the codebase:
```java
        for (int i = iterators.size() - 1; i >= 0; i--) {
```

aherbert (changes requested):

src/main/java/org/apache/commons/collections4/iterators/CartesianProductIterator.java
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * This iterator creates a Cartesian product of the input iterables,
+ * equivalent to nested for-loops.
+ * <p>
+ * The iterables provided to the constructor are used in reverse order, each
+ * until exhaustion before proceeding to the next element of the prior iterable
+ * and repeating. Consider the following example:
+ * <pre>
+ *     List<Character> iterable1 = Arrays.asList('A', 'B', 'C');
```
aherbert, line 29: I think this will not pass javadoc due to the `<>` characters. You need to include: `<pre>{@code` and `}</pre>` at the ends.

src/test/java/org/apache/commons/collections4/iterators/CartesianProductIteratorTest.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.*;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+
+/**
+ * Test class for CartesianProductIterator.
```
aherbert, line 28: Might as well put this in a `{@link ...}` tag.

aherbert (changes requested):
This is looking good apart from the javadoc which will fail the build. If you use `<,>` chars you need a code tag.

garydgregory:
@alexey-pelykh 
Make sure you run 'mvn' by itself before you push to run all build checks.

alexey-pelykh:
@garydgregory sorry, my oversight - should be green now

alexey-pelykh:
@garydgregory would it be possible to restart 23-ae? It's not exactly clear what is wrong

alexey-pelykh:
What would be the next steps?

src/main/java/org/apache/commons/collections4/iterators/CartesianProductIterator.java
```diff
@@ -0,0 +1,152 @@
...
+ * C, 1
+ * C, 2
+ * C, 3
+ * </pre>
+ * <p>
+ * The {@code remove()} operation is not supported, and will throw an
+ * {@code UnsupportedOperationException}.
+ *
+ * @param <E>  the type of the objects being permuted
+ */
```
garydgregory, line 62: We need a Javadoc 'since tag here.

alexey-pelykh: I'd be glad to add that if I knew what to specify after `@since`

garydgregory: The pom.xml and changes.xml files say the next version will be 4.5.0.

alexey-pelykh: @garydgregory thanks! Done 👍

garydgregory:
@alexey-pelykh 
This PR has zero text in its description.

alexey-pelykh:
@garydgregory that is true, any specific requirements what to write there apart of what I've already added to it?

garydgregory:
@alexey-pelykh 
I want to email the dev list and ask for reviews, and the first I do at least is read the description to know what it is I'm looking at... it's up to you to help reviewers...

alexey-pelykh:
@garydgregory okies, done 👍

alexey-pelykh:
Seems not much of a discussion :)

alexey-pelykh:
👋 @garydgregory anything I can do to push this forward?

src/main/java/org/apache/commons/collections4/iterators/CartesianProductIterator.java
```diff
@@ -0,0 +1,157 @@
...
+     */
+    private final List<Iterator<? extends E>> iterators;
+
+    /**
+     * The previous generated tuple of elements.
+     */
+    private List<E> previousTuple;
+
+    /**
+     * Standard constructor for this class.
```
garydgregory, line 84: "Standard" does not mean anything IMO. I would say "Constructs a new instance that ...".

src/test/java/org/apache/commons/collections4/iterators/CartesianProductIteratorTest.java
```diff
@@ -0,0 +1,100 @@
...
+    @Override
+    public boolean supportsRemove() {
+        return false;
+    }
+
+    @Test
+    public void testEmptyCollection() {
+        final CartesianProductIterator<Character> it = new CartesianProductIterator<>(letters, Collections.emptyList());
+        assertFalse(it.hasNext());
+        assertThrows(NoSuchElementException.class, () -> it.next());
```
garydgregory, line 74: Use a method reference for `next()`.

src/test/java/org/apache/commons/collections4/iterators/CartesianProductIteratorTest.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /**
+     * test checking that all the tuples are returned
+     */
+    @Test
+    public void testCartesianProductExhaustivity() {
+        final List<Character[]> resultsList = new ArrayList<>();
+
```
garydgregory, line 83: Whitespace is not needed.

src/main/java/org/apache/commons/collections4/iterators/CartesianProductIterator.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    /**
+     * Indicates if there are more tuples to return.
+     *
+     * @return true if there are more tuples, otherwise false
+     */
+    @Override
+    public boolean hasNext() {
+        for (final Iterator<? extends E> iterator : iterators) {
```
garydgregory, line 112: You can use the less verbose `return iterators.stream().anyMatch(Iterator::hasNext);`

src/main/java/org/apache/commons/collections4/iterators/CartesianProductIterator.java
```diff
@@ -0,0 +1,157 @@
...
+     */
+    private List<E> previousTuple;
+
+    /**
+     * Standard constructor for this class.
+     *
+     * @param iterables the iterables to create the Cartesian product from
+     * @throws NullPointerException if any of the iterables is null
+     */
+    public CartesianProductIterator(final Iterable<? extends E>... iterables) {
```
garydgregory, line 89: Since the code has null-guards, you can use `@SafeVarargs`.

src/main/java/org/apache/commons/collections4/iterators/CartesianProductIterator.java
```diff
@@ -0,0 +1,157 @@
...
+            Iterator<? extends E> iterator = iterators.get(i);
+            if (iterator.hasNext()) {
+                previousTuple.set(i, iterator.next());
+                return new ArrayList<>(previousTuple);
+            }
+            iterator = iterables.get(i).iterator();
+            iterators.set(i, iterator);
+            previousTuple.set(i, iterator.next());
+        }
+        throw new IllegalStateException();
```
garydgregory, line 150: Describe the state in the exception message.

src/main/java/org/apache/commons/collections4/iterators/CartesianProductIterator.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    @Override
+    public void remove() {
+        throw new UnsupportedOperationException("remove() is not supported");
```
garydgregory, line 155: Keep the message language-neutral, like `java.util.Iterator.remove()`:
```java
throw new UnsupportedOperationException("remove");
```

src/main/java/org/apache/commons/collections4/iterators/CartesianProductIterator.java
```diff
@@ -0,0 +1,157 @@
...
+     * @throws NullPointerException if any of the iterables is null
+     */
+    public CartesianProductIterator(final Iterable<? extends E>... iterables) {
+        Objects.requireNonNull(iterables, "iterables");
+        this.iterables = Arrays.asList(iterables);
+        this.iterators = new ArrayList<>(iterables.length);
+        for (final Iterable<? extends E> iterable : iterables) {
+            Objects.requireNonNull(iterable, "iterable");
+            final Iterator<? extends E> iterator = iterable.iterator();
+            if (iterator.hasNext()) {
```
garydgregory, line 96: You don't need both blocks nested if you say:
```java
            if (!iterator.hasNext()) {
                iterators.clear();
                break;
            }
            iterators.add(iterator);
```

src/main/java/org/apache/commons/collections4/iterators/CartesianProductIterator.java
```diff
@@ -0,0 +1,157 @@
...
+                iterators.add(iterator);
+            } else {
+                iterators.clear();
+                break;
+            }
+        }
+    }
+
+    /**
+     * Indicates if there are more tuples to return.
```
garydgregory, line 106: "Indicates" is not what you normally see in Javadocs. When in doubt, look at the specification in the interface:
```
Returns {@code true} if the iteration has more elements.
```

garydgregory (changes requested):
Hello @alexey-pelykh 
- You are missing a unit test for Iterator.forEachRemaining()
- Also edge cases like what if the first list, a middle list or the last list is empty.
- IOW what is the expected behavior if in your example `letters`, `numbers`, or `symbols` is empty
- What should happen if the lists point to each other or have cycles?

garydgregory:
Just FTR https://stackoverflow.com/questions/32131987/how-can-i-make-cartesian-product-with-java-8-streams

src/main/java/org/apache/commons/collections4/iterators/CartesianProductIterator.java
alexey-pelykh: @garydgregory reasonable, yet that would fall out of line with other iterators present in the library - most of them use `remove() is not supported` or more detailed `remove() method is not supported for an ObjectArrayIterator`. Taking that into account, is this change request remains and is a blocking one?

src/main/java/org/apache/commons/collections4/iterators/CartesianProductIterator.java
alexey-pelykh: I agree that `@SafeVarargs` should be here, yet I don't get it why do you refer to the null-guards as the reason since it's not related with Object[] arriving in real-time. However, I did address that anyways so the annotation should be applied.

src/main/java/org/apache/commons/collections4/iterators/CartesianProductIterator.java
alexey-pelykh: While I agree that it's less verbose, I don't think that justifies all the call chain and objects creation that need to happen in that line. Is this change request a blocking one?

alexey-pelykh:
Hi @garydgregory:

> You are missing a unit test for Iterator.forEachRemaining()

What is the rationale to add that given that `forEachRemaining` is a `hasNext()`/`next()` wrapper and those are explicitly tested?

> Also edge cases like what if the first list, a middle list or the last list is empty.

Reasonable, added.

> IOW what is the expected behavior if in your example letters, numbers, or symbols is empty

Added tests to be explicit. If at least one list is empty, cartesian product returns no tuples.

> What should happen if the lists point to each other or have cycles?

If any of the passed iterables has cycle and thus is infinite, the iterator would also produce an infinite result. Here the behavior is the same as in nested for-loop. Could you please elaborate on the lists pointing to each other case as I'm not exactly sure I got you there.

src/main/java/org/apache/commons/collections4/iterators/CartesianProductIterator.java
alexey-pelykh: Fair enough, not my preferred form yet no objection either.

src/main/java/org/apache/commons/collections4/iterators/CartesianProductIterator.java
garydgregory: Hello @alexey-pelykh 
What I am saying is:
- Change the exception message
- Add a test that asserts the exception is always thrown

src/main/java/org/apache/commons/collections4/iterators/CartesianProductIterator.java
garydgregory: The difference (to me) is whether you make the change or I do it, which affects how fast I can process PRs and requests in general. If a PR is more work for me, I just put it aside and get back to it, later. If I took the time to review the PR, test it, and write the comment, then I expect the change to be made.

garydgregory:
> Hi @garydgregory:
> 
> > You are missing a unit test for Iterator.forEachRemaining()
> 
> What is the rationale to add that given that `forEachRemaining` is a `hasNext()`/`next()` wrapper and those are explicitly tested?
> 
> > Also edge cases like what if the first list, a middle list or the last list is empty.
> 
> Reasonable, added.
> 
> > IOW what is the expected behavior if in your example letters, numbers, or symbols is empty
> 
> Added tests to be explicit. If at least one list is empty, cartesian product returns no tuples.
> 
> > What should happen if the lists point to each other or have cycles?
> 
> If any of the passed iterables has cycle and thus is infinite, the iterator would also produce an infinite result. Here the behavior is the same as in nested for-loop. Could you please elaborate on the lists pointing to each other case as I'm not exactly sure I got you there.

If the API allows an infinite loop, this is what will happen: Someone will send an email to our security mailing list saying they found a critical DOS vulnerability and ask for a CVE to be credited to them; time will be taken analyzing, deciding what to do and replying to the poster. This is why we've fixed these types of bugs overtime and added tests like 
- `HashCodeBuilderTest.testReflectionObjectCycle()`
- `ToStringBuilderTest.testSelfInstanceTwoVarsReflectionObjectCycle()`
- `ToStringBuilderTest.testSelfInstanceVarReflectionObjectCycle()`
- `ToStringBuilderTest.testSimpleReflectionObjectCycle()`
- `StrSubstitutorTest.testCyclicReplacement()`
- `ToStringBuilderTest.testReflectionArrayAndObjectCycle()`
- `ToStringBuilderTest.testReflectionArrayArrayCycle()`
- ...

And yes, there are probably more of these types of bugs lurking around, so let's not add new ones ;-)

src/main/java/org/apache/commons/collections4/iterators/CartesianProductIterator.java
alexey-pelykh: That is what I meant as "blocking" - as in, the reviewer is not open to discuss the change request if the author has counter-arguments towards it.

Change done.

src/main/java/org/apache/commons/collections4/iterators/CartesianProductIterator.java
alexey-pelykh: Thanks for clarifying, change done.

alexey-pelykh:
> And yes, there are probably more of these types of bugs lurking around, so let's not add new ones ;-)

Absolutely not my intent! That said, I guess I'm too dumb to understand how that bug may be exploited in this case.

- If a `CartesianProductIterator` is passed an infinite iterable, the result is expectedly infinite - the same was as for loop over infinite iterable is going to be infinite.
- If a `CartesianProductIterator` is passed an iterable that is modified while iterating over the iterator, it may or may not yield into an infinite iterator.
- If a `CartesianProductIterator` is passed a fake iterable that always returns reference to the `CartesianProductIterator` it was passed to ("iterating over self") - it may be infinite.

However, the infinite loop won't happen in the`CartesianProductIterator` itself, but the `hasNext()` will always be true - so iterating over such an iterator will be infinite, alike iterating over a "random generator" iterator.

garydgregory:
> @garydgregory would it be possible to restart 23-ae? It's not exactly clear what is wrong

EA builds are marked experimental in the CI and fail when the tooling has not been adapted Java EA versions. In this case probably some locale changes.

garydgregory:
> > You are missing a unit test for Iterator.forEachRemaining()
> 
> What is the rationale to add that given that `forEachRemaining` is a `hasNext()`/`next()` wrapper and those are explicitly tested?
> 

The new class implements an interface. The tests should show what are the expectations for all interface methods, the ones that are actually implemented by the new class _and_ the behavior that is inherited from the default method.

There is no guarantee that implementing the methods that a default interface calls is actually correctly implementing the contract for the default method. Testing the default methods shows that it works as expected, regardless of whether the default method is implemented or inherited. You are also showing as a PR author that you've considered the big picture and are thorough, all the pieces work individually and together, as expected.

alexey-pelykh:
> Testing the default methods shows that it works as expected, regardless of whether the default method is implemented or inherited.

👍  test added

garydgregory:
@alexey-pelykh 

Thank you for your PR and patience. Merged! 🚀

PR #515, opened 2024-07-16
Document interaction between peek and filter iterator

benjamin-confino:
Resolves: https://issues.apache.org/jira/browse/COLLECTIONS-856

This documents the interaction between a peeking iterator and a filtering iterator; that if you call `peek()` or `element()`, `fill()` will find the next predicate approved element from the underlying filtering iterator, stash it in `slot`, ensuring it will be returned in the future even if the object changes and no longer matches the predicate.

(This PR also removes an incorrect line saying `element()` will return null if the underlying iterator is exhausted)


garydgregory:
As I could not resolve the conflicts from here, I committed this directly to git master with @benjamin-confino as the author. TY for you patience! Closing.

benjamin-confino:
@garydgregory No worries, and thanks for merging.

PR #517, opened 2024-07-19
Increase test coverage for ListUtils

sigee:
- isEqualList: same size, but different elements in the lists
 - isEqualList: same size, same elements, but different order
 - hashCodeForList: with null element
 - partition.get: edge cases (Exception)
 - union: multiple cases
 - sum: happy path

### DIFF
--- a/src/test/java/org/apache/commons/collections4/ListUtilsTest.java
+++ b/src/test/java/org/apache/commons/collections4/ListUtilsTest.java
@@ -91,6 +91,15 @@ public class ListUtilsTest {
         assertFalse(ListUtils.isEqualList(a, null));
         assertFalse(ListUtils.isEqualList(null, b));
         assertTrue(ListUtils.isEqualList(null, null));
+
+        b.clear();
+        a.add("a");
+        b.add("b");
+        assertFalse(ListUtils.isEqualList(a, b));
+
+        a.add("b");
+        b.add("a");
+        assertFalse(ListUtils.isEqualList(a, b));
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


src/test/java/org/apache/commons/collections4/ListUtilsTest.java
```diff
@@ -81,16 +81,25 @@ public void testEmptyIfNull() {
...
-        final List<String> a = new ArrayList<>( data );
-        final List<String> b = new ArrayList<>( data );
-
-        assertEquals(a, b);
-        assertTrue(ListUtils.isEqualList(a, b));
-        a.clear();
-        assertFalse(ListUtils.isEqualList(a, b));
-        assertFalse(ListUtils.isEqualList(a, null));
-        assertFalse(ListUtils.isEqualList(null, b));
+        final List<String> list1 = new ArrayList<>( data );
```
garydgregory, line 84: Please don't rename variables. I can't tell what's changed.

sigee: I renamed them because they overlapped the String constants from lines 46, 47. And they were confusing.
```
private static final String a = "a";
private static final String b = "b";
```
```
final List<String> a = new ArrayList<>( data );
final List<String> b = new ArrayList<>( data );
```

sigee:
I split the changes into two commits to make them more understandable.
- Increase the coverage
- Rename local variables

garydgregory:
@sigee 
Merged. TY.

PR #518, opened 2024-07-23
Use consistent junit (jupiter) api everywhere

sigee:


garydgregory:
TY. Too bad Guava still sucks in JUnit 4.

PR #537, opened 2024-08-31
Add findDuplicates Method to ListUtils with Comprehensive Test Cases

hemanth0525:
### Summary

This pull request introduces the `findDuplicates` method to the `ListUtils` class, a feature designed to enhance the functionality of list operations by providing a straightforward way to detect duplicate elements.

### Why This Change?

In many software development scenarios, detecting duplicates within a list is a common task. This functionality is essential for tasks such as data validation, cleaning, and processing. Instead of requiring developers to implement custom solutions or rely on third-party libraries, this method integrates seamlessly into the `ListUtils` class, offering a reliable and efficient way to handle duplicate detection.

### How This Change is Useful

- **Efficiency:**
  - The `findDuplicates` method is implemented with optimal performance in mind, ensuring that duplicate detection is handled efficiently even for large lists.

- **Simplification:**
  - Provides a single, reusable method for duplicate detection, reducing the need for repetitive code across different projects.

- **Versatility:**
  - Handles a variety of scenarios, including lists with different types of elements, empty lists, and lists with no duplicates.

### Example Usage

```java
List<Integer> numbers = Arrays.asList(1, 2, 3, 2, 4, 5, 3);
List<Integer> duplicates = ListUtils.findDuplicates(numbers); // [2, 3]

### DIFF
--- a/src/test/java/org/apache/commons/collections4/IterableUtilsTest.java
+++ b/src/test/java/org/apache/commons/collections4/IterableUtilsTest.java
@@ -39,7 +39,6 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
-import java.util.Objects;
 
 import org.apache.commons.collections4.bag.HashBag;
 import org.apache.commons.lang3.StringUtils;
--- a/src/test/java/org/apache/commons/collections4/IterableUtilsTest.java
+++ b/src/test/java/org/apache/commons/collections4/IterableUtilsTest.java
@@ -38,6 +38,7 @@ import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Set;
+import java.util.Objects;
 
 import org.apache.commons.collections4.bag.HashBag;
 import org.apache.commons.lang3.StringUtils;
@@ -298,6 +299,92 @@ public class IterableUtilsTest {
     public void testDuplicatSetAllSameInDeque() {
         final Deque<Integer> input = new ArrayDeque<>(Arrays.asList(5, 5, 5, 5));
         assertEquals(new HashSet<>(Arrays.asList(5)), IterableUtils.duplicateSet(input));
+    }
+
+    @Test
+    public void testDuplicateListMixedElementTypes() {
+        final List<Object> input = Arrays.asList(1, "a", 2, "a", 3, "b", "b");
+        final List<Object> expected = Arrays.asList("a", "b");
+        assertEquals(expected, IterableUtils.duplicateList(input));
+    }
+
+    @Test
+    public void testDuplicateListNestedCollections() {
+        final List<List<Integer>> input = Arrays.asList(
+            Arrays.asList(1, 2, 3),
+            Arrays.asList(1, 2, 3),
+            Arrays.asList(4, 5, 6),
+            Arrays.asList(4, 5, 6)
+        );
+        final List<List<Integer>> expected = Arrays.asList(
+            Arrays.asList(1, 2, 3),
+            Arrays.asList(4, 5, 6)
+        );
+        assertEquals(expected, IterableUtils.duplicateList(input));
+    }
+
+    @Test
+    public void testDuplicateListCustomObjects() {
+        class Person {
+            String name;
+            int age;
+
+            Person(String name, int age) {
+                this.name = name;
+                this.age = age;
+            }
+
+            @Override
+            public boolean equals(Object o) {
+                if (this == o) {
+                    return true;
+                }
+                if (o == null || getClass() != o.getClass()) {
+                    return false;
+                }
+                Person person = (Person) o;
+                return age == person.age && name.equals(person.name);
+            }
+
+            @Override
+            public int hashCode() {
+                return Objects.hash(name, age);
+            }
+        }
+
+        final List<Person> input = Arrays.asList(
+            new Person("Linus", 30),
+            new Person("Torvalds", 25),
+            new Person("Linus", 30),
+            new Person("Tim", 35),
+            new Person("Torvalds", 25)
+        );
+        final List<Person> expected = Arrays.asList(
+            new Person("Linus", 30),
+            new Person("Torvalds", 25)
+        );
+        assertEquals(expected, IterableUtils.duplicateList(input));
+    }
+
+    @Test
+    public void testDuplicateListLargeScaleData() {
+        final List<Integer> input = new ArrayList<>();
+        for (int i = 0; i < 1000; i++) {
+            input.add(i % 100);
+        }
+        final List<Integer> expected = new ArrayList<>();
+        for (int i = 0; i < 100; i++) {
+            expected.add(i);
+        }
+        final List<Integer> actual = IterableUtils.duplicateList(input);
+        assertEquals(expected, actual);
+    }
+
+    @Test
+    public void testDuplicateListImmutableCollections() {
+        final List<Integer> input = Collections.unmodifiableList(Arrays.asList(1, 2, 3, 1, 2, 4));
+        final List<Integer> expected = Arrays.asList(1, 2);
+        assertEquals(expected, IterableUtils.duplicateList(input));
     }
 
     @Test

### CODE REVIEW

**Review by garydgregory — COMMENTED** [review]
Since the result is effectively a set and the input does not have to be restricted to a list, it would be better to do this IMO:
```java
    public static <E> Set<E> duplicates(final Collection<E> list) {
        final HashSet<E> dups = new HashSet<>();
        final HashSet<E> set = new HashSet<>();
        for (E e : list) {
            (set.contains(e) ? dups : set).add(e);
        }
        return dups;
    }
```
and put it in `CollectionUtils` probably.

hemanth0525:
> Since the result is effectively a set and the input does not have to be restricted to a list, it would be better to do this IMO:
> 
> ```java
>     public static <E> Set<E> duplicates(final Collection<E> list) {
>         final HashSet<E> dups = new HashSet<>();
>         final HashSet<E> set = new HashSet<>();
>         for (E e : list) {
>             (set.contains(e) ? dups : set).add(e);
>         }
>         return dups;
>     }
> ```
> 
> and put it in `CollectionUtils` probably.

Thanks for the feedback. The `ListUtils` implementation is tailored for `List<E>`, preserving the order of elements and identifying duplicates where the sequence matters. This approach is particularly useful in scenarios such as:

* **Preserving Order**: When the sequence of elements is important, such as in order-sensitive algorithms or data processing tasks.
* **List-Specific APIs**: When working with APIs that return or expect `List`, ensuring that the utility aligns with the data structure's inherent properties.
* **Performance Considerations**: In cases where developers need efficient duplicate detection while maintaining the list's order, which can be crucial in performance-sensitive applications.

Keeping this method in `ListUtils` supports these specific needs and maintains consistency for developers working with ordered collections.

src/main/java/org/apache/commons/collections4/ListUtils.java
```diff
@@ -173,6 +174,39 @@ public static <E> List<E> fixedSizeList(final List<E> list) {
...
+     * duplicates in a result list.
+     * </p>
+     *
+     * @param <E> the type of elements in the list
+     * @param list the list to check for duplicates, must not be null
+     * @return a list of duplicate elements, or an empty list if no duplicates are found
+     * @throws NullPointerException if the list is null
+     */
+    public static <E> List<E> findDuplicates(final List<E> list) {
+        if (list == null) {
```
garydgregory, line 191: Use `Objects.requireNonNull(list, "list");`

src/main/java/org/apache/commons/collections4/ListUtils.java
```diff
@@ -173,6 +174,39 @@ public static <E> List<E> fixedSizeList(final List<E> list) {
...
+     * This method uses two sets: one for tracking seen elements and one for
+     * collecting duplicates. It iterates through the list once and collects
+     * duplicates in a result list.
+     * </p>
+     *
+     * @param <E> the type of elements in the list
+     * @param list the list to check for duplicates, must not be null
+     * @return a list of duplicate elements, or an empty list if no duplicates are found
+     * @throws NullPointerException if the list is null
+     */
```
garydgregory, line 189: Add `@since 4.5.0-M3`

src/test/java/org/apache/commons/collections4/ListUtilsTest.java
```diff
@@ -102,6 +102,54 @@ public void testEquals() {
...
+        List<Integer> actual = ListUtils.findDuplicates(input);
+        assertEquals(expected, actual, "The list should be empty as the input list is empty.");
+    }
+
+    @Test
+    public void testFindDuplicatesSingleElement() {
+        List<Integer> input = Arrays.asList(1);
+        List<Integer> expected = Arrays.asList();
+        List<Integer> actual = ListUtils.findDuplicates(input);
+        assertEquals(expected, actual, "The list should be empty as there is only one element.");
```
garydgregory, line 134: Empty results can be simply asserted as `assertTrue(ListUtils.duplicates(input).isEmpty());`

I'm going to create a CollectionUtils version of the method called `duplicates`, I suggest the method be named the same here.

garydgregory (changes requested):
See my comments. The input can still be

src/test/java/org/apache/commons/collections4/ListUtilsTest.java
hemanth0525: tests for CollectionUtils are written in CollectionUtilsTest.java right. do we still we gotta change the method name ?

garydgregory:
Hello @hemanth0525 

Please see git master for:

- Add CollectionUtils.duplicateList(Collection<E>)
- Add CollectionUtils.duplicateSequencedSet(Collection<E>)
- Add CollectionUtils.duplicateSet(Collection<E>)

You are credited in `changes.xml`

Are there tests or functionality we could provide specific to ListUtils?

I believe that this PR's new ListUtils method is now the same in functionality as `CollectionUtils.duplicateList(Collection<E>)`



TY!

hemanth0525:
hello @garydgregory,

TY !

i have run mvn locally and build success !

hemanth0525:
hello @garydgregory 
is this PR ready to merge ?

src/main/java/org/apache/commons/collections4/ListUtils.java
```diff
@@ -173,6 +174,36 @@ public static <E> List<E> fixedSizeList(final List<E> list) {
...
+     * </p>
+     *
+     * @param <E> the type of elements in the list
+     * @param list the list to check for duplicates, must not be null
+     * @return a list of duplicate elements, or an empty list if no duplicates are found
+     * @throws NullPointerException if the list is null
+     * @since 4.5.0-M3
+     */
+    public static <E> List<E> findDuplicates(final List<E> list) {
+        Objects.requireNonNull(list, "The input list must not be null.");
```
garydgregory, line 192: @hemanth0525 

If you replace the body of this (git master) method with:
```java
return IterableUtils.duplicateList(list);
```
You'll find that all tests pass.
So I don't think we need this method in this class.

hemanth0525: u mean from CollectionUtils ?

hemanth0525: if yes, I have done it. So yeah utilizing the other function to help list functionalities.

src/main/java/org/apache/commons/collections4/ListUtils.java
```diff
@@ -173,6 +173,24 @@ public static <E> List<E> fixedSizeList(final List<E> list) {
...
+     * </p>
+     *
+     * @param <E> the type of elements in the list
+     * @param list the list to check for duplicates, must not be null
+     * @return a list of duplicate elements, or an empty list if no duplicates are found
+     * @throws NullPointerException if the list is null
+     * @since 4.5.0-M3
+     */
+    public static <E> List<E> findDuplicates(final List<E> list) {
+        return CollectionUtils.duplicateList(list);
```
garydgregory, line 191: If you rebase on git master, you'll find this method in 'IterableUtils'.

hemanth0525:
Oh Yeah !!
Done!!

garydgregory:
Hello @hemanth0525 
If all the new method does is delegate to `IterableUtils.duplicateList()`, it is unnecessary. The tests in this PR cover may use cases not in `IterableUtilsTest`.

hemanth0525:
@garydgregory Okay so what u want me to do in this case ?

garydgregory:
I would either close the PR or update it if there are missing use cases in IterableUtilsTest (if yes, then also remove the new method from this PR).

garydgregory:
I don't understand what you mean.

hemanth0525:
sorry !
I will look for the use cases.

Suggested Improvement
For the duplicateList method, it might be more appropriate to call duplicateSequencedSet instead of duplicateSet. Since the method returns a List, preserving the order of elements is often expected or desirable. By using duplicateSequencedSet, we can ensure that the order of duplicate elements is maintained, providing a more intuitive and consistent behavior for users of this method.
```
    /**
     * Finds and returns the List of duplicate elements in the given collection.
     *
     * @param <E> the type of elements in the collection.
     * @param iterable the list to test, must not be null.
     * @return the set of duplicate elements, may be empty.
     * @since 4.5.0-M3
     */
    public static <E> List<E> duplicateList(final Iterable<E> iterable) {
        return new ArrayList<>(duplicateSet(iterable));
    }
```
since we are using list . we tend to preserve the order in many cases

garydgregory:
@hemanth0525 

In git master, we now make sure `IterableUtils.duplicateList()` has predictable order in its
result.

- Add `testDuplicateListMultipleDuplicatesInDequeReverse()`
- Add `testDuplicateListMultipleDuplicatesInListReverse()`
- If you do not care about order, you can use `IterableUtils.duplicateSet()`

hemanth0525:
@garydgregory I just did that and thought of commiting...........

hemanth0525:
Hello @garydgregory,

I have added several diverse test cases to `IterableUtilsTest.java`. Kindly review them.

hemanth0525:
update on this PR ?

garydgregory:
Hello @hemanth0525 

Does the PR now provide additional code coverage? You can tell when you run "mvn clean install site" and you look at the JaCoCo report. 

Gary

hemanth0525:
i dont see any additional code coverage

hemanth0525:
this PR started to provide functionalities to find duplicates to list which u have implemented in IterableUtils for all collections. so ig lets close this PR now

PR #538, opened 2024-09-02
Fix the bug of MapUtils.getString

F0otman:
MapUtils have added several new methods as getXXX allowing 3rd param as functional input
But for String, it allowed this in 4.4
image

But now it would **make static compilation error** in 4.5.0-M2
image

**It's a design bug for conflict input param and stopped our old project**, there are 3 ways to fix:
--Type1 is like this pull, adjust the sequence of params. (Recommended)

### DIFF
--- a/src/main/java/org/apache/commons/collections4/MapUtils.java
+++ b/src/main/java/org/apache/commons/collections4/MapUtils.java
@@ -1142,15 +1142,14 @@ public class MapUtils {
      * produce the default value if the conversion fails.
      *
      * @param <K> the key type
+     * @param defaultFunction what to produce the default value if the value is null or if the conversion fails
      * @param map the map whose value to look up
      * @param key the key of the value to look up in that map
-     * @param defaultFunction what to produce the default value if the value is null or if the conversion fails
      * @return the value in the map as a string, or defaultValue produced by the defaultFunction if the original value
      *         is null, the map is null or the string conversion fails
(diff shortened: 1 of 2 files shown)


F0otman:
Type2 is like this, add a param for new method. (Recommended) d022a3c816f8d0eb4fcbb09502770f83f3ace4c2

F0otman:
Type3 is like this, fix 3rd param as Object. 338cc08fe3378fa5e8e75db1cb1c5bf8bf14384b

aherbert:
This being a compile error, you can you adjust your code to allow it to compile. E.g. can you give the compiler a hint on the method to call:
```java
String string = MapUtils.getString(new HashMap<>(), key: "key", defaultValue: (String) null);
```
I presume your old code that was was previously compiled against 4.4 does not have a runtime error when running against 4.5 (i.e. the JVM links the method correctly).

garydgregory:
I agree with @aherbert 
The code is binary compatible, please adjust your sources.

F0otman:
Actually it indeed stopped our sources, cause we use CI/CD to continually compile project.

The main problem is 4.5.X has compatibility issues now, it's much more important to downward compatibility than change the new method.
("Null is allowed as 3rd param but not in 4.5", it's not only for getString but also for other getXXX new methods with ambiguating  problem)

garydgregory:
Closing: Information provided.

PR #541, opened 2024-09-11
Add reverse and reverseInPlace methods to ArrayUtils

hemanth0525:
- **Implemented `reverse()` method**: Returns a new array with elements in reversed order. Useful for cases where the original array should remain unchanged and a reversed copy is needed for operations or display.
- **Implemented `reverseInPlace()` method**: Reverses the elements of the array in place. This method is efficient for scenarios where modifying the original array is acceptable and memory usage needs to be minimized.
- **Enhanced test coverage**: Added comprehensive unit tests for both methods, including edge cases such as null arrays and empty arrays. Ensures reliability and correctness of the new functionality.

These additions enhance the utility of `ArrayUtils` by providing common array manipulation capabilities, which can simplify code and improve performance in scenarios requiring array reversals.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/ArrayUtils.java
+++ b/src/main/java/org/apache/commons/collections4/ArrayUtils.java
@@ -109,6 +109,52 @@ final class ArrayUtils {
     }
 
     /**
+     * <p> 
+     * Reverses the given array and returns a new array.
+     * </p>
+     * <p>
+     * This method returns {@code null} if the input array is {@code null}.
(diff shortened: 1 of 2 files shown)


garydgregory (changes requested):
-1: Why add package-private methods that this component does not use and non-component call sites cannot call? Maybe you should consider Commons Lang's ArrayUtils class.

hemanth0525:
Yeah will do it !!

PR #542, opened 2024-09-13
Sorted List binarySearch functions

kvr000:
Based on PR for commons-lang3 https://github.com/apache/commons-lang/pull/1270

There is `Collections.binarySearch()` in Java core. However, it searches for exact element which is often not usable, as the elements are typically sorted by particular field.

These methods introduce `binarySearch()` on (sorted) `List`, with ability to provide sorting key extraction function and key comparator.

The rest of semantics, behavior, parameters and return values remain consistent with `Collections.binarySearch()` from Java core.


garydgregory:
@kvr000 
Please only provide only one binarySearch() API in this PR, like `ArrayUtils` in Commons Lang, no First/Last API.

PR #551, opened 2024-10-04
Add IterableUtils.any() method

spijkercenter:
- Introduces a new utility method 'any()' in IterableUtils
- Returns an Optional containing the first element of an Iterable
- Handles null and empty iterables, returning an empty Optional

The method is named 'any()' to reflect its behavior of returning any element from the Iterable (specifically, the first one). This naming is consistent with similar concepts in other programming languages and libraries where 'any' often denotes "give me any element that satisfies a condition" - in this case, the condition is simply existence within the Iterable.

This addition provides a convenient way to safely retrieve the first  element of an Iterable, wrapped in an Optional, with null-safety built-in.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/IterableUtils.java
+++ b/src/main/java/org/apache/commons/collections4/IterableUtils.java
@@ -484,6 +484,22 @@ public class IterableUtils {
     }
 
     /**
+     * Returns an {@link Optional} containing the first element in the
+     * {@code iterable}, or an empty {@code Optional} if the iterable is
+     * empty or null.
+     * <p>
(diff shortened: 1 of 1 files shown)


garydgregory (commented):
Hello @spijkercenter 

-1

- This PR does not compile.
- Please validate your submissions to avoid wasting the maintainers' time. Compiling code is the most basic requirement.
- Unfortunately, the PR also has zero unit tests, so it can't be validated.
- This class does not use Optional and I don't think we want to add variants
- Try to use the existing APIs, like get() and first() and use Optional in your code

PR #562, opened 2024-10-18
COLLECTIONS-867: Switched iterator filters to use java.util.function.Predicate

Claudenw:
Replaced usage of commons-collections Predicate with java.util.function.Predicate in iterator filtering.


garydgregory:
Hi @Claudenw 

-1: Please see the email thread from way back in July: https://lists.apache.org/thread/44dr03zs4zr92yypgxgsq84273n5g211

Yep, it's a pain, but Collections predates Java 8 and uses custom functional interfaces.

This would be for the next major release.

Claudenw:
Gary's objection makes sense.  Closing this without merging.

PR #564, opened 2024-10-20
COLLECTIONS-870: Added ExtendedIterator and tests

Claudenw:
Implement COLLECTIONS-870

Adds an ExtendedIterator as defined in COLLECTIONS-870 as well as in the discussion on the mailing list.

mailing list discussion: https://lists.apache.org/thread/rpcsoz3jwz7x9sp3cgzf315ls7pcd90w

### DIFF
--- a/src/main/java/org/apache/commons/collections4/iterators/ExtendedIterator.java
+++ b/src/main/java/org/apache/commons/collections4/iterators/ExtendedIterator.java
@@ -26,18 +26,17 @@ import java.util.stream.Stream;
 
 
 /**
- * A ExtendedIterator is an Iterator wrapping around a plain
- * (or presented as plain) Iterator. The wrapping allows the usual
- * operations found on streams (filtering, concatenating, mapping) to be done on an Iterator derived
- * from some other source.  It also provides convenience methods for common operations.
+ * Extends Iterator functionality to include operations commonly found on streams (e.g. filtering, concatenating, mapping).
+ * It also provides convenience methods for common operations.
  * @param <T> The type of object returned from the iterator.
+ * @since 4.5.0-M3
  */
 public final class ExtendedIterator<T> implements Iterator<T> {
     /**
      * Set to <code>true</code> if this wrapping doesn't permit the use of
      * {@link #remove()}, otherwise removal is delegated to the base iterator.
      */
-    private final boolean removeDenied;
+    private final boolean throwOnRemove;
 
     /**
      * Creates an ExtendedIterator wrapped round <code>it</code>,
@@ -109,11 +108,11 @@ public final class ExtendedIterator<T> implements Iterator<T> {
     /**
      * Initialise this wrapping with the given base iterator and remove-control.
      * @param base the base iterator that this iterator wraps
-     * @param removeDenied true if .remove() must throw an exception
+     * @param throwOnRemove true if .remove() must throw an exception
      */
-    private ExtendedIterator(final Iterator<? extends T> base, final boolean removeDenied) {
+    private ExtendedIterator(final Iterator<? extends T> base, final boolean throwOnRemove) {
         this.base = base;
-        this.removeDenied = removeDenied;
+        this.throwOnRemove = throwOnRemove;
     }
 
     @Override
@@ -133,7 +132,7 @@ public final class ExtendedIterator<T> implements Iterator<T> {
 
     @Override
     public void remove() {
-        if (removeDenied) {
+        if (throwOnRemove) {
             throw new UnsupportedOperationException();
         }
         base.remove();
@@ -160,7 +159,7 @@ public final class ExtendedIterator<T> implements Iterator<T> {
             ((IteratorChain<T>) base).addIterator(other);
             return this;
         }
-        return new ExtendedIterator<T>(new IteratorChain<T>(this.base, other), this.removeDenied);
+        return new ExtendedIterator<T>(new IteratorChain<T>(this.base, other), this.throwOnRemove);
     }
 
     /**
@@ -170,7 +169,7 @@ public final class ExtendedIterator<T> implements Iterator<T> {
      * @return An iterator filtered by the predicate.
      */
     public ExtendedIterator<T> filter(final Predicate<T> predicate) {
-        return new ExtendedIterator<T>(new FilterIterator<>(this, predicate::test), this.removeDenied);
+        return new ExtendedIterator<T>(new FilterIterator<>(this, predicate::test), this.throwOnRemove);
     }
 
     /**
--- a/src/main/java/org/apache/commons/collections4/iterators/ExtendedIterator.java
+++ b/src/main/java/org/apache/commons/collections4/iterators/ExtendedIterator.java
@@ -1,0 +1,197 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.collections4.iterators;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+
+
+/**
+ * A ExtendedIterator is an Iterator wrapping around a plain
+ * (or presented as plain) Iterator. The wrapping allows the usual
+ * operations found on streams (filtering, concatenating, mapping) to be done on an Iterator derived
+ * from some other source.  It also provides convenience methods for common operations.
+ * @param <T> The type of object returned from the iterator.
+ */
+public final class ExtendedIterator<T> implements Iterator<T> {
+    /**
+     * Set to <code>true</code> if this wrapping doesn't permit the use of
+     * {@link #remove()}, otherwise removal is delegated to the base iterator.
+     */
+    private final boolean removeDenied;
+
+    /**
+     * Creates an ExtendedIterator wrapped round <code>it</code>,
+     * which does not permit <code>.remove()</code>
+     * even if <code>it</code> does.
+     * @param it The Iterator to wrap.
+     * @return an Extended iterator on {@code it}
+     * @throws UnsupportedOperationException if remove() is called on the resulting iterator.
+     */
+    public static <T> ExtendedIterator<T> createNoRemove(final Iterator<T> it) {
+        return new ExtendedIterator<>(it, true);
+    }
+
+    /**
+     * Creates an ExtendedIterator wrapped round a {@link Stream}.
+     * The extended iterator does not permit <code>.remove()</code>.
+     * <p>
+     * The stream should not be used directly. The effect of doing so is
+     * undefined.
+     * </p>
+     * @param stream the Stream to create an iterator from.
+     * @return an Extended iterator on the {@code stream} iterator.
+     */
+    public static <T> ExtendedIterator<T> create(final Stream<T> stream) {
+        return new ExtendedIterator<T>(stream.iterator(), true);
+    }
+
+    /**
+     * Flattens an iterator of iterators into an Iterator over the next level values.
+     * Similar to list splicing in lisp.
+     * @param iterators An iterator of iterators.
+     * @return An iterator over the logical concatenation of the inner iterators.
+     */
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/collections4/iterators/ExtendedIterator.java
```diff
@@ -0,0 +1,197 @@
...
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+
+
+/**
+ * A ExtendedIterator is an Iterator wrapping around a plain
```
garydgregory, line 29: I would either fix "A" -> "An" or just say "Extends functionality to do ..."

src/main/java/org/apache/commons/collections4/iterators/ExtendedIterator.java
```diff
@@ -0,0 +1,197 @@
...
+ * (or presented as plain) Iterator. The wrapping allows the usual
+ * operations found on streams (filtering, concatenating, mapping) to be done on an Iterator derived
+ * from some other source.  It also provides convenience methods for common operations.
+ * @param <T> The type of object returned from the iterator.
+ */
```
garydgregory, line 34: Add `@since 4.5.0-M3`.

src/main/java/org/apache/commons/collections4/iterators/ExtendedIterator.java
```diff
@@ -0,0 +1,197 @@
...
+public final class ExtendedIterator<T> implements Iterator<T> {
+    /**
+     * Set to <code>true</code> if this wrapping doesn't permit the use of
+     * {@link #remove()}, otherwise removal is delegated to the base iterator.
+     */
+    private final boolean removeDenied;
```
garydgregory, line 40: IMO "Denied" could imply noop, so I would be explicit and call it "throwOnRemove".

src/main/java/org/apache/commons/collections4/iterators/ExtendedIterator.java
```diff
@@ -0,0 +1,197 @@
...
+     * @return An Extended iterator wrapping {@code it}
+     */
+    public static <T> ExtendedIterator<T> create(final Iterator<T> it) {
+        return it instanceof ExtendedIterator<?>
+                ? (ExtendedIterator<T>) it
+                : new ExtendedIterator<>(it, false);
+    }
+
+    /** the base iterator that we wrap */
+    private final Iterator<? extends T> base;
```
garydgregory, line 107: Extend `FilterIterator` and you don't need this and most other delegating methods.

src/main/java/org/apache/commons/collections4/iterators/ExtendedIterator.java
```diff
@@ -0,0 +1,197 @@
...
+        return new ExtendedIterator<T>(new IteratorChain<T>(this.base, other), this.removeDenied);
+    }
+
+    /**
+     * Filter this iterator using a predicate.  Only items for which the predicate returns {@code true} will
+     * be included in the result.
+     * @param predicate The predicate to filter the items with.
+     * @return An iterator filtered by the predicate.
+     */
+    public ExtendedIterator<T> filter(final Predicate<T> predicate) {
```
garydgregory, line 172: Extend `FilterIterator` and you don't need this.

garydgregory (changes requested):
Hi @Claudenw 

This looks good but I would extend `FilterIterator` instead of `Iterator`. I've added a few other comments.

src/main/java/org/apache/commons/collections4/iterators/ExtendedIterator.java
Claudenw: Extending `FilterIterator` exposes `setIterator()` and `setPredicate()` and changes the semantics of the class.
If we do this then why not expose `setFunction()` and something else for `andThen()`.

I think the semantics are cleaner if we do not extend FilterIterator but use it instead.

src/main/java/org/apache/commons/collections4/iterators/ExtendedIterator.java
Claudenw: changed

src/main/java/org/apache/commons/collections4/iterators/ExtendedIterator.java
Claudenw: see above.

src/main/java/org/apache/commons/collections4/iterators/ExtendedIterator.java
garydgregory: `setFunction()` ? This class does not track a function that I can see. 

There is just too much duplication with `FilterIterator` to avoid subclassing IMO. Having 2 almost identical classes feels confusing IMO.  

I could even see initially creating a subclass and then considering pulling up the functionality into FilterIterator. The cuurent FilterIterator already does more than a plain "Filter" in the traditional `Filter[Input|Output]Stream` model (The same model we follow Commons IO where we provide subclasses to `Filter[Input|Output]Stream` (most) extending `Proxy[Input|Output]Stream`.)

As an aside, ideally, the FilterIterator class should have been immutable, or that's what I would do if I were to write one today, but that's beside the point IMO. I say this as I do not think we want to add more mutable state.

Claudenw: This class does not track a function.  Extended iterator only adds the ability to force an iterator to not support delete.  All other methods simply create instances of other collection iterator classes.  Am I missing something here?

garydgregory: Hi @Claudenw 
You wrote:
> If we do this then why not expose setFunction() and something else for andThen().

Which is why I asked about `setFunction()`. Maybe I'm the one missing something? 🤔

Claudenw: OK.  Let me start again.

EnhancedIterator is an Iterator that adds Streaming like functionality (e.g. filter, map, etc.). 

By that definition it should have methods like `filter(Predicate)`, `map(Function)`, etc.   (Note Predicate and Function are the java classes not the org.apache.commons based ones).
I would not expect it to have methods like `setFilter(Predicate)` or `setFunction(Function)`

If EnhancedIterator extends FilterIterator it exposes 2 undesirable methods.  `setPredicate(Predicate)` (apache commons based class) and `setIterator(Iterator)`. 

EnhancedIterator does not really extend FilterIterator in concept, and adding FilterIterator as a super class to remove a few lines of code is a bad idea as it just muddies the water.

Claudenw:
I responded to this on github but it does not seem to have come through
here.

src/test/java/org/apache/commons/collections4/iterators/ExtendedIteratorTest.java
```diff
@@ -0,0 +1,150 @@
...
+import java.util.Arrays;
+import java.util.Iterator;
+import java.util.List;
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class ExtendedIteratorTest {
```
garydgregory, line 34: I just realized this test should extent `AbstractIteratorTest`, especially now that I added testing `forEachRemaining()`.

garydgregory (changes requested):
Hi @Claudenw 
Please see my test comment. Also rebase on master. I also see we have `ObjectGraphIterator`. I will experiement more this week.
TY!

src/main/java/org/apache/commons/collections4/iterators/ExtendedIterator.java
```diff
@@ -0,0 +1,196 @@
...
+
+    /**
+     * Creates an ExtendedIterator wrapped round <code>it</code>,
+     * which does not permit <code>.remove()</code>
+     * even if <code>it</code> does.
+     * @param it The Iterator to wrap.
+     * @return an Extended iterator on {@code it}
+     * @throws UnsupportedOperationException if remove() is called on the resulting iterator.
+     */
+    public static <T> ExtendedIterator<T> createNoRemove(final Iterator<T> it) {
```
garydgregory, line 49: Missing Javadoc for `T`.

src/main/java/org/apache/commons/collections4/iterators/ExtendedIterator.java
```diff
@@ -0,0 +1,196 @@
...
+     * Creates an ExtendedIterator wrapped round a {@link Stream}.
+     * The extended iterator does not permit <code>.remove()</code>.
+     * <p>
+     * The stream should not be used directly. The effect of doing so is
+     * undefined.
+     * </p>
+     * @param stream the Stream to create an iterator from.
+     * @return an Extended iterator on the {@code stream} iterator.
+     */
+    public static <T> ExtendedIterator<T> create(final Stream<T> stream) {
```
garydgregory, line 63: Missing Javadoc for `T`.

src/main/java/org/apache/commons/collections4/iterators/ExtendedIterator.java
```diff
@@ -0,0 +1,196 @@
...
+        return new ExtendedIterator<T>(stream.iterator(), true);
+    }
+
+    /**
+     * Flattens an iterator of iterators into an Iterator over the next level values.
+     * Similar to list splicing in lisp.
+     * @param iterators An iterator of iterators.
+     * @return An iterator over the logical concatenation of the inner iterators.
+     */
+    public static <T> ExtendedIterator<T> flatten(final Iterator<Iterator<T>> iterators) {
```
garydgregory, line 73: Missing Javadoc for `T`.

src/main/java/org/apache/commons/collections4/iterators/ExtendedIterator.java
```diff
@@ -0,0 +1,196 @@
...
+            }
+
+        });
+    }
+
+    /**
+     * Creates an empty Extended iterator.
+     * @return An empty Extended iterator.
+     */
+    public static ExtendedIterator<?> emptyIterator() {
```
garydgregory, line 88: Needs generics IMO (not `?`).

src/main/java/org/apache/commons/collections4/iterators/ExtendedIterator.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    /**
+     * Create an ExtendedIterator returning the elements of <code>it</code>.
+     * If <code>it</code> is itself an ExtendedIterator, return that;
+     * otherwise wrap <code>it</code>.
+     * @param it The iterator to wrap.
+     * @return An Extended iterator wrapping {@code it}
+     */
+    public static <T> ExtendedIterator<T> create(final Iterator<T> it) {
```
garydgregory, line 99: Missing Javadoc for `T`.

garydgregory (changes requested):
Hi @Claudenw 
There are missing Javadoc for `T`.

src/main/java/org/apache/commons/collections4/iterators/ExtendedIterator.java
Claudenw: @garydgregory  Did you see my comment above?

src/main/java/org/apache/commons/collections4/iterators/ExtendedIterator.java
garydgregory: Hi @Claudenw 
If you rebase on git master, you can say:
```java
return create(IteratorUtils.chainedIterator(iterators));
```
instead of:
```
        return create(new LazyIteratorChain<T>() {

            @Override
            protected Iterator<? extends T> nextIterator(final int count) {
                return iterators.hasNext() ? iterators.next() : null;
            }

        });
```

src/main/java/org/apache/commons/collections4/iterators/ExtendedIterator.java
garydgregory: Hi @Claudenw,

> EnhancedIterator is an Iterator that adds Streaming like functionality

This is causing me to pause and think that adding streaming-like functionality here is the wrong way to go. Instead of adding a 2 or 3 methods now and who knows how many later, and considering that the `Stream` interface and functionality is large and rich, so instead of inventing a new vocabulary here, why not just use a `Stream`? We could instead add:

- org.apache.commons.collections4.IteratorUtils.stream(Iterable<E>)
- org.apache.commons.collections4.IteratorUtils.stream(Iterator<E>)

This provides a bridge to the full Stream functionality, which everyone is familiar with, no need to add and learn new custom methods like "addTo", "map", and so on.


Details: Something like
```java
    /**
     * Creates a stream on the given Iterable.
     *
     * @param <E> the type of elements in the Iterable.
     * @param iterable the Iterable to stream or null.
     * @return a new Stream or {@link Stream#empty()} if the Iterable is null.
     * @since 4.5.0-M3
     */
    public static <E> Stream<E> stream(final Iterable<E> iterable) {
        return iterable == null ? Stream.empty() : StreamSupport.stream(iterable.spliterator(), false);
    }

    /**
     * Creates a stream on the given Iterator.
     *
     * @param <E> the type of elements in the Iterator.
     * @param iterator the Iterator to stream or null.
     * @return a new Stream or {@link Stream#empty()} if the Iterator is null.
     * @since 4.5.0-M3
     */
    public static <E> Stream<E> stream(final Iterator<E> iterator) {
        return iterator == null ? Stream.empty() : StreamSupport.stream(Spliterators.spliteratorUnknownSize(iterator, Spliterator.ORDERED), false);
    }
```

PR #581, opened 2024-12-08
Use Java Generics for `MultiKeyMap`-keys as method parameters

Chrimle:

- [ ] Read the [contribution guidelines](CONTRIBUTING.md) for this project.
- [ ] Run a successful build using the default [Maven](https://maven.apache.org/) goal with `mvn`; that's `mvn` on the command line by itself.
- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This may not always be possible but is a best-practice.
- [ ] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [ ] Each commit in the pull request should have a meaningful subject line and body. Note that commits might be squashed by a maintainer on merge.

> This is purely a draft. The suggestion made by @garydgregory was to simply update the JavaDocs.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/map/MultiKeyMap.java
+++ b/src/main/java/org/apache/commons/collections4/map/MultiKeyMap.java
@@ -162,7 +162,7 @@ public class MultiKeyMap<K, V> extends AbstractMapDecorator<MultiKey<? extends K
      * @param key2  the second key
      * @return true if the map contains the key
      */
-    public boolean containsKey(final Object key1, final Object key2) {
+    public boolean containsKey(final K key1, final K key2) {
         final int hashCode = hash(key1, key2);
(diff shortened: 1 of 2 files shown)


garydgregory:
@Chrimle 
See item 2 in the checklist.

Chrimle:
I've assessed this, and I would say that the keys should not be of type `K`. The reason, is that there is ambiguity whether all keys should of the same type or not. Regardless, I believe the better solution would allow any combination of types for the keys. Hence, if generics should be used - it wouldn't be done in this way.

These changes will be reverted, and instead, all JavaDocs in `MultiKeyMap` will be assessed.

@garydgregory this was just a draft. In hindsight, it could've been shared on Jira instead.

PR #630, opened 2025-08-01
[collections-838] adds the conviences methods unionSet(), intersectio…

JoergBudi:
…nSet(), differenceSet(), disjunctionSet() to SetUtils to emphasize the difference between SetView and SetView.toSet() return values. Adapts javadoc for the existing methods

One can argue, that adding the 4 methods with one codeline each adds little value and bloats the API. 

But note, that there are multiple bugs (at least 838, 770, but I saw one more I can't find atm) where the main reason is, that the reporter use the 4 original methods without invoking toSet() inbetween, so it is a clear source coding errors that may also be found later on (as only chaining the results more often than twenty times leads to significant performance problems due to exponential growth). Thus offering these 4 additional methods side by side with the original ones gives the API user a better understanding on the differences and which one he might use under given circumstances than just improving the toSet() javadoc.

I shortly thought about reimplemting the methods with plain Java API (addAll, removeAll, etc.), but these are not mandatory for sets to be implemented, so reusing the existing 4 methods seems the better solution, so that they will always work.

@garydgregory : You might have a look at this PR when it is suitable for you.

- [-] Read the [ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html) if you use Artificial Intelligence (AI).
- [-] I used AI to create any part of, or all of, this pull request.
- [x] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This may not always be possible, but it is a best-practice.
- [ x Each commit in the pull request should have a meaningful subject line and body. Note that a maintainer may squash commits during the merge process.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/SetUtils.java
+++ b/src/main/java/org/apache/commons/collections4/SetUtils.java
@@ -117,10 +117,12 @@ public class SetUtils {
 
     /**
      * Returns an unmodifiable <strong>view</strong> containing the difference of the given
-     * {@link Set}s, denoted by {@code a \ b} (or {@code a - b}).
+     * {@link Set}s, denoted by {@code setA \ setB} (or {@code setA - setBb}).
      * <p>
-     * The returned view contains all elements of {@code a} that are not a member
-     * of {@code b}.
+     * The returned view at all times contains all elements of {@code setA} that are not a member
+     * of {@code setB} also when {@code setA} and {@code setB} change later on.
+     * ; in case you don't need this view behaviour and you want to subtract the sets only once,
+     * use {@link SetUtils.SetView#toSet} on the return value or use directly {@link SetUtils#differenceSet} instead.
      * </p>
      *
      * @param <E> the generic type that is able to represent the types contained
@@ -150,14 +152,36 @@ public class SetUtils {
     }
 
     /**
+     * Returns a set of the difference of the given
+     * {@link Set}s, denoted by {@code setA \ setB} (or {@code setA - setB}).
+     * <p>
+     * The returned set contains all elements of {@code setA} that are not a member
+     * of {@code setB} at the moment of invocation. Later changes to {@code setA}
+     * or {@code setB} do not affect the returned set.
+     * Convenience method for difference(...).toSet().
+     * </p>
+     *
+     * @param <E> the generic type that is able to represent the types contained
+     *   in both input sets.
+     * @param setA  the set to subtract from, must not be null
+     * @param setB  the set to subtract, must not be null
+     * @return a new set with the difference of the given {@link Set}s
+     * @since 4.5.1
+     */
+    public static <E> Set<E> differenceSet(final Set<? extends E> setA, final Set<? extends E> setB) {
+        return difference(setA, setB).toSet();
+    }
+
+    /**
      * Returns an unmodifiable <strong>view</strong> of the symmetric difference of the given
      * {@link Set}s.
      * <p>
-     * The returned view contains all elements of {@code a} and {@code b} that are
-     * not a member of the other set.
-     * </p>
+     * The returned view at all times contains all elements of {@code setA} and {@code setB} that are
+     * not a member of the other set also when {@code setA} and {@code setB} change later on; in case you don't
+     * need this view behaviour and you want to disjunct the sets only once,
+     * use {@link SetUtils.SetView#toSet} on the return value or use directly {@link SetUtils#disjunctionSet} instead.
      * <p>
-     * This is equivalent to {@code union(difference(a, b), difference(b, a))}.
+     * This is equivalent to {@code union(difference(setA, setB), difference(setB, setA))}.
      * </p>
      *
      * @param <E> the generic type that is able to represent the types contained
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/collections4/SetUtils.java
```diff
@@ -149,15 +151,37 @@ public Iterator<E> createIterator() {
...
+     * or {@code setB} do not affect the returned set.
+     * Convenience method for difference(...).toSet().
+     * </p>
+     *
+     * @param <E> the generic type that is able to represent the types contained
+     *   in both input sets.
+     * @param setA  the set to subtract from, must not be null
+     * @param setB  the set to subtract, must not be null
+     * @return a new set with the difference of the given {@link Set}s
+     * @since 4.5.1
```
garydgregory, line 169: Per semantic versioning, this would be for 4.6.0; 4.5.1 would be a bug fix release only.

garydgregory:
Hello @ppkarwasz @chtompki @henrib @kinow ,

Do you guys have any thoughts on the utility of these new APIs versus increasing the public footprint of this class?

ppkarwasz:
@JoergBudi, I see your point, but I believe the root of the issue you're highlighting is a misunderstanding around how Java Collections typically behave, particularly the distinction between **views** and **copies**.

Many methods in the core Java Collections API return **views**, not copies. Here's a non-exhaustive list of such methods:

**List:**

* `List.subList(int fromIndex, int toIndex)`

**Map:**

* `Map.keySet()`
* `Map.values()`
* `Map.entrySet()`

**SortedSet / NavigableSet:**

* `SortedSet.subSet(E fromElement, E toElement)`
* `SortedSet.headSet(E toElement)`
* `SortedSet.tailSet(E fromElement)`
* `NavigableSet.descendingSet()`

**SortedMap / NavigableMap:**

* `SortedMap.subMap(K fromKey, K toKey)`
* `SortedMap.headMap(K toKey)`
* `SortedMap.tailMap(K fromKey)`
* `NavigableMap.descendingMap()`

The `SetView` in Commons Collections follows this same model.

While it's true that some bugs (e.g. COLLECTIONS-838, COLLECTIONS-770) have been caused by users neglecting to call `toSet()` on `SetView`, this highlights a documentation or education gap, not necessarily an API deficiency. Adding four one-liner convenience methods like `unionSet()`, `intersectionSet()`, etc., may be potentially confusing for users (e.g., “What’s the difference between `union()` and `unionSet()`?”).

JoergBudi:
@ppkarwasz: Thx for your feeback, my thoughts:
The JDK samples don't invite a nested/chained usage as much as the 4 functions in SetUtils do.
It is also hard to explain, that 5 chainings are fine, 40 are disastrous, and everything
in between is just unexpectingly expensive.
For many usecases, from a logical point of view, it doesn't matter if views or intermediate sets are used,
and as developers tend to lazyness, it will be hard to convince them in javadoc to use
the additional toSet() method when for low number of iterations everything works fine. 

Seeing the methods side-by-side, the API user is forced to think which method to use and is
lured to check the javadoc of both methods, and that is a good thing here (when I have a look
at the javadoc on this branch). 

But I understand, that one can come to a different conclusion.
I am happy to transform this PR to a javadoc-only-PR, but a short guidance
would then be good. I would suggest, to update all 4 methods with paragraph similar to this:


Note: Chaining/Nesting the SetView result of a union() invocation as input into union(), difference(), disjunction(), intersection() 
      multiple times becomes expensive, to avoid this, use SetView#toSet() on each intermediate result when it 
      still suits your usecase.


Would this be ok ?

I got one thing to add:

In case collections-722 (PR 628 which improves chaining the chainedIterator) gets merged,
chaining via SetUtils.union( chainedSet, someOtherSet) as it is in Testcase of this collections-838, is fast even
for a high number of chainings.
However doing it the other way round (SetUtils.union(someOtherSet, chainedSet)) suffers again from
exponential growth, as I can't think of a good solution to optimize IteratorChain+FilterIterator +
the 4 SetView implementations and the various predicates.
This makes the expected chaining runtime behaviour even more challenging to properly document,
as some invocations patterns don't lead to exponential growth.

ppkarwasz:
Hi @JoergBudi,

> I am happy to transform this PR to a javadoc-only-PR, but a short guidance would then be good. I would suggest, to update all 4 methods with paragraph similar to this:
>
> Note: Chaining/Nesting the SetView result of a union() invocation as input into union(), difference(), disjunction(), intersection() multiple times becomes expensive, to avoid this, use SetView#toSet() on each intermediate result when it still suits your usecase.
> 
> Would this be ok ?

I’ve followed your comments on [COLLECTIONS-838](https://issues.apache.org/jira/browse/COLLECTIONS-838) and added an explanation of the root cause of the exponential performance degradation.

While improving the Javadoc is always welcome, I think the more impactful next step is to fix the underlying issue first. My proposal:

* **Close this PR** instead of converting it to a Javadoc-only change.
* **Address the exponential performance** in `IteratorChain.hasNext()` when wrapping `IteratorChain` instances, in a dedicated PR.
* **Revisit `SetUtils` Javadoc** after that fix lands, to see if additional guidance is still needed.

How does that sound? If you’re willing to take on the `IteratorChain` work, I can assign you to the JIRA issue.

PR #668, opened 2025-12-25
[COLLECTIONS-880] Add MultiMapUtils.invert(Map, MultiValuedMap) overload

clapppp:
This PR overloads MultiMapUtils.invert to support java.util.Map as an input.

Changes:
- Added invert(Map, MultiValuedMap) method to MultiMapUtils.
- Added unit tests for the new method testInvertMap
- Imported java.util.Map and java.util.HashMap
- Renamed testInvert into testInvertMultiValuedMap

[COLLECTIONS-880](https://issues.apache.org/jira/browse/COLLECTIONS-880)

### DIFF
--- a/src/test/java/org/apache/commons/collections4/MultiMapUtilsTest.java
+++ b/src/test/java/org/apache/commons/collections4/MultiMapUtilsTest.java
@@ -24,8 +24,10 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 import org.apache.commons.collections4.multimap.ArrayListValuedHashMap;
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/commons/collections4/MultiMapUtilsTest.java
```diff
@@ -142,6 +144,30 @@ void testInvert() {
...
+
+        final Map<String, String> deps = new HashMap<>();
+        deps.put("commons-configuration1", "commons-logging");
+        deps.put("commons-configuration2", "commons-lang3");
+        deps.put("commons-configuration3", "commons-text");
+        deps.put("commons-beanutils1", "commons-collections");
+        deps.put("commons-beanutils2", "commons-logging");
+        MultiMapUtils.invert(deps, usages);
+        final Set<String> loggingUsagesCompile = usages.get("commons-logging");
+        assertEquals("[commons-beanutils2, commons-configuration1]", loggingUsagesCompile.toString());
```
ytatichno, line 159: Minor observation: I’m not sure about relying on the set’s iteration order here (and other asserts in this test with sets casted `toString()`). `HashSetValuedHashMap.get()` returns a `Set<String>`, but `WrappedSet`/`HashSet` doesn’t guarantee order, so `toString()` output could vary across JVMs or even runs.

garydgregory:
I don't think this PR is needed anymore due to https://github.com/apache/commons-collections/blob/f24800fa4565482581053a3d7b767ff84ee44a82/src/main/java/org/apache/commons/collections4/MultiMapUtils.java#L146

Please confirm your use case.

PR #669, opened 2025-12-25
[COLLECTIONS-881] MultiMapUtils.getXXX ensure safe copy

clapppp:
This PR modifies getXXX method to return new Collection.

Updated getValuesAsSet, getValuesAsList, and getValuesAsBag to always return a new copy.
Added unit tests to verify that returned collections are independent of the backing map.
Updated Javadocs to reflect that a new collection is returned.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/MultiMapUtils.java
+++ b/src/main/java/org/apache/commons/collections4/MultiMapUtils.java
@@ -102,14 +102,11 @@ public class MultiMapUtils {
      * @param <V> the value type
      * @param map  the {@link MultiValuedMap} to use
      * @param key  the key to look up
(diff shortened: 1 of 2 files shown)


garydgregory (changes requested):
@clapppp 

Thank you for your PR.

Please see my comments.

clapppp:
@garydgregory 

Thank you for the feedback!

I reviewed my test cases and found the issues.
I updated the test for `getValuesAsSet` accordingly.

However, I couldn't find any `MultiValuedMap`implementation whose `get(key)`returns a `Bag`(Only Collections/List/Set).
So I'm not sure how to write a failing test case.
If it's okay to make test case with Mockito I'll make it!

clapppp:
Hi @garydgregory 

I made the changes as you suggested.
Could you please review the latest changes?

PR #678, opened 2026-06-05
Validate deserialized size in CircularFifoQueue.readObject

digi-scrypt:
1. maxElements is reconstructed from the stream and the backing array is sized to it, but the element count read with readInt() is never checked against it.
2. a count greater than maxElements writes past elements[] (ArrayIndexOutOfBoundsException), and a negative count yields a queue whose start/end are inconsistent so size() reports elements that were never read.

What happens with a tampered stream: have we considered that maxElements is final and trusted while the count beside it is not? Validated both (maxElements >= 1, 0 <= size <= maxElements) in readObject and throw InvalidObjectException, which keeps the check at the only layer that sees the raw stream.


garydgregory:
TY @digi-scrypt , merged 🚀

PR #679, opened 2026-06-13
Reject non-positive entry count in bag/multiset doReadObject

rootvector2:
`AbstractMapBag.doReadObject` and `AbstractMapMultiSet.doReadObject` read each entry's `count` with `readInt()` and store it straight into the backing map (`map.put(obj, new MutableInteger(count))`) with no lower bound, unlike the public API where `add` only stores when `nCopies > 0` and `setCount` rejects `count < 0`. Found while auditing the custom `readObject` paths: a serialized `HashBag`/`TreeBag`/`HashMultiSet` whose entry count is patched to `0` or a negative value deserializes into a collection that reports a negative `size()` and `getCount()`, a state unreachable through the API. Both overrides now reject `count < 1` with `InvalidObjectException`, matching `doWriteObject` which only ever writes existing (>= 1) entries. Regression tests added in `HashBagTest` and `HashMultiSetTest`.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/bag/AbstractMapBag.java
+++ b/src/main/java/org/apache/commons/collections4/bag/AbstractMapBag.java
@@ -17,6 +17,7 @@
 package org.apache.commons.collections4.bag;
 
 import java.io.IOException;
+import java.io.InvalidObjectException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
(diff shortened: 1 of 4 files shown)


garydgregory (commented):
@rootvector2 
In the future, do NOT copy-pasta tests.

rootvector2:
> @rootvector2 In the future, do NOT copy-pasta tests.

ok

PR #682, opened 2026-06-15
Re-validate entries in PredicatedMap/PredicatedCollection readObject

rootvector2:
`PredicatedMap.readObject` and `PredicatedCollection.readObject` rebuild the decorated map/collection straight from the stream and never re-run the predicate that the constructors apply to every element. A serialized form whose backing data was not produced through `put`/`add` (a tampered or hand-built stream) therefore deserializes into a decorator that holds elements its own predicate rejects. A `PredicatedMap` created with `NotNullPredicate` to forbid null keys, for instance, can be made to contain a null key, and code that trusts the decorator's guarantee then breaks.

I found this while checking the decorators' `readObject` paths against the invariant their constructors enforce (`map.forEach(this::validate)` / the per-element loop).

The fix re-validates each deserialized entry/element against the configured predicate and throws `InvalidObjectException` on a violation, mirroring the constructor. `PredicatedSortedMap` and the `PredicatedList` / `PredicatedSet` / `PredicatedBag` / `PredicatedQueue` / `PredicatedMultiSet` / `PredicatedNavigableSet` decorators inherit the check through these two base classes, so the whole predicated family is covered. The `Transformed*` decorators are intentionally left alone: their stored values are already transformed, so re-running the transformer on read would transform twice.

Existing serialized forms still load (the version-4 compatibility tests pass); only streams carrying predicate-violating data are now rejected.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/collection/PredicatedCollection.java
+++ b/src/main/java/org/apache/commons/collections4/collection/PredicatedCollection.java
@@ -421,8 +421,12 @@ public class PredicatedCollection<E> extends AbstractCollectionDecorator<E> {
      */
     private void readObject(final ObjectInputStream in) throws IOException, ClassNotFoundException {
         in.defaultReadObject();
-        Objects.requireNonNull(decorated(), "collection");
-        Objects.requireNonNull(predicate, "predicate");
+        if (decorated() == null) {
+            throw new InvalidObjectException("Null collection");
+        }
+        if (predicate == null) {
+            throw new InvalidObjectException("Null predicate");
+        }
         try {
             decorated().forEach(this::validate);
         } catch (final IllegalArgumentException ex) {
--- a/src/main/java/org/apache/commons/collections4/collection/PredicatedCollection.java
+++ b/src/main/java/org/apache/commons/collections4/collection/PredicatedCollection.java
@@ -16,6 +16,9 @@
  */
 package org.apache.commons.collections4.collection;
 
+import java.io.IOException;
+import java.io.InvalidObjectException;
+import java.io.ObjectInputStream;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -407,6 +410,22 @@ public class PredicatedCollection<E> extends AbstractCollectionDecorator<E> {
             validate(item);
         }
         return decorated().addAll(coll);
+    }
+
+    /**
+     * Deserializes the collection in using a custom routine.
+     *
+     * @param in  the input stream
+     * @throws IOException if an error occurs while reading from the stream
+     * @throws ClassNotFoundException if an object read from the stream cannot be loaded
+     */
+    private void readObject(final ObjectInputStream in) throws IOException, ClassNotFoundException {
+        in.defaultReadObject();
+        try {
+            decorated().forEach(this::validate);
+        } catch (final IllegalArgumentException ex) {
+            throw new InvalidObjectException(ex.getMessage());
+        }
     }
 
     /**
--- a/src/main/java/org/apache/commons/collections4/collection/PredicatedCollection.java
+++ b/src/main/java/org/apache/commons/collections4/collection/PredicatedCollection.java
@@ -421,10 +421,12 @@ public class PredicatedCollection<E> extends AbstractCollectionDecorator<E> {
      */
     private void readObject(final ObjectInputStream in) throws IOException, ClassNotFoundException {
         in.defaultReadObject();
+        Objects.requireNonNull(decorated(), "collection");
+        Objects.requireNonNull(predicate, "predicate");
         try {
             decorated().forEach(this::validate);
         } catch (final IllegalArgumentException ex) {
-            throw new InvalidObjectException(ex.getMessage());
+            throw (InvalidObjectException) new InvalidObjectException(ex.getMessage()).initCause(ex);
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/commons/collections4/collection/PredicatedCollection.java
```diff
@@ -409,6 +412,22 @@ public boolean addAll(final Collection<? extends E> coll) {
...
+     * @param in  the input stream
+     * @throws IOException if an error occurs while reading from the stream
+     * @throws ClassNotFoundException if an object read from the stream cannot be loaded
+     */
+    private void readObject(final ObjectInputStream in) throws IOException, ClassNotFoundException {
+        in.defaultReadObject();
+        try {
+            decorated().forEach(this::validate);
+        } catch (final IllegalArgumentException ex) {
+            throw new InvalidObjectException(ex.getMessage());
```
garydgregory: Don't loose the exception cause (`java.lang.Throwable.initCause(Throwable)`).

src/main/java/org/apache/commons/collections4/map/PredicatedMap.java
```diff
@@ -155,6 +156,11 @@ public void putAll(final Map<? extends K, ? extends V> mapToCopy) {
     private void readObject(final ObjectInputStream in) throws IOException, ClassNotFoundException {
         in.defaultReadObject();
         map = (Map<K, V>) in.readObject(); // (1)
+        try {
+            map.forEach(this::validate);
+        } catch (final IllegalArgumentException ex) {
+            throw new InvalidObjectException(ex.getMessage());
```
garydgregory: Don't loose the exception cause (`java.lang.Throwable.initCause(Throwable)`).

src/main/java/org/apache/commons/collections4/collection/PredicatedCollection.java
```diff
@@ -409,6 +412,22 @@ public boolean addAll(final Collection<? extends E> coll) {
...
 
+    /**
+     * Deserializes the collection in using a custom routine.
+     *
+     * @param in  the input stream
+     * @throws IOException if an error occurs while reading from the stream
+     * @throws ClassNotFoundException if an object read from the stream cannot be loaded
+     */
+    private void readObject(final ObjectInputStream in) throws IOException, ClassNotFoundException {
+        in.defaultReadObject();
```
garydgregory, line 423: Invariant checks for nulls to match the constructor is also missing.

src/main/java/org/apache/commons/collections4/map/PredicatedMap.java
```diff
@@ -155,6 +156,11 @@ public void putAll(final Map<? extends K, ? extends V> mapToCopy) {
     private void readObject(final ObjectInputStream in) throws IOException, ClassNotFoundException {
         in.defaultReadObject();
         map = (Map<K, V>) in.readObject(); // (1)
+        try {
```
garydgregory, line 162: Invariant checks for nulls to match the constructor is also missing.

garydgregory (changes requested):
Hello @rootvector2 

Thank you for your PR.

Please see my comments.

src/main/java/org/apache/commons/collections4/map/PredicatedMap.java
rootvector2: done, kept the cause with `initCause(ex)` on the `InvalidObjectException`.

src/main/java/org/apache/commons/collections4/collection/PredicatedCollection.java
rootvector2: fixed, it chains the original `IllegalArgumentException` through `initCause` now.

src/main/java/org/apache/commons/collections4/map/PredicatedMap.java
rootvector2: added `Objects.requireNonNull` on the map read back from the stream, matching the `super(map)` guard in the constructor. the key/value predicates can be null by design so those stay unchecked.

src/main/java/org/apache/commons/collections4/collection/PredicatedCollection.java
rootvector2: added `Objects.requireNonNull` on `decorated()` and `predicate` before the loop, same null guards the constructor enforces.

src/main/java/org/apache/commons/collections4/collection/PredicatedCollection.java
```diff
@@ -409,6 +412,24 @@ public boolean addAll(final Collection<? extends E> coll) {
...
+        Objects.requireNonNull(decorated(), "collection");
```
garydgregory: Throw `InvalidObjectException`, not NPE.

garydgregory (changes requested):
Hello @rootvector2 
Please see my comment.

src/main/java/org/apache/commons/collections4/collection/PredicatedCollection.java
rootvector2: done. swapped the `Objects.requireNonNull` calls here and in `PredicatedMap.readObject` for explicit null checks that throw `InvalidObjectException` instead of an NPE.

PR #683, opened 2026-06-17
[COLLECTIONS-889] Fix non-unique IndexedCollection remove

OldTruckDriver:
## Summary

Fix `IndexedCollection.remove(Object)` so non-unique indexes keep remaining values that share the same transformed key.

## Details

Previously, removing one object removed the entire key bucket from the index. For non-unique indexes, this made other collection elements with the same transformed key disappear from indexed lookups even though they were still present in the decorated collection.

This change removes only the matching key/value mapping from the index.

## Tests

- `mvn -q -Dtest=org.apache.commons.collections4.collection.IndexedCollectionTest test`
- `mvn -q`


garydgregory:
Jira ticket is https://issues.apache.org/jira/browse/COLLECTIONS-889

PR #684, opened 2026-06-19
Validate order and uniqueness invariants in decorator readObject()

rootvector2:
`readObject` in `ListOrderedMap`, `ListOrderedSet` and `SetUniqueList` restores the backing collection and the parallel structure that records iteration order / enforces uniqueness (`insertOrder`, `setOrder`, and the dedup `set`) as two independent stream fields and never checks the two agree, unlike the constructors which derive the order/dedup structure from the backing collection so they always match. Found while auditing the custom `readObject` paths against the invariant the constructors hold: a tampered or hand-built stream whose two fields disagree deserializes into a decorator whose state is impossible through the API. A `SetUniqueList` can be made to hold duplicates in its backing list, and a `ListOrderedMap`/`ListOrderedSet` can be made to carry an order list that names keys/elements absent from the backing collection (or repeats one), so `keyList()`/iteration surface phantom or duplicate entries while `size()` reports the backing count. Each `readObject` now re-checks that the order/dedup structure is a duplicate-free match for the backing collection and throws `InvalidObjectException` otherwise, mirroring the constructor contract. Existing serialized forms still load (the version-4 compatibility tests pass); only streams carrying inconsistent data are rejected. Regression tests added in `ListOrderedMapTest`, `ListOrderedSetTest` and `SetUniqueListTest`.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/list/SetUniqueList.java
+++ b/src/main/java/org/apache/commons/collections4/list/SetUniqueList.java
@@ -16,6 +16,9 @@
  */
 package org.apache.commons.collections4.list;
 
+import java.io.IOException;
+import java.io.InvalidObjectException;
+import java.io.ObjectInputStream;
 import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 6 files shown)


garydgregory:
TY @rootvector2 , merged 🚀 
BUT... in the future, less copy-pasta please, see 6742fee39960967a679468e118e4185dbc9aeb2a

rootvector2:
noted. should've reused the existing `serializeDeserialize` helper instead of inlining the stream boilerplate in each test. will keep them dry going forward.

PR #685, opened 2026-06-19
[COLLECTIONS-890] Fix TransformIterator null transformer handling

OldTruckDriver:
[COLLECTIONS-890] Fix TransformIterator null transformer handling

TransformIterator documents null transformers as pass-through/no-op, but transform(...) unconditionally called transformer.apply(...). As a result, next() threw NullPointerException when the iterator was constructed with a null transformer or when setTransformer(null) was used.

This change makes TransformIterator return the original element when the transformer is null, matching the documented contract.

Tests cover both constructor-supplied null transformers and setTransformer(null).

Tests run:
- mvn -q -Dtest=org.apache.commons.collections4.iterators.TransformIteratorTest test
- mvn -q '-Dtest=org.apache.commons.collections4.iterators.*Test' test
- mvn -q


garydgregory (changes requested):
@OldTruckDriver 

Please update the Javadoc to reflect that null is a valid transformer.

OldTruckDriver:
Updated the Javadoc. The constructor/setTransformer @param and getTransformer @return now state the transformer may be null (pass-through). Thanks!

garydgregory:
TY @OldTruckDriver, merged with additional changes 🚀

PR #687, opened 2026-06-19
[COLLECTIONS-891] Fix IndexedCollection contains equality

OldTruckDriver:
[COLLECTIONS-891] Fix IndexedCollection contains equality

IndexedCollection#contains(Object) used the transformed key only, so it could return true for an object that was never added when another object mapped to the same key.

This change keeps the indexed lookup as a fast candidate filter, then checks object equality within the indexed values before returning true.

Tests cover both unique and non-unique indexed collections where "01" and "1" map to the same key but are not equal objects.

Tests run:
- mvn -q -Dtest=org.apache.commons.collections4.collection.IndexedCollectionTest#testContainsUsesObjectEqualityNotOnlyTransformedKey test
- mvn -q -Dtest=org.apache.commons.collections4.collection.IndexedCollectionTest#testContainsUsesObjectEqualityWithNonUniqueIndex test
- mvn -q -Dtest=org.apache.commons.collections4.collection.IndexedCollectionTest test
- mvn -q


garydgregory:
I've converted this PR to a Draft PR until we resolve the discussing in [COLLECTIONS-891](https://issues.apache.org/jira/browse/COLLECTIONS-891).

PR #688, opened 2026-06-19
[COLLECTIONS-892] Fix unique IndexedCollection add atomicity

OldTruckDriver:
[COLLECTIONS-892] Fix unique IndexedCollection add atomicity

IndexedCollection#add(Object) updated the decorated collection before checking the unique index constraint. If a duplicate transformed key was rejected, the exception was thrown after the decorated collection had already been mutated, leaving the collection and index out of sync.

This change checks duplicate keys before mutating the decorated collection for unique indexed collections. Non-unique indexed collections keep the existing add path.

Tests cover a failed unique add where "01" and "1" map to the same key, and verify that the decorated collection and index still contain only the original value.

Tests run:
- mvn -q -Dtest=org.apache.commons.collections4.collection.IndexedCollectionTest#testFailedUniqueAddDoesNotPolluteDecoratedCollection test
- mvn -q -Dtest=org.apache.commons.collections4.collection.IndexedCollectionTest test
- mvn -q
- git diff --check


garydgregory:
Jira ticket is [COLLECTIONS-892](https://issues.apache.org/jira/browse/COLLECTIONS-892).

garydgregory:
Hello @OldTruckDriver 
In the future, please consider simpler, less verbose code like 3cb647e7eadba3f9d6dcb8a566d52f78134061e9

PR #690, opened 2026-06-20
Reject non-positive maxSize in LRUMap.doReadObject()

rootvector2:
found auditing the map package readObject paths: `LRUMap.doReadObject` reads `maxSize` with `readInt()` but skips the constructor's `maxSize >= 1` check, so a crafted stream yields a map with a non-positive bound that no constructor allows and that corrupts the eviction linked list on the next `put`; this re-applies the check and throws `InvalidObjectException`.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/map/LRUMap.java
+++ b/src/main/java/org/apache/commons/collections4/map/LRUMap.java
@@ -17,6 +17,7 @@
 package org.apache.commons.collections4.map;
 
 import java.io.IOException;
+import java.io.InvalidObjectException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.Serializable;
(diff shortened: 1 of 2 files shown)


garydgregory (changes requested):
Hello @rootvector2 

What about the `initialSize` constraint?

rootvector2:
`initialSize` isn't part of the serialized form, so there's nothing to re-validate on read. it's a constructor-only param (the proposed table capacity); what gets written is the resulting `data.length`, which already exceeds `maxSize` for normal small maps. `new LRUMap<>(3)` rounds the table up to a power of two, so it serializes a capacity of 4 against a `maxSize` of 3. a `capacity > maxSize` check on read would reject those valid streams.

the constructor's `initialSize > maxSize` guard only constrains the proposed capacity at build time, and a deserialized table that differs from `maxSize` is harmless since entries are re-inserted through `put()`. `maxSize < 1` is the only reconstructable value that corrupts state: it reuses the header sentinel as a data entry on the first `put`, which is what this rejects.

garydgregory:
Thank you for checking @rootvector2 , merged 🚀

PR #691, opened 2026-06-21
Reject invalid load factor in hashed/reference map doReadObject

rootvector2:
found auditing the map package deserialization paths: the `AbstractHashedMap` and `AbstractReferenceMap` constructors reject a `loadFactor` that is non-positive or `NaN`, but their `doReadObject` overrides read it straight from the stream, so a crafted serialized map deserializes with a `loadFactor` no constructor allows and a `threshold` of 0 that makes `checkCapacity` resize the bucket array on every `put` (a 12-entry map grows its backing array from 16 to 65536); this re-applies the check in both methods (`AbstractReferenceMap` does not call `super.doReadObject`) and throws `InvalidObjectException`.

### DIFF
--- a/src/test/java/org/apache/commons/collections4/map/HashedMapTest.java
+++ b/src/test/java/org/apache/commons/collections4/map/HashedMapTest.java
@@ -92,18 +92,14 @@ public class HashedMapTest<K, V> extends AbstractIterableMapTest<K, V> {
     }
 
     /**
-     * A crafted stream can carry a load factor the constructor rejects. AbstractHashedMap and
-     * AbstractReferenceMap (its own doReadObject override) must reapply that contract on read.
+     * A crafted stream can carry a load factor the constructor rejects. AbstractHashedMap.doReadObject
+     * must reapply that contract on read.
      */
     @ParameterizedTest
     @ValueSource(floats = {0.0f, -1.0f, Float.NaN})
     void testDeserializeRejectsInvalidLoadFactor(final float badLoadFactor) {
-        final HashedMap<K, V> hashed = new HashedMap<>();
-        hashed.loadFactor = badLoadFactor;
-        assertThrows(InvalidObjectException.class, () -> serializeDeserialize(hashed));
-
-        final ReferenceMap<K, V> reference = new ReferenceMap<>();
-        reference.loadFactor = badLoadFactor;
-        assertThrows(InvalidObjectException.class, () -> serializeDeserialize(reference));
+        final HashedMap<K, V> map = new HashedMap<>();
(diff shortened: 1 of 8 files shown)


garydgregory (changes requested):
@rootvector2 
Thank you for the PR.
Please use a parameterized test instead of a loop.

rootvector2:
done, switched it to a `@ParameterizedTest` with `@ValueSource` over the three bad load factors.

src/test/java/org/apache/commons/collections4/map/HashedMapTest.java
```diff
@@ -85,4 +90,20 @@ void testInternalState() {
...
         assertEquals(96, map.threshold);
     }
+
+    /**
+     * A crafted stream can carry a load factor the constructor rejects. AbstractHashedMap and
+     * AbstractReferenceMap (its own doReadObject override) must reapply that contract on read.
+     */
+    @ParameterizedTest
+    @ValueSource(floats = {0.0f, -1.0f, Float.NaN})
+    void testDeserializeRejectsInvalidLoadFactor(final float badLoadFactor) {
```
garydgregory, line 100: Hello @rootvector2 
Thank you for your update.
I now realize this teste doesn't look right. The PR modifies two abstract classes. This test is in one of the concrete subclasses for only two subclasses. Instead, there should be one new test method in each concrete subclasses' Test class.

garydgregory (changes requested):
Hello @rootvector2 
Please see my additional comment. TY!

src/test/java/org/apache/commons/collections4/map/HashedMapTest.java
rootvector2: good catch. moved the `ReferenceMap` case into `ReferenceMapTest` so each concrete subclass has its own test method, and `HashedMapTest` now only covers `HashedMap`.

garydgregory: @rootvector2 

Thank you for the update. You forgot to add tests for the other concrete subclasses of the two abstract classes this PR modifies.

rootvector2: added them: `CaseInsensitiveMapTest`, `LinkedMapTest` and `LRUMapTest` on the `AbstractHashedMap` side, and `ReferenceIdentityMapTest` on the `AbstractReferenceMap` side. each one is a small method using the shared `serializeDeserialize` helper rather than copy-pasted boilerplate.

garydgregory:
Thank you for the update @rootvector2 , merged 🚀

PR #692, opened 2026-06-22
Fix AbstractMapBag.retainAll removing all copies when the other bag has more

vasiliy-mikhailov:
## Problem

`AbstractMapBag.retainAll(Bag)` should keep, for each element, the minimum of this bags cardinality and the other bags cardinality. The current logic:

```java
final int myCount = getCount(current);
final int otherCount = other.getCount(current);
if (1 <= otherCount && otherCount <= myCount) {
    excess.add(current, myCount - otherCount);
} else {
    excess.add(current, myCount);   // removes ALL copies
}
```

The `else` branch removes all copies of an element whenever the guarded case (`1 <= otherCount <= myCount`) does not apply. That branch also covers `otherCount > myCount`, so when the other bag contains **more** copies than this bag, every copy is removed instead of all being kept.

Example: `bag{A:2, B:3}.retainAll(other{A:5, B:10})` empties the bag, when it should leave it unchanged (the retained count is `min(2,5)=2` and `min(3,10)=3`).

## Fix

Only remove all copies when `otherCount == 0`. When `otherCount > myCount`, no copies are removed, leaving this bags cardinality intact.

## Test

Adds `testBagRetainAllOtherHasMoreCopies` to `AbstractBagTest`, so it runs against every `Bag` implementation. It fails on `master` (`expected: <2> but was: <0>`) and passes with this change; the existing `HashBagTest` suite (51 tests, including `testBagRetainAll`) stays green.

## AI assistance disclosure

### DIFF
--- a/src/main/java/org/apache/commons/collections4/bag/AbstractMapBag.java
+++ b/src/main/java/org/apache/commons/collections4/bag/AbstractMapBag.java
@@ -497,7 +497,7 @@ public abstract class AbstractMapBag<E> implements Bag<E> {
             final int otherCount = other.getCount(current);
             if (1 <= otherCount && otherCount <= myCount) {
                 excess.add(current, myCount - otherCount);
-            } else {
+            } else if (otherCount == 0) {
                 excess.add(current, myCount);
             }
         }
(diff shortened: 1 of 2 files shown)


garydgregory:
Thank you @vasiliy-mikhailov , merged 🚀

PR #694, opened 2026-06-23
[COLLECTIONS-776] Fix expiration bypass in PassiveExpiringMap when wrapped in SynchronizedMap

s8sankalp:
## Description
When `PassiveExpiringMap` is decorated with `Collections.synchronizedMap()`, calling view methods like `entrySet()`, `keySet()`, or `values()` returns cached view decorators managed by `SynchronizedMap`. This bypasses `PassiveExpiringMap`'s view methods and avoids triggering the passive eviction logic `removeAllExpired()` during iteration, resulting in expired entries remaining in the map indefinitely.

This PR introduces custom collection views (`EntrySet`, `KeySet`, `ValuesCollection`) and their respective iterators to:
1. Ensure `removeAllExpired()` is triggered on all read and write collection view operations.
2. Propagate iterator `remove()` actions to the map's internal `expirationMap`.
3. Preserve the O(1) complexity of `remove()` and `removeAll()` for the set views.

## Verification
- Added JUnit regression test `testCollectionsSynchronizedMapExpiration()` verifying correct passive eviction when iterating/accessing collection views of a synchronized `PassiveExpiringMap`.
- Added JUnit regression test `testCollectionViewRemoval()` verifying correct cleanup of `expirationMap` when elements are removed via view iterators.
- All unit tests pass successfully.

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -24,7 +24,7 @@
   <body>
   <release version="4.6.0" date="YYYY-MM-DD" description="This is a feature and maintenance release. Java 8 or later is required.">
     <!-- FIX -->
-    <action type="fix" dev="ggregory" due-to="Gary Gregory" issue="COLLECTIONS-776">Wrapping PassiveExpiringMap in a SynchronizedMap breaks expiration.</action>
+    <action type="fix" dev="ggregory" due-to="Gary Gregory" issue="COLLECTIONS-776">Wrapping PassiveExpiringMap in Collections.synchronizedMap breaks expiration.</action>
     <action type="fix" dev="ggregory" due-to="Naveed Khan, Gary Gregory">[javadoc] Document that BloomFilter merge operations are non-atomic and a filter that throws during a merge should be considered invalid.</action>
     <action type="fix" dev="ggregory" due-to="Gary Gregory">org.apache.commons.collections4.map.AbstractReferenceMap.ReferenceEntry.toReference(ReferenceStrength, T, int) now throws IllegalArgumentException instead of Error.</action>
     <action type="fix" dev="ggregory" due-to="Eric Hubert, Gary Gregory">Remove deprecation annotation of org.apache.commons.collections4.Factory; this will be deprecated in 5.0 in favor of java.util.function.Supplier.</action>
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -24,6 +24,7 @@
   <body>
   <release version="4.6.0" date="YYYY-MM-DD" description="This is a feature and maintenance release. Java 8 or later is required.">
     <!-- FIX -->
+    <action type="fix" dev="ggregory" due-to="Gary Gregory" issue="COLLECTIONS-776">Wrapping PassiveExpiringMap in a SynchonizedMap breaks expiration.</action>
     <action type="fix" dev="ggregory" due-to="Gary Gregory">org.apache.commons.collections4.map.AbstractReferenceMap.ReferenceEntry.toReference(ReferenceStrength, T, int) now throws IllegalArgumentException instead of Error.</action>
     <action type="fix" dev="ggregory" due-to="Eric Hubert, Gary Gregory">Remove deprecation annotation of org.apache.commons.collections4.Factory; this will be deprecated in 5.0 in favor of java.util.function.Supplier.</action>
     <action type="fix" dev="ggregory" due-to="Eric Hubert, Gary Gregory">Remove deprecation annotation of org.apache.commons.collections4.Predicate; this will be deprecated in 5.0 in favor of java.util.function.Predicate.</action>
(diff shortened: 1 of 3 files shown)


garydgregory:
-1 since applying only the `test` side of the patch doesn't cause any tests to fail. TDD is calling...

garydgregory (changes requested):
-1, see my comment.

garydgregory (changes requested):
Hello @s8sankalp 

Thank you for the PR.

There is untested code in this PR. To see what's missing:
- Run: `mvn clean verify site`
- Open the site in: `target/site/index.html`
- Navigate to the JaCoCo report

You'll see in red what's missing, for example, null inputs on the `removeAll()` and `retainAll()` methods in the new classes.

In unit tests, this is an anti-pattern unless you are specifically testing for `size()` implementation instead of state:
```
assertEquals(0, entrySet.size());
```
Update or add:
```
assertTrue(entrySet.isEmpty());
```
You don't need blanks lines all over the place since you nicely document the steps with `//` comments.

garydgregory (changes requested):
Hello @s8sankalp 

Why does `PassiveExpiringMap.EntrySet.iterator()` call `PassiveExpiringMap.removeAllExpired(long)` and `PassiveExpiringMap.KeySet.iterator()` does  not?

Isn't that call either superfluous in the former or missing in the later?

Commenting out the call of `PassiveExpiringMap.removeAllExpired(long)` in `PassiveExpiringMap.EntrySet.iterator()` doesn't cause any test to fail.

This tells me at lease one test is missing.

Please check.

s8sankalp:
> Hello @s8sankalp
> 
> Why does `PassiveExpiringMap.EntrySet.iterator()` call `PassiveExpiringMap.removeAllExpired(long)` and `PassiveExpiringMap.KeySet.iterator()` does not?
> 
> Isn't that call either superfluous in the former or missing in the later?
> 
> Commenting out the call of `PassiveExpiringMap.removeAllExpired(long)` in `PassiveExpiringMap.EntrySet.iterator()` doesn't cause any test to fail.
> 
> This tells me at lease one test is missing.
> 
> Please check.

hi @garydgregory 

The call to removeAllExpired() is not needed in KeySet.iterator() because both the key set and values iterators delegate directly to entrySet().iterator(), which already triggers the eviction check. Conversely, the eviction call in EntrySet.iterator() is necessary to support cached views. If a client caches a collection view (like the entry set) and waits for entries to expire before obtaining an iterator, the iterator must run the eviction check to avoid returning those expired elements. To verify this behavior and prevent future regressions, I have added a new unit test asserting that iterators created from cached views after expiration do not return any elements; this test now correctly fails if the eviction call in EntrySet.iterator() is removed.

s8sankalp:
Hi @garydgregory  ,

Just wanted to check in on this one. I replied above about why removeAllExpired() is needed in EntrySet.iterator() but not in KeySet.iterator(), and I added a new unit test to cover the cached view eviction case you pointed out. Let me know if that clears things up or if you'd like me to change anything else.
Thanks again for reviewing this, really appreciate the feedback.

garydgregory:
@s8sankalp 
That's a ton of new code to review. I need to consider if there's a simpler solution. More later.

s8sankalp:
@garydgregory 
Take your time

s8sankalp:
Hi @garydgregory ,

Thanks for the review!

I've addressed the feedback in the latest update:

* Updated the custom `EntrySet`, `KeySet`, and `ValuesCollection` implementations to align `removeAll(null)` and `retainAll(null)` with standard `Collection` semantics by rejecting null arguments via `Objects.requireNonNull(...)`.
* Updated the regression tests accordingly to expect `NullPointerException` for null inputs instead of returning `false`.
* Added additional regression coverage to verify that removals performed through collection views and iterators also clean up the associated expiration metadata, preventing stale entries from remaining in `expirationMap`.

The changes have been pushed to the PR branch for review.

s8sankalp:
Hi @garydgregory, just checking in. I pushed the requested changes over a day ago and it's ready for another look whenever you have a chance. Thanks!

garydgregory:
It's in my queue, not urgent.

garydgregory (commented):
@s8sankalp 
Please see the latest comments from copilot.

s8sankalp:
hi @garydgregory 

I have addressed all the remaining Copilot review findings:

1. **Corrected Misleading Comments in `PassiveExpiringMapTest`**: Updated the comments in `testCollectionsSynchronizedMapExpiration()` from `"iterator triggers expiration"` to `"view access triggers expiration"` (for `entrySet`, `keySet`, and `values`). This accurately reflects that the assertions verify expiration using `.isEmpty()` on the cached view rather than an iterator.
2. **Local Verification**: Checkstyle has 0 violations and all 19,894 tests in the suite pass successfully.

garydgregory:
TY @s8sankalp , merged 🚀

garydgregory:
The Jira ticket is [COLLECTIONS-776](https://issues.apache.org/jira/browse/COLLECTIONS-776).

s8sankalp:
> TY @s8sankalp , merged 🚀

Thanks for the review and for merging this, @garydgregory! I learned a lot working through the edge cases in PassiveExpiringMap/SynchronizedMap. Appreciate the guidance.

PR #695, opened 2026-06-25
Document non-atomic merge semantics for BloomFilter

rootvector2:
Per the discussion, dropped the per-index guard and documented the merge contract on the `BloomFilter` interface instead. A merge enables indexes one at a time and is not atomic, so an out-of-range index throws after earlier indexes are already set, and a filter that throws during a merge should be treated as invalid. Reverted the `SparseBloomFilter` change and its test, so only the Javadoc and a changes.xml note remain.

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -24,6 +24,7 @@
   <body>
   <release version="4.6.0" date="YYYY-MM-DD" description="This is a feature and maintenance release. Java 8 or later is required.">
     <!-- FIX -->
+    <action type="fix" dev="ggregory" due-to="Naveed Khan, Gary Gregory">[javadoc] Document that BloomFilter merge operations are non-atomic and a filter that throws during a merge should be considered invalid.</action>
     <action type="fix" dev="ggregory" due-to="Gary Gregory">org.apache.commons.collections4.map.AbstractReferenceMap.ReferenceEntry.toReference(ReferenceStrength, T, int) now throws IllegalArgumentException instead of Error.</action>
     <action type="fix" dev="ggregory" due-to="Eric Hubert, Gary Gregory">Remove deprecation annotation of org.apache.commons.collections4.Factory; this will be deprecated in 5.0 in favor of java.util.function.Supplier.</action>
     <action type="fix" dev="ggregory" due-to="Eric Hubert, Gary Gregory">Remove deprecation annotation of org.apache.commons.collections4.Predicate; this will be deprecated in 5.0 in favor of java.util.function.Predicate.</action>
--- a/src/main/java/org/apache/commons/collections4/bloomfilter/BloomFilter.java
+++ b/src/main/java/org/apache/commons/collections4/bloomfilter/BloomFilter.java
@@ -23,6 +23,13 @@ import java.util.Objects;
  * <p>
  * <em>See implementation notes for {@link BitMapExtractor} and {@link IndexExtractor}.</em>
  * </p>
+ * <p>
+ * The {@code merge} operations enable bit indexes one at a time and are not atomic. If an index is out of
+ * range an {@link IllegalArgumentException} is raised, but indexes processed before the bad one may already
+ * have been enabled. By design the filter is not rolled back: such an exception signals misuse with bad
+ * indexes and must not be ignored. A filter that throws during a merge should be considered invalid, since
+ * it may hold bits that do not correspond to any valid hashed value.
+ * </p>
  *
  * @param <T> The BloomFilter type.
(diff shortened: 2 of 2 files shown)


garydgregory:
Hi @aherbert A second set of eyes please 👀

aherbert:
Short answer: Bloom filter operations are not atomic. Making them so is too expensive. Whatever option we choose here will be a partial fix and the filter is likely compromised in any case. The library should be used when you have control over the entire stack of code that generates indices. Then you will not have a problem. This is a bug for an atypical use case.

---

I think that the library is mainly optimised for correct usage. If you deliberately use bad content then this lazy checking will result in the filter notifying the caller via an exception that it failed after it has modified its internal state. The filter should then be considered broken.

Note that if you do early checking on the indices and raise an exception on the first bad index, you may have already merged some other indices and so have a broken filter anyway. Unfortunately making the filter merge operations atomic is far too expensive for practical use. So any merge will likely break the filter if an index is invalid, no matter what code changes we make here.

The OP notes that index checking is performed in SimpleBloomFilter so that cannot have a corrupted state. But it would be corrupted if indices have been already merged. Note also in the same class the ``SimpleBloomFilter.merge(BitMapExtractor)`` can merge bits above the size of the filter defined by the shape. It will raise an exception but not fix its own state (which it could do with masking out the bad bits before raising the exception). So that also has an issue with broken cardinality after a bad merge.

We could modify the code to remove invalid indices before raising. Here the indices are in a TreeSet. So the indices could be removed with:
```java
indices.tailSet(shape.getNumberOfBits()).clear();
```

But this only partially corrects something that is broken. It is still broken, just less so. I think there are several solutions which require some discussion:

1. Clean up bad indices before throwing. The filter could have partially merged indices that do not correspond to any real hashed object.
2. Leave it as is and document that you should never ignore exceptions and then forever consider the filter compromised.
3. Set an internal flag before throwing the exception. This internal state flag would be similar to the one used in the ``ArrayCountingBloomFilter``. It knows when it has been broken based on this state flag. You can check this using ``isValid()`` at any point.
4. Make them atomic. You could go through all the indices and check them; them go through them all again and add them. This is double the computation cost for all correct usage of the classes to stop malicious usage which really should not be a problem anyway.

Unfortunately the ``isValid`` flag is part of the ``CountingBloomFilter`` interface. Perhaps it should have been put into the ``BloomFilter`` interface instead. Adding it now would have to be a default method that returns true. We could then implement it correctly in all the filters in this codebase.

Note: You cannot go through the indices again and remove them. You may remove indices that are from other object hashes added to the filter.

garydgregory:
Thank you for the review @aherbert
Pulling up `isValid()` to `BloomFiler` sounds reasonable to me. Would you be willing to create a PR? Otherwise, calrifying the Javadoc would be good.

rootvector2:
Fair point, the per-index guard only stops the single out-of-range value from being stored, it doesn't make `merge` atomic. Any valid indices ahead of the bad one are already in the `TreeSet` when it throws, so the filter can still hold bits that map to no real hash. It narrows the common single-bad-index case but isn't a full fix.

Pulling `isValid()` up to `BloomFilter` as a default returning `true`, then implementing it per filter, sounds like the right direction. Happy to defer to @aherbert if he picks up that PR, or I can add the Javadoc note here in the meantime.

aherbert:
I'd rather deal with this with no code changes. IMO the Bloom filter is used by someone with control of the entire stack that creates hashes. This should never happen in real use. The ``isValid`` flag in the ``CountingBloomFilter`` is to satisfy a real use case where you may add too many items to count. Moving it to ``BloomFilter`` to satisfy cases that will not happen in practical use is overkill.

I think this package could do with more javadoc to indicate that the filter merge operates using non-atomic additions of a set of indices. By design this is not meant to be recoverable if an exception occurs during the merge. Exceptions are a sign of misuse with bad indices. They should not be ignored by the user and a filter that emits an exception is considered invalid due to potentially containing bits not corresponding to a valid hash.

garydgregory:
Clarifying the Javadoc is reasonable IMO.

rootvector2:
Makes sense, agreed a guard can't make the merge atomic. Reverted the code change and documented the contract instead: added a note on the `BloomFilter` interface that merge enables indexes one at a time and isn't atomic, so an exception leaves the filter holding earlier indexes and it should be treated as invalid. Dropped the `SparseBloomFilter` change and its test, so the PR is now just the Javadoc plus a changes.xml line.

garydgregory:
LGTM.

garydgregory:
Merged 🚀 , thank you @rootvector2

PR #696, opened 2026-06-26
Do not mutate the input map in SwitchTransformer.switchTransformer

vasiliy-mikhailov:
`SwitchTransformer.switchTransformer(Map)` extracts the default transformer with `map.remove(null)`, which **mutates the caller's map** by removing its null key entry — a surprising destructive side effect from a factory method. Use `map.get(null)` instead so the input map is left unchanged.

Added a test asserting the input map is not mutated.

## AI assistance disclosure

### DIFF
--- a/src/main/java/org/apache/commons/collections4/functors/SwitchTransformer.java
+++ b/src/main/java/org/apache/commons/collections4/functors/SwitchTransformer.java
@@ -17,6 +17,7 @@
 package org.apache.commons.collections4.functors;
 
 import java.io.Serializable;
+import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Objects;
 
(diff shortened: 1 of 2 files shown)


garydgregory:
Hello @vasiliy-mikhailov 
The build fails. What happened when you ran a build locally?

vasiliy-mikhailov:
Thanks @garydgregory, and sorry for the red build. You're right that the full suite wasn't run.

Honest root cause: I prepare these changes with an AI pipeline, and it had a bug where a fix could be recorded as "verified" from a bare `exit 0` without the test suite actually having run. So this shipped looking green when it wasn't. The change swapped `map.remove(null)` for `map.get(null)` to stop the factory from mutating the caller's map, but `remove(null)` was load-bearing: it also dropped the `null` entry before the predicate/transformer arrays are built, so leaving it in puts a `null` predicate in the array and `testSwitchTransformer` hits an NPE in `SwitchTransformer.transform`.

I've fixed the pipeline so a fix is only treated as verified when there is a real green test run (actual `Tests run … Failures: 0` output), not a bare exit code. And I've pushed a corrected change: copy the map into a `LinkedHashMap` (preserving iteration order) and `remove(null)` from the copy, so the caller's map is untouched and the arrays are built correctly. Both `TransformerUtilsTest#testSwitchTransformer` and the new mutation test pass locally now. Thanks for the catch.

garydgregory:
Hello @vasiliy-mikhailov 
What about:
- `org.apache.commons.collections4.functors.SwitchClosure.switchClosure(Map<Predicate<E>, Closure<E>>)`
- `org.apache.commons.collections4.ClosureUtils.switchMapClosure(Map<? extends E, Closure<E>>)`
- `org.apache.commons.collections4.TransformerUtils.switchMapTransformer(Map<I, Transformer<I, O>>)`

?

PR #699, opened 2026-06-29
[COLLECTIONS-804] Fix non-deterministic testCollectionToArray2() and …

s8sankalp:
COLLECTIONS-804: Fix non-deterministic testCollectionToArray2() and flaky iterator tests under NonDex

### Problem
The test case `testCollectionToArray2()` in `AbstractCollectionTest` compares two generated arrays from `toArray()` calls. For collections that do not guarantee iteration order (such as standard hash-based maps, sets, or their decorators), the elements returned can vary in order between executions—especially when executed under the NonDex tool which shuffles iteration orders to expose non-determinism.

Similarly, `testForEachRemaining()` in `AbstractIteratorTest` creates two separate iterators from the same unordered backing data structure and expects them to traverse elements in the same order, causing flaky assertions.

### Solution
1. **`AbstractCollectionTest.java`**:
   - Modified `testCollectionToArray2()` to build an expected element list from the collection's iterator.
   - Asserts the outputs of `toArray()` directly against this expected list. If the collection behaves in an unordered manner (`getIterationBehaviour() & UNORDERED != 0`), the elements are compared ignoring order using `assertUnorderedArrayEquals`. Otherwise, strict order matches are validated.

2. **`AbstractMapTest.java`**:
   - Configured `getIterationBehaviour()` in `AbstractMapTest` to automatically return `UNORDERED` for unordered maps (by default, if the map is not a `SortedMap`, `OrderedMap`, `LinkedHashMap`, or does not have "Linked" or "Ordered" in its class name). This configures all collection views of hash-based maps to use unordered matching.

3. **`MapBackedSetTest.java`**:
   - Overrode `getIterationBehaviour()` to return `UNORDERED` as it is backed by an unordered `HashedMap`.

4. **`AbstractIteratorTest.java` & `AbstractMapIteratorTest.java`**:
   - Overrode and updated `testForEachRemaining()` to perform an unordered comparison using `HashBag` when the iterator traverses an unordered data structure, resolving flakiness under NonDex.

### Verification
All tests pass successfully under standard Maven runs and NonDex verification checks:
```bash
mvn edu.illinois:nondex-maven-plugin:2.2.5:nondex -Dtest=HashMapSanityTest,ConcurrentHashMapSanityTest,DualHashBidiMapTest,MapBackedSetTest -Drat.skip

### DIFF
--- a/src/test/java/org/apache/commons/collections4/collection/AbstractCollectionTest.java
+++ b/src/test/java/org/apache/commons/collections4/collection/AbstractCollectionTest.java
@@ -1225,13 +1225,20 @@ public abstract class AbstractCollectionTest<E> extends AbstractObjectTest {
                 "toArray(null) should raise NPE");
         verify();
 
+        final List<E> expectedList = new ArrayList<>();
+        for (final E element : getCollection()) {
+            expectedList.add(element);
+        }
+
         array = getCollection().toArray(ArrayUtils.EMPTY_OBJECT_ARRAY);
         a = getCollection().toArray();
 
         if ((getIterationBehaviour() & UNORDERED) != 0) {
-            assertUnorderedArrayEquals(array, a, "toArray(Object[]) and toArray()");
+            assertUnorderedArrayEquals(expectedList.toArray(), array, "toArray(Object[]) and iterator");
+            assertUnorderedArrayEquals(expectedList.toArray(), a, "toArray() and iterator");
         } else {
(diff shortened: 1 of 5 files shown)


garydgregory:
Closing: not worth maintainer time to change something that works. All builds are green and have been green for a long time. Look in Jira if you want to fix real bugs 😉

PR #700, opened 2026-07-01
Do not mutate the input map in switch closure/transformer factories

rootvector2:
`switchClosure(Map)`, `switchMapClosure(Map)` and `switchMapTransformer(Map)` pulled the default out with `map.remove(null)` on the caller's map, so the null-key entry vanished from the caller's map as a side effect; spotted while extending the sibling fix in #696 that only touched `SwitchTransformer.switchTransformer`, they now copy into a `LinkedHashMap` first.


garydgregory:
Hello @rootvector2 
Thank you for the PR.

Please update the tests here and the new one from https://github.com/apache/commons-collections/pull/696 to _also_ make sure the result from calling the various `switch*()` is valid.

Even though we might have other tests that do so, seeing a simple check in these new tests give me more confidence that there are no unintended side-effects.

TY!

rootvector2:
Done. Each test now executes the built `switch*()` result and checks it routes the match and the default correctly, and I added the same check to the `switchTransformer` test from #696.

PR #701, opened 2026-07-01
Fix stale size in AbstractMapMultiSet view iterator remove

rootvector2:
`AbstractMapMultiSet`'s `uniqueSet()` and `entrySet()` view iterators drop the whole entry from the backing map but never decrement the cached `size` field, so removing an element with `N` copies through either view leaves `size() == N` while `isEmpty()` is `true` and `toArray()` (sized by `size()`) returns `N` trailing nulls.

`UniqueSetIterator.remove()` calls `super.remove()` (which deletes the key) and then `parent.remove(lastElement, count)`, but that call short-circuits at `map.get(...) == null` and never adjusts `size`; `EntrySetIterator.remove()` calls `decorated.remove()` and never touches `size` at all. Both overrides now subtract the removed entry count from `size` and bump `modCount`, matching the size-correct `AbstractMapBag.BagIterator` and `remove(Object, int)`. Found by comparing the two map-backed count collections. `HashMultiSet` and `TreeMultiSet` inherit the fix.

Repro before the patch:

```java
MultiSet<String> ms = new HashMultiSet<>();
ms.add("a", 3);
Iterator<String> it = ms.uniqueSet().iterator();
it.next();
it.remove();
ms.size();    // 3  (expected 0)
ms.toArray(); // [null, null, null]
```

### DIFF
--- a/src/main/java/org/apache/commons/collections4/multiset/AbstractMapMultiSet.java
+++ b/src/main/java/org/apache/commons/collections4/multiset/AbstractMapMultiSet.java
@@ -92,7 +92,10 @@ public abstract class AbstractMapMultiSet<E> extends AbstractMultiSet<E> {
             if (!canRemove) {
                 throw new IllegalStateException("Iterator remove() can only be called once after next()");
             }
+            final int count = last.getCount();
             decorated.remove();
+            parent.size -= count;
+            parent.modCount++;
             last = null;
             canRemove = false;
         }
@@ -267,7 +270,8 @@ public abstract class AbstractMapMultiSet<E> extends AbstractMultiSet<E> {
             }
             final int count = parent.getCount(lastElement);
             super.remove();
-            parent.remove(lastElement, count);
+            parent.size -= count;
+            parent.modCount++;
             lastElement = null;
             canRemove = false;
         }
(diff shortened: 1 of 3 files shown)


garydgregory:
Hello @rootvector2 
- Please read the `MultiSet` contract and it seems like this PR would break it. WDYT?
- This PR changes a `main` abstract class but only tests a single implementation. If this PR is an actual fix (see above), we need to test all concrete subclasses.
TY!

rootvector2:
The removal behavior is unchanged and still follows the contract. `uniqueSet()` and `entrySet()` removal drops the whole entry (all occurrences of the element), exactly as before. The patch only fixes the cached `size` field, which those two view iterators never decremented. That left `size()` reporting `N` while `isEmpty()` returned `true` and `toArray()` (sized by `size()`) came back with `N` trailing nulls, so `size()` was inconsistent with the other views. Found it comparing this class against `AbstractMapBag`, whose bag iterator already adjusts `size`.

On testing: moved the regression test up into `AbstractMultiSetTest` so it runs for every concrete implementation (`HashMultiSet`, `PredicatedMultiSet`, `SynchronizedMultiSet`; skipped for `UnmodifiableMultiSet`, which does not support removal) instead of just `HashMultiSet`.

garydgregory:
Thank you @rootvector2 , merged 🚀 Note that I added 2 assertions to check that `remove()` can only be called once per `next()`.

PR #704, opened 2026-07-06
Fix MapUtils.getIntValue(Map, K, Function) returning a byte

rootvector2:
`MapUtils.getIntValue(Map, K, Function)` narrows the resolved `Integer` through `.byteValue()`, so any value outside signed-byte range comes back truncated: a map entry of `1000` returns `-24`, and a `defaultFunction` yielding `Integer.MAX_VALUE` returns `-1`. This is the same defect fixed under COLLECTIONS-874 for the sibling `getLongValue(Map, K, Function)`, which moved to `.longValue()`; the `int` overload kept `.byteValue()`. Found while checking that each `getXxxValue(Map, K, Function)` matches its declared return type.

Use `.intValue()` like every other numeric overload. `testGetIntValue` only drove the `Function` variant with `key -> 0`, which is unaffected by the narrowing, so the gap stayed invisible; the added assertions pin a map value and a default-function value that fall outside byte range.

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -32,6 +32,7 @@
     <action type="fix" dev="ggregory" due-to="Eric Hubert, Gary Gregory">Remove deprecation annotation of org.apache.commons.collections4.Transformer; this will be deprecated in 5.0 in favor of java.util.function.Function.</action>
     <action type="fix" dev="ggregory" due-to="Eric Hubert, Gary Gregory">Remove deprecation annotation of org.apache.commons.collections4.Closure; this will be deprecated in 5.0 in favor of java.util.function.Consumer.</action>
     <action type="fix" dev="ggregory" due-to="Sebastian Götz, Gary Gregory" issue="COLLECTIONS-874">MapUtils.getLongValue(Map, K, Function) returns a byte instead of a long.</action>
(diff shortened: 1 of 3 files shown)


garydgregory:
@rootvector2 
Thank you for the PR.
Do you see any other issues where there is a conversion mismatch like the one fixed here. It seems like something that happened with some copy-paste. I wonder if we have other missing unit tests for this class. Would you please check?
TY!

rootvector2:
checked all 18 `getXxxValue` methods against their declared return type. this `getIntValue(Map, K, Function)` narrowing to `.byteValue()` is the only mismatch, everything else (`byte`/`short`/`long`/`float`/`double` function overloads, plus every `defaultValue` overload) narrows to its own type. so no other copy-paste like COLLECTIONS-874 is lurking.

on tests: the `Function` overloads were the blind spot. `getIntValue` was only driven with `key -> 0`, which survives the truncation, so the bug stayed invisible. `getLongValue` already covers `Long.MIN/MAX_VALUE` through the function variant. `getShortValue` had the same weak spot, its function overload was only exercised with `10` (fits in a byte), while the `MIN/MAX` assertions go through the `short` overload instead. pushed a one-liner there driving `key -> (short) 1000` so a future `.byteValue()` slip would be caught. `byte`/`double`/`float` can't narrow past their own source type, so nothing to add there.

garydgregory:
Thank you @rootvector2 , merged 🚀 I appreciate the extra look.

rootvector2:
Thanks👍

PR #705, opened 2026-07-07
Clamp overflow in AbstractMapBag and AbstractMapMultiSet add

rootvector2:
`add(Object, int)` in `AbstractMapBag` and `AbstractMapMultiSet` accumulates the running `size` and the per-element count into plain `int` fields with no upper bound, so a second `add` whose total crosses `Integer.MAX_VALUE` overflows both: `add(x, Integer.MAX_VALUE)` then `add(x, 1)` leaves `getCount(x) == Integer.MIN_VALUE` and `size()` negative. Clamp both to `Integer.MAX_VALUE` inside `add()`, matching the `Collection.size()` contract and the existing `BloomFilter`/`CompositeMap.size()` idiom. Found while auditing the bag/multiset count arithmetic; the regression tests in `HashBagTest` and `HashMultiSetTest` fail before the change (`-4` / `Integer.MIN_VALUE`) and pass after.

- [x] Each commit in the pull request should have a meaningful subject line and body.

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -24,6 +24,7 @@
   <body>
   <release version="4.6.0" date="YYYY-MM-DD" description="This is a feature and maintenance release. Java 8 or later is required.">
     <!-- FIX -->
+    <action type="fix" dev="ggregory" due-to="Naveed Khan, Gary Gregory">AbstractMapBag and AbstractMapMultiSet.add(Object, int) overflow the size and element count past Integer.MAX_VALUE.</action>
     <action type="fix" dev="ggregory" due-to="Gary Gregory" issue="COLLECTIONS-776">Wrapping PassiveExpiringMap in Collections.synchronizedMap breaks expiration.</action>
     <action type="fix" dev="ggregory" due-to="Naveed Khan, Gary Gregory">[javadoc] Document that BloomFilter merge operations are non-atomic and a filter that throws during a merge should be considered invalid.</action>
     <action type="fix" dev="ggregory" due-to="Gary Gregory">org.apache.commons.collections4.map.AbstractReferenceMap.ReferenceEntry.toReference(ReferenceStrength, T, int) now throws IllegalArgumentException instead of Error.</action>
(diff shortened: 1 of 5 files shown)


garydgregory:
Hello @rootvector2 
Isn't `remove()` broken in the same kind of way? `add()` is guarded for negative `nCopies` but `remove()` isn't.

garydgregory:
@rootvector2 
Please see the copilot comments.

rootvector2:
`remove()` already rejects negatives on both sides: `AbstractMapBag.remove(Object, int)` returns false for `nCopies <= 0` and `AbstractMapMultiSet.remove(Object, int)` throws `IllegalArgumentException`, so counts cannot grow through it.

The copilot comments are right about the stored clamp though: once `size` saturated, removals dropped `size()` below `Integer.MAX_VALUE` while the true cardinality was still above it. Reworked: `size` is now a `long` kept exact and clamped on return in `size()`. Per-element counts still saturate at `Integer.MAX_VALUE` since they are stored as `int`, and `size` only accrues the applied delta so it always equals the sum of counts. Both tests now cover removal while the true size is still above `Integer.MAX_VALUE`. Full suite, checkstyle, pmd and japicmp are green.

garydgregory:
Hello @rootvector2 
Re-running CI. Would you please check the rest of Collections to see for other int/long mismatches?

garydgregory:
@rootvector2 
Please review `org.apache.commons.collections4.multiset.AbstractMapMultiSet.add(E, int)` for the same problem.

rootvector2:
Went over the merged `AbstractMapMultiSet.add(E, int)` once more: negative `occurrences` throws, the per-element count saturates through the applied delta so `mut.value` can't wrap, and the exact total sits in the `long` `size` field and clamps in `size()`. The multiset regression test covers add saturation plus removal while the true size is still above `Integer.MAX_VALUE`, so I don't see anything left in that path.

On the wider int/long pass: the same summation overflow exists in `CompositeCollection.size()`, `CompositeSet.size()` and `CompositeMap.size()`, which sum member sizes into an int. `CompositeMap`'s javadoc even promises `Integer.MAX_VALUE` but the sum wraps negative, and it's cheap to hit now by compositing two bags at max size. `AbstractMultiValuedMap.size()` and `AbstractMultiSet.size()` (for subclasses that don't override it) accumulate the same way. I'll put together a follow-up PR for those.

Thanks for merging this one.

garydgregory:
@rootvector2 Great, thank you.

PR #706, opened 2026-07-09
[COLLECTIONS-732]: Document UnsupportedOperationException on MultiValue…

s8sankalp:
## Problem
When calling `setValue(V value)` on the `MapIterator` returned by `MultiValuedMap.mapIterator()`, it throws an `UnsupportedOperationException`. However, this behavior was previously undocumented in the interface and its implementing classes/decorators, causing confusion for users.

For details, see [JIRA COLLECTIONS-732](https://issues.apache.org/jira/browse/COLLECTIONS-732).

## Proposed Changes
Documented that the returned map iterator's `setValue` method is unsupported and throws `UnsupportedOperationException` in:
- `MultiValuedMap.java` (Interface)
- `AbstractMultiValuedMap.java` (Base Implementation, including the inner classes `MultiValuedMapIterator` and `MultiValuedMapEntry`)
- `AbstractMultiValuedMapDecorator.java` (Decorator base)
- `UnmodifiableMultiValuedMap.java` (Unmodifiable wrapper)

## Verification
- Verified that all unit tests compile and pass successfully (`mvn test`).
- Existing tests (such as `testMapIteratorUnsupportedSet` in `AbstractMultiValuedMapTest.java`) already assert that `setValue()` throws `UnsupportedOperationException`, confirming the behavior matches the new documentation.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/multimap/AbstractMultiValuedMap.java
+++ b/src/main/java/org/apache/commons/collections4/multimap/AbstractMultiValuedMap.java
@@ -343,6 +343,11 @@ public abstract class AbstractMultiValuedMap<K, V> implements MultiValuedMap<K, 
             it.remove();
         }
 
+        /**
+         * {@inheritDoc}
+         *
+         * @throws UnsupportedOperationException always, as this method is not supported by the iterator.
+         */
         @Override
         public V setValue(final V value) {
             if (current == null) {
@@ -742,6 +747,13 @@ public abstract class AbstractMultiValuedMap<K, V> implements MultiValuedMap<K, 
         return getMap().keySet();
     }
 
+    /**
+     * {@inheritDoc}
+     * <p>
+     * The returned map iterator's {@link MapIterator#setValue(Object)} method is not supported
+     * and will throw an {@link UnsupportedOperationException}.
+     * </p>
+     */
     @Override
     public MapIterator<K, V> mapIterator() {
         if (isEmpty()) {
--- a/src/main/java/org/apache/commons/collections4/multimap/AbstractMultiValuedMapDecorator.java
+++ b/src/main/java/org/apache/commons/collections4/multimap/AbstractMultiValuedMapDecorator.java
@@ -130,13 +130,6 @@ public abstract class AbstractMultiValuedMapDecorator<K, V>
         return decorated().keySet();
     }
 
-    /**
-     * {@inheritDoc}
-     * <p>
-     * The returned map iterator's {@link MapIterator#setValue(Object)} method is not supported
-     * and will throw an {@link UnsupportedOperationException}.
-     * </p>
-     */
     @Override
     public MapIterator<K, V> mapIterator() {
         return decorated().mapIterator();
--- a/src/main/java/org/apache/commons/collections4/multimap/AbstractMultiValuedMapDecorator.java
+++ b/src/main/java/org/apache/commons/collections4/multimap/AbstractMultiValuedMapDecorator.java
@@ -130,6 +130,13 @@ public abstract class AbstractMultiValuedMapDecorator<K, V>
         return decorated().keySet();
     }
 
+    /**
+     * {@inheritDoc}
+     * <p>
+     * The returned map iterator's {@link MapIterator#setValue(Object)} method is not supported
+     * and will throw an {@link UnsupportedOperationException}.
+     * </p>
+     */
     @Override
     public MapIterator<K, V> mapIterator() {
         return decorated().mapIterator();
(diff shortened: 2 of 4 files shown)


src/main/java/org/apache/commons/collections4/multimap/AbstractMultiValuedMap.java
```diff
@@ -285,6 +285,13 @@ private final class MultiValuedMapEntry extends AbstractMapEntry<K, V> {
             super(key, value);
         }
 
+        /**
+         * {@inheritDoc}
+         * <p>
+         * This implementation always throws an {@link UnsupportedOperationException}
+         * as modification of the entry is not supported.
+         * </p>
+         */
```
garydgregory, line 293: That's not how to document exceptions in Javadoc, you need to read https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#@exception and use `@throws`.

src/main/java/org/apache/commons/collections4/multimap/AbstractMultiValuedMap.java
```diff
@@ -337,6 +344,10 @@ public void remove() {
             it.remove();
         }
 
+        /**
+         * @throws UnsupportedOperationException always, as this method is not supported
```
garydgregory: Missing inherit tag.

src/main/java/org/apache/commons/collections4/multimap/AbstractMultiValuedMapDecorator.java
```diff
@@ -130,6 +130,13 @@ public Set<K> keySet() {
         return decorated().keySet();
     }
 
+    /**
+     * {@inheritDoc}
+     * <p>
```
garydgregory: See above.

garydgregory (changes requested):
@s8sankalp 
I added some comments that apply to all changes. Keep it simple for the simple cases, for example 09048c1b6b76063630196e58afecdd10f15a4ab5

s8sankalp:
hi @garydgregory ,

I have addressed the feedback and resolved the conflicts with the default branch:

1)Updated the Javadoc in MultiValuedMapIterator.setValue to include the {@inheritDoc} tag and structured the @throws UnsupportedOperationException tag correctly.

2)Rebased the branch on the latest master to resolve all merge conflicts cleanly.

Everything builds successfully and the Javadoc generates without errors. Ready for another review!

src/main/java/org/apache/commons/collections4/multimap/AbstractMultiValuedMapDecorator.java
```diff
@@ -130,6 +130,13 @@ public Set<K> keySet() {
...
+     * The returned map iterator's {@link MapIterator#setValue(Object)} method is not supported
+     * and will throw an {@link UnsupportedOperationException}.
+     * </p>
+     */
```
garydgregory: @s8sankalp 
How can this be true for all implementations of the decorated object?

garydgregory:
Converted to draft, not fully baked.

src/main/java/org/apache/commons/collections4/multimap/AbstractMultiValuedMapDecorator.java
s8sankalp: @garydgregory 
Good catch. AbstractMultiValuedMapDecorator simply delegates mapIterator() to the wrapped MultiValuedMap, so it cannot guarantee the behavior of MapIterator#setValue() for all implementations. I've removed the incorrect Javadoc and pushed the update. Thanks!

src/main/java/org/apache/commons/collections4/multimap/UnmodifiableMultiValuedMap.java
```diff
@@ -114,6 +114,13 @@ public Set<K> keySet() {
         return UnmodifiableSet.unmodifiableSet(decorated().keySet());
     }
 
+    /**
+     * {@inheritDoc}
+     * <p>
+     * The returned map iterator's {@link MapIterator#setValue(Object)} method is not supported
```
garydgregory, line 120: This is only partially true, why mention one mutator method and not another? Wouldn't it be simpler to say that the returned object is unmodifiable? The same type of issue must be verified for each of the changes in this PR and this is not something I want to spend time doing ATM. I think we should close this PR and move on to actual bug fixes.

s8sankalp: Thanks for the review and the explanation. That makes sense. I understand the concern about the review effort, and I'm happy to close this PR and focus on bug fixes instead.

PR #712, opened 2026-07-13
[COLLECTIONS-714] PatriciaTrie ignores trailing null characters in keys

s8sankalp:
## Description

This PR fixes **COLLECTIONS-714**, where `PatriciaTrie` cannot distinguish between the keys `"x"` and `"x\u0000"`.

The root cause is in `StringKeyAnalyzer`. It treats the end of a string as `'\u0000'`, so when comparing keys of different lengths, an actual `'\u0000'` character becomes indistinguishable from the end of the string. As a result, `bitIndex()` returns `EQUAL_BIT_KEY` for keys that are not equal, causing `PatriciaTrie.put()` to replace an existing entry instead of inserting a new one.

This patch changes the internal bit representation used by `StringKeyAnalyzer`. Each character position is represented using a presence bit followed by the 16 character bits. The presence bit indicates whether a character exists at that position, allowing the analyzer to distinguish between the end of a string and an actual `'\u0000'` character.

With this representation:

* `"x"` and `"x\u0000"` are treated as distinct keys.
* Prefix keys branch at the presence bit of the first missing character, preventing bit-index collisions while preserving PatriciaTrie invariants.
* The existing ordering semantics of `PatriciaTrie` are preserved.
* All valid Java UTF-16 code units, including control characters, are treated uniformly without relying on a sentinel character.

The following regression tests were added:

* `testNullTerminatedKey1`
* `testNullTerminatedKey2`
* `testControlCharactersRegression`

The regression tests verify insertion, lookup, removal, iteration order, and trie view operations (`prefixMap`, `headMap`, `tailMap`, and `subMap`) for keys containing `'\u0000'` and other UTF-16 control characters. The behavior was also compared against `HashMap` and `TreeMap` where appropriate.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/MultiValuedMap.java
+++ b/src/main/java/org/apache/commons/collections4/MultiValuedMap.java
@@ -199,10 +199,6 @@ public interface MultiValuedMap<K, V> {
      * A map iterator is an efficient way of iterating over maps. There is no
      * need to access the entries collection or use {@code Map.Entry} objects.
      * </p>
-     * <p>
-     * The returned map iterator's {@link MapIterator#setValue(Object)} method is not supported
-     * and will throw an {@link UnsupportedOperationException}.
-     * </p>
      *
      * @return a map iterator
      */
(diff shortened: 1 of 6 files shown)


garydgregory:
Hello @s8sankalp 
What about other control characters?

s8sankalp:
hi @garydgregory 

Thanks for pointing that out.

I added regression tests for additional control characters (`\u0001`, `\u0002`, `\u001F`, and `\u007F`) as well. The new encoding no longer uses `\u0000` as a sentinel, so all UTF-16 control characters are treated as regular character values. The new tests pass, and the full Maven test suite completes successfully without regressions.

garydgregory:
Jira ticket [COLLECTIONS-714](https://issues.apache.org/jira/browse/COLLECTIONS-714).

PR #713, opened 2026-07-14
Reject out-of-range index before mutating in indexed add

rootvector2:
`SetUniqueList.add(int, E)`/`addAll(int, Collection)` and `ListOrderedSet.add(int, E)`/`addAll(int, Collection)` mutate the uniqueness set before the index-validating backing collection, so an out-of-range index throws from the list but the element is already in the set, leaving `contains(x)` true while the element is absent, `size()` out of step with `asSet()`/`asList()`, and the element silently dropped by the next `add`; `ListOrderedMap.put(int, K, V)` already range-checks the index up front, so these four now do the same. Found while auditing the ordered/unique decorators for state left behind by a rejected operation.

- [x] Each commit in the pull request should have a meaningful subject line and body.

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -24,6 +24,7 @@
   <body>
   <release version="4.6.0" date="YYYY-MM-DD" description="This is a feature and maintenance release. Java 8 or later is required.">
     <!-- FIX -->
+    <action type="fix" dev="ggregory" due-to="Naveed Khan">SetUniqueList and ListOrderedSet leave the backing collection inconsistent when add(int, Object) or addAll(int, Collection) is called with an out-of-range index.</action>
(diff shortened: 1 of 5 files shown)


garydgregory:
@rootvector2 
Thank you for the PR.
I triggered builds. Where else are we missing index bounds check?

rootvector2:
Thanks for merging. I audited the same pattern (a decorator mutating a parallel structure before the delegate validates the index) and these two were the only ones exposed: `SetUniqueList.set(int, E)` already goes through `super.set` first so the index is checked before the uniqueness set is touched, and `ListOrderedMap.put(int, K, V)` range-checks up front. The plain `List` decorators (`TransformedList`, `PredicatedList`, `GrowthList`, `FixedSizeList`) just delegate straight to the backing list, so it validates for them. Will keep an eye out if I spot more.

garydgregory:
Great, thank you @rootvector2

PR #714, opened 2026-07-16
Fix Flat3Map entrySet remove ignoring the entry value

rootvector2:
noticed `Flat3Map.EntrySet.remove` diverges from `AbstractHashedMap`: it removes on a key match alone, so `entrySet().remove(entry)` drops a mapping even when the entry value differs; guard with `contains(obj)` so only a full key-and-value match removes.

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -24,8 +24,6 @@
   <body>
   <release version="4.6.0" date="YYYY-MM-DD" description="This is a feature and maintenance release. Java 8 or later is required.">
     <!-- FIX -->
-    <action type="fix" dev="ggregory" due-to="Naveed Khan">Flat3Map entrySet().remove(Object) removes a mapping when only the key matches, ignoring the entry value.</action>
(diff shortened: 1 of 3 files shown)


garydgregory:
Good one @rootvector2 , will merge 🚀 . We have many of these types of types which may suffer from the same issue. Would you please check other Maps? TY! In the future, don't edit `changes.xml`.

rootvector2:
went through the other `EntrySet.remove` implementations: `AbstractHashedMap` and `AbstractPatriciaTrie` guard with `contains(obj)`, `TreeBidiMap` (both views), `AbstractDualBidiMap`, `PassiveExpiringMap` and `ListOrderedMap` all check the value, `StaticBucketMap` compares entries via `Node.equals` (key and value), and `ConcurrentReferenceHashMap` delegates to `remove(key, value)`. `Flat3Map` was the only one removing on key alone, so nothing else needs fixing. noted on `changes.xml`, will leave it out going forward. thanks for merging.

PR #715, opened 2026-07-18
Fix CollectionUtils collate keeping duplicate nulls when includeDuplicates is false

rootvector2:
`collate` used `lastItem == null` as its first-iteration sentinel, so a real null element (legal when a null-tolerant `Comparator` like `Comparator.nullsFirst` is supplied) was treated as unseen and re-added, leaving duplicate nulls in the result when `includeDuplicates` is false; found while tracing the merge dedup path with null-permitting comparators.

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -24,6 +24,7 @@
   <body>
   <release version="4.6.0" date="YYYY-MM-DD" description="This is a feature and maintenance release. Java 8 or later is required.">
     <!-- FIX -->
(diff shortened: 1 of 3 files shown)


garydgregory:
Good one @rootvector2 , merged 🚀

garydgregory:
In the future, don't edit `changes.xml`.

rootvector2:
thanks for merging, and noted on `changes.xml`, I'll leave it out of future PRs.

PR #721, opened 2026-08-03
[COLLECTIONS-897] Add LexicographicPermutationIterator

hextriclosan:
Add an Iterator<List<E>> that generates the permutations of a collection in lexicographical order, complementing PermutationIterator, which uses the Steinhaus-Johnson-Trotter ordering.

Elements are ordered by their natural ordering, or by a Comparator supplied to the two-argument constructor, which also allows permuting elements that do not implement Comparable. Each call to next() advances by the standard next-permutation step: locate the pivot, swap it with its successor, then reverse the descending tail. Equal elements are not distinguished, so an input with duplicates yields fewer than n! permutations. An empty collection yields exactly one empty list, as 0! = 1. remove() is unsupported.

Comparator dispatch follows the java.util.TreeMap pattern of testing the comparator field for null on each comparison; benchmarking showed no measurable difference against normalizing null to Comparator.naturalOrder() in the constructor.

Tests extend AbstractIteratorTest to cover the Iterator contract, and add cases for lexicographical exhaustivity, duplicate handling, custom and reverse comparators, non-Comparable elements, stream traversal, exhaustion, and equals/hashCode.

- [x] I used AI to create any part of, or all of, this pull request. Which AI tool was used to create this pull request, and to what extent did it contribute? Claude for polishing documentation.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/iterators/LexicographicPermutationIterator.java
+++ b/src/main/java/org/apache/commons/collections4/iterators/LexicographicPermutationIterator.java
@@ -159,38 +159,6 @@ public class LexicographicPermutationIterator<E> implements Iterator<List<E>> {
     }
 
     /**
-     * Compares this iterator to another for equality. Two iterators are equal when
-     * they use equal comparators and are positioned at an equal next permutation.
-     *
-     * @param o  The object to compare to this instance
-     * @return true if the given object is an equal iterator, otherwise false
-     */
-    @Override
-    public boolean equals(final Object o) {
-        if (this == o) {
-            return true;
-        }
-
-        if (o == null || getClass() != o.getClass()) {
-            return false;
-        }
-
-        final LexicographicPermutationIterator<?> that = (LexicographicPermutationIterator<?>) o;
-        return Objects.equals(comparator, that.comparator) && Objects.equals(nextPermutation, that.nextPermutation);
-    }
-
-    /**
-     * Returns a hash code consistent with {@link #equals(Object)}. Note that the
-     * hash code changes as the iterator advances.
-     *
-     * @return A hash code for this instance
-     */
-    @Override
-    public int hashCode() {
-        return Objects.hash(comparator, nextPermutation);
-    }
-
-    /**
      * Compares two elements using the comparator, or their natural ordering if no
      * comparator was supplied.
      *
--- a/src/main/java/org/apache/commons/collections4/iterators/LexicographicPermutationIterator.java
+++ b/src/main/java/org/apache/commons/collections4/iterators/LexicographicPermutationIterator.java
@@ -65,7 +65,7 @@ import java.util.Objects;
  *
  * @param <E>  the type of the objects being permuted
  * @see PermutationIterator
- * @since 4.6.0
+ * @since 4.7.0
  */
 public class LexicographicPermutationIterator<E> implements Iterator<List<E>> {
 
--- a/src/main/java/org/apache/commons/collections4/iterators/LexicographicPermutationIterator.java
+++ b/src/main/java/org/apache/commons/collections4/iterators/LexicographicPermutationIterator.java
@@ -29,15 +29,38 @@ import java.util.Objects;
  * This iterator creates permutations of an input collection, using the
  * lexicographical order.
  * <p>
+ * Iteration starts at the arrangement in which the elements are given, and each
+ * call to {@code next()} advances to the smallest arrangement greater than the
+ * current one. Iteration therefore ends at the largest arrangement, and the ones
+ * preceding the given arrangement are never returned: only a collection already
+ * sorted according to the ordering in use yields the complete set of
+ * permutations. Callers wanting the complete set must sort the collection
+ * beforehand, just as callers of
+ * {@link java.util.Collections#binarySearch(java.util.List, Object) binarySearch}
+ * must. Callers wanting to enumerate one part of the set, to resume from a
+ * previously reached arrangement or to split the work, may start anywhere.
+ * </p>
+ * <p>
+ * The starting arrangement is the iteration order of the given collection, so
+ * collections whose iteration order is unspecified, such as {@link java.util.HashSet},
+ * make poor input: which permutations are returned is then unspecified too.
+ * </p>
+ * <p>
  * The iterator might return fewer than n! permutations of the input collection,
- * because duplicated permutations are skipped: equal elements are not
- * distinguished from one another.
+ * either because the collection was not sorted according to the ordering in use,
+ * as described above, or because duplicated permutations are skipped: equal
+ * elements are not distinguished from one another.
  * The {@code remove()} operation is not supported, and will throw an
  * {@code UnsupportedOperationException}.
  * </p>
  * <p>
  * NOTE: in case an empty collection is provided, the iterator will
  * return exactly one empty list as result, as 0! = 1.
+ * </p>
+ * <p>
+ * NOTE: {@link PermutationIterator} differs on both counts. It returns exactly n!
+ * permutations whatever the order of the input collection. The two iterators are
+ * therefore not interchangeable.
  * </p>
  *
  * @param <E>  the type of the objects being permuted
@@ -60,6 +83,10 @@ public class LexicographicPermutationIterator<E> implements Iterator<List<E>> {
 
     /**
      * Standard constructor for this class, using the natural ordering of the elements.
+     * <p>
+     * Iteration starts at the arrangement in which the collection iterates its
+     * elements; sort the collection first to obtain the complete set of permutations.
+     * </p>
      *
      * @param collection  The collection to generate permutations for
      * @throws NullPointerException if collection is null
@@ -70,6 +97,11 @@ public class LexicographicPermutationIterator<E> implements Iterator<List<E>> {
 
     /**
      * Constructs an instance using the given comparator to order the elements.
+     * <p>
+     * Iteration starts at the arrangement in which the collection iterates its
+     * elements; sort the collection with the same comparator first to obtain the
+     * complete set of permutations.
+     * </p>
      *
      * @param collection  The collection to generate permutations for
      * @param comparator  The comparator used to define the order of generation,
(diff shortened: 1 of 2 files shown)


garydgregory:
Hello @hextriclosan 

Thank you for the PR.

Please review each Copiot comment and address them in comments here, in the code, or both. If you update the code, do make sure new unit tests cover all execution paths.

TY!

src/main/java/org/apache/commons/collections4/iterators/LexicographicPermutationIterator.java
```diff
@@ -0,0 +1,198 @@
...
+     * @param collection  The collection to generate permutations for
+     * @param comparator  The comparator used to define the order of generation,
+     *                    or null to use the natural ordering of the elements
+     * @throws NullPointerException if collection is null
+     */
+    public LexicographicPermutationIterator(final Collection<? extends E> collection, final Comparator<? super E> comparator) {
+        Objects.requireNonNull(collection, "collection");
+        nextPermutation = new ArrayList<>(collection);
+        this.comparator = comparator;
+    }
```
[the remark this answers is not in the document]
hextriclosan: The described behavior is real and the javadoc was actively misleading about it. I've pushed doc and test changes rather than the sort, and here's the reasoning.

The truncation is intentional. This class is the iterator form of the classic next-permutation step: it starts wherever the input puts it and advances to the smallest arrangement greater than the current one, exactly as `std::next_permutation` does in C++. Sorted input is a precondition for enumerating the full set, in the same way sorted input is a precondition for `Collections.binarySearch`. That precondition simply wasn't documented.

I'd rather not sort in the constructor, because sorting is not a neutral addition. It removes a capability that cannot be recovered:

- Resuming. A caller that persisted the last arrangement it processed can construct an iterator from it and carry on. With a constructor sort there is no way to express "start here".
- Splitting the work. The permutation space can be divided across threads or machines by handing each worker a different starting arrangement. Same problem.

A caller who wants the complete set can always sort before constructing, and that is one line at the call site. A caller who wants to start partway through has no recourse if the constructor sorts. The asymmetry is what decides it for me: preserving the given order is strictly the more expressive of the two designs.

There's also a ready alternative for callers who want all n! without thinking about order, namely `PermutationIterator`, which reaches every arrangement from any starting point because Steinhaus-Johnson-Trotter enumerates the whole group. The two classes are genuinely different tools, and I've added a note making the differences explicit so the @see link stops implying they're interchangeable.

src/test/java/org/apache/commons/collections4/iterators/LexicographicPermutationIteratorTest.java
```diff
@@ -0,0 +1,343 @@
...
+    @Override
+    public boolean supportsRemove() {
+        return false;
+    }
+
+    @Test
+    void testCustomComparator() {
+        final Iterator<List<Character>> permutationIterator = new LexicographicPermutationIterator<>(Arrays.asList('C', 'B', 'A'),
+                Comparator.reverseOrder());
+
```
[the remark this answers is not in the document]
hextriclosan: Good news on this one: the test already covers what you're after, and I can show it. I patched `compareElements` to ignore the comparator and ran the suite. `testCustomComparator` fails, and so does `testCustomComparatorWithNonComparableObjects`.                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                   
The reason `['C','B','A']` works as input is that it's the maximum under natural ordering. An implementation that ignored the comparator would find no pivot, terminate after a single permutation, and fail on the second `assertTrue`. The test therefore separates "comparator honoured, 6 permutations in reverse-lexicographic order" from "comparator ignored, 1 permutation".                                                                                                                                                                        

You're right that neither test exercised input unsorted under its own comparator. I've added `testUnsortedCollectionStartsAtGivenArrangementWithComparator` for exactly that, passing ['B','C','A'] with `reverseOrder()` and asserting the four permutations that follow it. Starting at the given arrangement is the intended contract here rather than a bug, for the reasons in  the other thread, and that test now pins it.

hextriclosan:
@garydgregory,

I've pushed doc and test changes. The implementation is unchanged, my reasoning is in the inline replies.

garydgregory:
C++? That's completely irrelevant. The code should only care about (1) the specifics of Commons Collections, and (2) how we extend Java Collections. Whatever happens in a C++ library doesn't come into play.

hextriclosan:
Fair point, the C++ reference doesn't belong here. Let me restate it in terms of this library.

Within Java Collections, `Collections.binarySearch` is the pattern: "The list must be sorted into ascending order according to the natural ordering of its elements... If it is not sorted, the results are undefined." The JDK documents the precondition and trusts the caller rather than sorting defensively. 

Within Commons Collections, `CollatingIterator` in this same package does the same thing. It provides an ordered iteration over a collection of ordered iterators, states that in the first line of its javadoc, and never sorts its inputs. Unordered input gives unordered output. That's the precedent I should have cited from the start.                                      
                                                                                                                                                                                                                                                                                                                                                                                   
The substantive reason is about what each design permits. A caller who wants the complete set can sort before constructing, one line at the call site. A caller who wants to resume from a previously reached arrangement, or to split the permutation space across workers by giving each a different starting point, has no recourse if the constructor sorts. Sorting is not recoverable from outside the class, so the version that preserves the given order is strictly the more capable of the two.

What I've pushed documents the precondition, since the javadoc previously contradicted it, and adds tests pinning it for both natural ordering and a supplied comparator.

If you'd still rather the complete set be the default, I'd suggest a static factory such as `overAll(collection)` that sorts a copy, keeping the start-anywhere behavior available on the constructor.

src/main/java/org/apache/commons/collections4/iterators/LexicographicPermutationIterator.java
```diff
@@ -0,0 +1,232 @@
...
+
+    /**
+     * Indicates if there are more permutation available.
```
garydgregory:
```suggestion
     * Indicates if there are more permutations available.
```

src/main/java/org/apache/commons/collections4/iterators/LexicographicPermutationIterator.java
```diff
@@ -0,0 +1,232 @@
...
+    /**
+     * Returns the next permutation of the input collection.
+     *
+     * @return A list of the permutator's elements representing a permutation
+     * @throws NoSuchElementException if there are no more permutations
+     * @throws ClassCastException if no comparator was supplied and the elements are
+     *         not mutually {@link Comparable}
+     */
+    @Override
+    public List<E> next() {
```
garydgregory, line 141: The method returns the internal List<E> reference that was used for the previous step. `PermutationIterator` does the same, so the behavior is consistent with the existing code base, but it means callers receive a mutable list that is no longer referenced internally after the call. Would a defensive copy or unmodifiable view would be safer?

Alternatively, the Javadoc could state that the returned `List` is mutable and should be treated as read-only.

garydgregory (changes requested):
Hello @hextriclosan 

Thank you for the update. Please see my 2 comments in the code.

WRT tests, how about adding:

- Explicit null-collection test: the constructor uses Objects.requireNonNull but there is no explicit assertThrows(NullPointerException.class, …) test.
- Equality when comparators differ, and when the iterators are at different positions.
- Hash-code stability/change after next().
- Behavior with all-equal elements: only one permutation should be produced.
- Defensive-copy expectation: a test that mutating a returned list does not affect subsequent next() calls.
- `forEachRemaining()` is inherited from AbstractIteratorTest is exercised, but a specific test for the lexicographic order of forEachRemaining would be nice.

TY!

hextriclosan:
Hello @garydgregory,

Thank you for the review and the advice.

What changed:
  - Javadoc polished
  - Missing tests added
  - `next()` now returns a defensive copy, covered by new tests

Two points where I would value your opinion:

**1. `next()` returning a copy.**
The copy is taken before the iterator advances, so the caller's list is detached regardless of what the successor computation does:
- pros: callers may modify the returned list freely, and the guarantee no longer depends on internal ordering.
- cons: one extra list allocation per permutation, and since `PermutationIterator` hands out its internal list, the two classes now differ internally.

**2. Are equals/hashCode worth keeping?**
They looked reasonable to me, but I have doubts:

- No other iterator in o.a.c.c.iterators overrides them, and neither do the JDK iterators. Identity semantics seem to be the convention.
- The hash code changes as the iterator advances, so an instance used as a HashMap key becomes unfindable after the first next().
- Two exhausted iterators built from different collections compare equal, since equality is defined on the comparator and the next permutation (see `testEqualsAndHashCodeForExhaustedIterators`). Defensible, as both will emit nothing further, but surprising.

Thanks again for taking the time.

src/main/java/org/apache/commons/collections4/iterators/LexicographicPermutationIterator.java
hextriclosan: Fixed, thanks

src/main/java/org/apache/commons/collections4/iterators/LexicographicPermutationIterator.java
hextriclosan: Switched to a defensive copy. Details in a separate comment.

garydgregory:
Hello @hextriclosan, picking this low-hanging fruit: equals/hashCode aren't worth keeping for the first version of this class. Keeping the initial API surface small reduces maintenance, and the feature can be added later once there's a clear use case without worrying about backward compatibility. As you noted, other iterators don't support it, so adding it here would be inconsistent and invite questions about why we don't implement it everywhere, which we shouldn't take on.

hextriclosan:
> Hello @hextriclosan, picking this low-hanging fruit: equals/hashCode aren't worth keeping for the first version of this class. Keeping the initial API surface small reduces maintenance, and the feature can be added later once there's a clear use case without worrying about backward compatibility. As you noted, other iterators don't support it, so adding it here would be inconsistent and invite questions about why we don't implement it everywhere, which we shouldn't take on.

Hello @garydgregory,
I've removed equals/hashCode

hextriclosan:
Hello @garydgregory, pushed the fixes for your comments, build is green. Could you take another look when you get a chance?

PR #726, opened 2026-08-15
Fix AbstractLinkedList addAll returning true for an empty collection

rootvector2:
`addAll(int, Collection)` hard-codes `return true`, so adding an empty collection reports a change that never happened, breaking the `List.addAll` contract (return `true` only if the list changed). Spotted because the class's own `LinkedSubList.addAll` already returns `false` for an empty collection, as does `TreeList.addAll`. The fix returns `false` for an empty collection after the `getNode` range check, so an out-of-range index still throws `IndexOutOfBoundsException`. `AbstractLinkedListJava21` is a copy of this class with the same defect and gets the same one-line fix. Regression test in `AbstractListTest` runs against every list impl; it fails before for `CursorableLinkedList`, `NodeCachingLinkedList` and the `AbstractLinkedListJava21` default list, and passes after.


garydgregory:
@rootvector2 Thank you for the PR, reviewed and merged 🚀

PR #727, opened 2026-08-17
[COLLECTIONS-898] AbstractSetTest cannot be used with an explicitly typed Set (unless it's based on String)

tmortagne:
Pick a valid element from getFullElements(), instead of adding a hardcoded String.

I did not dare making more changes to the logic of that test, but I'm honestly not sure there is that much value in the `"Empty set shouldn't equal nonempty set"` bit, compared to the `"Sets with different contents shouldn't be equal"` one at the end. I mean it's not really the point of this test to validate the `#add` vs `#addAll` APIs behaviors.

There does not seem to already exist a test class designed specifically to validate `AbstractSetTest` behavior (but there are various test classes which rely on `AbstractSetTest`), so I did not create a new one for such a small change. But I may have missed it.

- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This may not always be possible, but it is a best practice.

### DIFF
--- a/src/test/java/org/apache/commons/collections4/set/AbstractSetTest.java
+++ b/src/test/java/org/apache/commons/collections4/set/AbstractSetTest.java
@@ -124,7 +124,7 @@ public abstract class AbstractSetTest<E> extends AbstractCollectionTest<E> {
         verify();
 
         final Collection<E> set2 = makeConfirmedCollection();
-        set2.add((E) "foo");
+        set2.add(getFullElements()[0]);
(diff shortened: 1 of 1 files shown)


garydgregory:
Jira ticket is [COLLECTIONS-898](https://issues.apache.org/jira/browse/COLLECTIONS-898).

tmortagne:
Hmm, indeed sorry about that bad copy/paste. Fixed.

garydgregory:
@tmortagne Thank you, merged 🚀

tmortagne:
Thanks !

PR #728, opened 2026-08-19
Fix nextKey skipping a key and throwing for keys not in the map

rootvector2:
`AbstractSortedMapDecorator.nextKey` walks `tailMap(key)` and discards the first element to step over `key` itself, but never checks that `key` is present. When `key` is absent that first element is the successor, so it gets dropped and the wrong key comes back; when `key` is at or past the end `tailMap` is empty and `it.next()` throws `NoSuchElementException` instead of returning `null`. `DualTreeBidiMap.nextKey` repeats the same `tailMap().iterator().next()` shape and fails the same two ways, since its `isEmpty()` guard only covers a fully empty map.

Return `null` when the map does not contain `key`, matching the documented `null if no match` contract and every other `nextKey` in the library (`AbstractLinkedMap`, `ListOrderedMap`, `TreeBidiMap`, `PatriciaTrie` all return `null` for an absent key); the skip-first logic then runs only for a present key, the case it was written for. `FixedSizeSortedMap` and `UnmodifiableSortedMap` inherit the decorator fix. Found by auditing the `OrderedMap.nextKey` implementations against the contract.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/bidimap/DualTreeBidiMap.java
+++ b/src/main/java/org/apache/commons/collections4/bidimap/DualTreeBidiMap.java
@@ -25,6 +25,7 @@ import java.util.Comparator;
 import java.util.Iterator;
 import java.util.ListIterator;
 import java.util.Map;
+import java.util.NavigableMap;
 import java.util.SortedMap;
 import java.util.TreeMap;
 
@@ -364,15 +365,15 @@ public class DualTreeBidiMap<K, V> extends AbstractDualBidiMap<K, V>
 
     @Override
     public K nextKey(final K key) {
-        if (isEmpty()) {
+        if (isEmpty() || !normalMap.containsKey(key)) {
             return null;
         }
         if (normalMap instanceof OrderedMap) {
             return ((OrderedMap<K, ?>) normalMap).nextKey(key);
         }
         final SortedMap<K, V> sm = (SortedMap<K, V>) normalMap;
-        if (!sm.containsKey(key)) {
-            return null;
+        if (sm instanceof NavigableMap) {
+            return ((NavigableMap<K, V>) sm).higherKey(key);
         }
         final Iterator<K> it = sm.tailMap(key).keySet().iterator();
         it.next();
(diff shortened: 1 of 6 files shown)


garydgregory (changes requested):
Hello @rootvector2 

Thank you for the PR. How about adding tests for:

- An explicit test for `DualTreeBidiMap.nextKey()` with an absent key that is between existing keys. The abstract test uses `getOtherKeys()` which are outside the sample range, so the bug would still be hidden if the key were in-range.
- An explicit test for `UnmodifiableSortedMap.nextKey()`. It inherits the decorator fix but has no dedicated test.

Should the code be using `NavigableMap.higherKey()` when available? `TreeMap` implements `NavigableMap`, so `higherKey(key)` would give the correct successor in one call and would be null for absent keys automatically.

Should we guard the `OrderedMap` branch in `DualTreeBidiMap.nextKey()` for consistency; either ensure the delegated implementation also checks presence or apply the same containsKey guard before delegation.

Isn't there a similar problem with `previousKey()`? That could be addressed in a different PR.

garydgregory:
@rootvector2 ping 🔔

rootvector2:
Pushed the requested changes:

- `DualTreeBidiMapTest#testNextKeyAbsentKey` covers an absent key between existing keys ("b" in {a, c, e}) plus one past the end, and `UnmodifiableSortedMapTest#testNextKey` exercises the inherited decorator fix directly.
- Switched both implementations to `NavigableMap.higherKey` when the underlying map supports it. The `containsKey` guard has to stay though: `higherKey` returns the least key strictly greater than the argument whether or not it's present, so an absent in-range key would still get the successor instead of null. It only handles the past-the-end case automatically.
- Hoisted the guard in `DualTreeBidiMap.nextKey` above the `OrderedMap` delegation so both branches agree on absent keys.
- `previousKey` has the same absent-key inconsistency: it returns the predecessor instead of null (no throw, since `headMap` already excludes the key). I'll send that in a separate PR as you suggested.

Full default `mvn` goal is green.

garydgregory:
@rootvector2 Thank you, merged 🚀

PR #729, opened 2026-08-28
Reset last-returned element in EnumerationIterator.remove

rootvector2:
`remove()` never nulls `last`, so a second call without an intervening `next()` deletes another element equal to the last returned one from the backing collection instead of throwing `IllegalStateException`; found sweeping the package's `remove()` methods, where the sibling adapters (`EntrySetMapIterator`, `ZippingIterator`, `IteratorChain`) already clear their last-returned field.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/iterators/EnumerationIterator.java
+++ b/src/main/java/org/apache/commons/collections4/iterators/EnumerationIterator.java
@@ -120,6 +120,7 @@ public class EnumerationIterator<E> implements Iterator<E> {
             throw new IllegalStateException("next() must have been called for remove() to function");
         }
         collection.remove(last);
+        last = null;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


garydgregory:
@rootvector2 Thank you, merged 🚀

garydgregory:
@rootvector2 Do other `Iterator` implementations suffer from the same problem? WDYT?

rootvector2:
I swept every `remove()` in the iterators package before opening this. The other adapters that track a last-returned element (`EntrySetMapIterator`, `ZippingIterator`, `ObjectGraphIterator`, `IteratorChain`) all null it out after removing, and `CollatingIterator`/`LazyIteratorChain`/`ListIteratorWrapper` delegate to a child iterator whose own state check throws on a repeat call. `EnumerationIterator` was the only one removing by value from the collection without clearing its state, so I believe this was the sole offender. Thanks for merging.

garydgregory:
@rootvector2 
Thank you for looking into it.

PR #730, opened 2026-08-31
COLLECTIONS-900: IteratorChain#addIterator() does not work properly with IteratorChain parameters

mbaedke:
Added failing test case.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/iterators/IteratorChain.java
+++ b/src/main/java/org/apache/commons/collections4/iterators/IteratorChain.java
@@ -175,7 +175,11 @@ public class IteratorChain<E> implements Iterator<E> {
                 // in case it is an IteratorChain, wrap every underlying iterators as unmodifiable
                 // multiple rechainings would otherwise lead to exponential growing number of function calls
                 // when the iteratorChain gets used.
-                for (final Iterator<? extends E> nestedIterator : ((IteratorChain<? extends E>) underlyingIterator).iteratorQueue) {
+                IteratorChain<? extends E> iteratorChain = ((IteratorChain<? extends E>) underlyingIterator);
+                if (iteratorChain.currentIterator != null) {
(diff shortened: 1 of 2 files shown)


garydgregory:
main and test changes don't match as the build fails in tests.

garydgregory:
@mbaedke 
The build fails.

rishabhdaim (approved):

garydgregory:
Jira ticket is [COLLECTIONS-900](https://issues.apache.org/jira/browse/COLLECTIONS-900).

mbaedke:
Thx @garydgregory , fixed it.

garydgregory:
@mbaedke 
Thank you for the update. 
I think we need more tests to verify nothing breaks for the full sequence, exhaustion, `remove()`, empty leading iterators, and rechaining after at least one successful `next()` call.

mbaedke:
> @mbaedke Thank you for the update. I think we need more tests to verify nothing breaks for the full sequence, exhaustion, `remove()`, empty leading iterators, and rechaining after at least one successful `next()` call.

Yes, I'll get that done by EOW

mbaedke:
> I think we need more tests to verify nothing breaks for the full sequence, exhaustion, remove(), empty leading iterators, and rechaining after at least one successful next() call.

@garydgregory : Sorry for the late response. After thinking about it: this PR together with https://github.com/apache/commons-collections/pull/628 fundamentally change the way addIterator() treats ChainedIterators. Wouldn't it be very reasonable to parameterize the test so that it runs a second time with makeObject() creating a chain of ChainedIteraors instead of ListIterators?

mbaedke:
@garydgregory , what do you think about that?

garydgregory:
@mbaedke in my queue to review...

mbaedke:
@garydgregory , I did not do that yet. I would if you support the idea.

PR #731, opened 2026-09-02
Fix putAll(K, Iterable) NPE for null value transformer

rootvector2:
`putAll(K, Iterable)` sends values through `FluentIterable.of(values).transform(valueTransformer)`, and `IterableUtils.transformedIterable` rejects a null transformer with `NullPointerException`, so a map built with a null value transformer (documented as "no conversion", and handled fine by `put` via `transformValue`) throws on `putAll` while `put` works; skip the transform when the transformer is null.

- [x] I used AI to create any part of, or all of, this pull request. Which AI tool was used to create this pull request, and to what extent did it contribute? Claude Code (Anthropic) was used to locate the bug, write the fix and the regression test, and run the build.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01HUmTCFCnFFDWKGVYWbUDtA

### DIFF
--- a/src/main/java/org/apache/commons/collections4/multimap/TransformedMultiValuedMap.java
+++ b/src/main/java/org/apache/commons/collections4/multimap/TransformedMultiValuedMap.java
@@ -133,8 +133,7 @@ public class TransformedMultiValuedMap<K, V> extends AbstractMultiValuedMapDecor
     public boolean putAll(final K key, final Iterable<? extends V> values) {
         Objects.requireNonNull(values, "values");
 
-        final Iterable<? extends V> transformedValues =
-                valueTransformer == null ? values : FluentIterable.of(values).transform(valueTransformer);
+        final Iterable<V> transformedValues = FluentIterable.of(values).transform(this::transformValue);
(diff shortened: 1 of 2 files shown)


garydgregory (commented):
Hello @rootvector2 
That method still might not be correct.
The null check on `valueTransformer` fixes the NPE, but values are inserted without calling the `transformValue()` method when that field is null. A subclass can override `transformValue()` and pass `null` to the superclass, in which case `put()` transforms values while this `putAll()` now silently bypasses the override. We should likely route every element through `transformValue()`.
And add a regression test using an overriding subclass.
Since this might be in addition to the NPE you found, feel free to fix it here or in a new PR. Up to you.
TY!

rootvector2:
Good call. `putAll(K, Iterable)` now routes every value through `transformValue()` instead of checking `valueTransformer` directly, so a subclass that overrides `transformValue()` while passing `null` to the superclass still gets its conversion, matching `put`. Added a regression test with an overriding subclass alongside the null-transformer one.

garydgregory:
Thank you @rootvector2 , merged 🚀

PR #736, opened 2026-09-07
Fix previousKey returning the predecessor and throwing for keys not in the map

lenamonj:
`previousKey` carries the defect #728 fixed in `nextKey`. `AbstractSortedMapDecorator.previousKey` returns `headMap(key).lastKey()` without checking that `key` is present, so an absent key inside the range returns its predecessor and a key outside a range-restricted view throws `IllegalArgumentException`; `DualTreeBidiMap.previousKey` fails the same way. `OrderedMap` documents `null` for a key with no match, and every other `previousKey` in the library returns it.

Return `null` when the map does not contain `key`, and use `NavigableMap.lowerKey` when the decorated map offers it, mirroring #728 line for line.

Tests: both ordered-map conformance suites now assert the absent-key case on a full map, plus four explicit tests; red on master and green with the change. Default Maven goal green.

- [x] I used AI to create any part of, or all of, this pull request. Which AI tool was used to create this pull request, and to what extent did it contribute? Claude (Anthropic) found the defect and drafted the fix and tests; I reviewed them, reproduced the failure on a fresh clone of master, and ran the default Maven goal.

### DIFF
--- a/src/main/java/org/apache/commons/collections4/bidimap/DualTreeBidiMap.java
+++ b/src/main/java/org/apache/commons/collections4/bidimap/DualTreeBidiMap.java
@@ -385,13 +385,16 @@ public class DualTreeBidiMap<K, V> extends AbstractDualBidiMap<K, V>
 
     @Override
     public K previousKey(final K key) {
-        if (isEmpty()) {
+        if (isEmpty() || !normalMap.containsKey(key)) {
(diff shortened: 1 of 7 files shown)


lenamonj:
Superseded by #738, which Gary asked for on #728; closing.

